PR #276, opened 2025-10-20
feat(cli): Add graceful signal handling and configurable threading

JeanMertz:
Users can now gracefully interrupt long-running operations like streaming LLM responses using Ctrl+C or other OS signals. The application will save conversation state before exiting, preventing data loss during interruption.

A new `--threads` flag allows configuring the number of worker threads, defaulting to the number of available CPU cores. This provides better control over resource usage for different runtime scenarios.

Technical changes include a new `jp_macro` crate providing a `rustfmt`- compatible select! macro and manual runtime management replacing `#[tokio::main]`.

### DIFF
--- a/crates/jp_macro/src/lib.rs
+++ b/crates/jp_macro/src/lib.rs
@@ -4,7 +4,7 @@
 /// the same, but supports rustfmt. It uses closure syntax which rustfmt
 /// supports with support for `if branches` and the "biased" keyword.
 ///
-/// ```rust
+/// ```rust,ignore
 /// select!(
 ///     biased,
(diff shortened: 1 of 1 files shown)


JeanMertz:
This should be expanded such that `SIGINT` should trigger `SignalTo::Interrupt` and when an LLM query is running, we should show an interactive prompt that allows users to:

1. Interrupt the LLM by replying before it is done responding
2. Stop the LLM (e.g. current behaviour, which would also trigger if SIGINT is triggered again (e.g. Ctrl+C twice)
3. ... potential other options in the future.

Issue #620, opened 2026-05-07
git_diff ouput errors silently


rxgrant:
LLM confused by git_diff output regarding a staged file deletion.

Proposal: be sure to explain new and deleted files even when truncating.

For background, the LLM ran git diff earlier:

    Let me check for staged/unstaged changes.

    Calling tool git_diff(status: "staged")
    Calling tool git_diff(status: "unstaged")

    So there are staged changes but no unstaged changes.

Decoding the git_diff events, the first file, here assigned [file1] was deleted, but is shown as:

    Changed files:
      [file1] (158 lines, truncated to 50)

After the rest of the files are listed, the first 50 deleted lines of [file1] are shown in the diff.

Then at the moment of the incident:

    The [other] file diff is large (469 lines) and was truncated. 
    Let me get the full diff.

    The [other] diff is large, let me get it in full along with the
    remaining staged files.

    Calling tool git_diff(status: "staged", paths:
    ["[file1]","[file2]","[file3]"])

    Hmm, the first call returned "No changes" for [these] instruction
    files.  But those files appeared in the overall staged diff.  Let
    me try again - maybe they're using different filenames.

    Wait, I see - the staged diff showed:
    
    - `[file1]` (deleted)
    - `[file2]` (modified)
    - `[file3]` (new)

    But **these didn't show up when I queried them specifically**.  This
    is odd.  Maybe the files aren't actually staged?  Let me check.

      [ed: emphasis added]

    Actually, there's a contradiction here - the first diff call said
    "No changes" for those three paths, yet the overall staged diff
    clearly lists them as changed.  Let me verify the actual staged
    files more carefully and check if there's a typo in the paths I'm
    using.

    [...long exploration...]

    So the unstaged files also show "No changes" for those [path1]/
    files I was confused about.  This means those files ARE staged.
    The tool just returned "No changes" when I queried them
    individually, which is odd.

    Wait, actually - let me re-check.  If they appear in the broad
    staged diff but not in the specific staged path query, maybe
    they're tracked differently.  Let me check if `[file1]` exists on
    disk.

    Calling tool fs_list_files(prefixes: ["[path]"])

    So `[file1]` does NOT exist on disk - it was deleted.  And it
    appeared as a deletion in the broad staged diff.  The tool might
    just not be able to show deleted files when queried individually
    for staged diffs.  That's fine.

rxgrant:
Okay, we found that the git plumbing silently errored and actually returned no changes where the LLM claimed it was so.

[...]

PR #621, opened 2026-05-07
chore(tools): Add `git_diff_file` and fix deleted-file diff bugs

JeanMertz:
`git_diff` truncates each file's diff at 50 lines to keep the overview manageable. Previously, users had to re-run with `paths` to see a full diff, but `git_diff` applied the same per-file cap regardless — there was no escape hatch.

`git_diff_file` fills that gap. It is the working-tree analog of `git_diff_commit`: given a `status` and a required `paths` list, it returns the full diff for those files, with an optional `pattern` parameter to grep within large diffs (same behavior as `git_diff_commit`). Diffs beyond 500 lines are truncated and the user is directed to use `pattern`.

To share the `grep_diff` and `truncate_diff` logic between `git_diff_commit` and `git_diff_file`, those helpers were extracted into a new `diff_filter` module.

Two bugs in `git_diff` were also fixed:

- Deleted-file paths caused `git diff-index` and `git diff-files` to abort with "fatal: no such path in the working tree" because git resolved paths against the working tree before the `--` separator. Adding `--` to both commands fixes this.
- A non-zero git exit with empty stdout was previously reported as "No changes." instead of surfacing the actual error. The tool now returns an error outcome in that case.

The `git_diff` tool description and skill config were updated to clarify that `paths` narrows the file overview but does not bypass the per-file truncation cap, and to point users to `git_diff_file` when they need a full diff.


rxgrant:
fixes #620

PR #631, opened 2026-05-11
chore(tools): Add `git_blame` tool to project tooling

JeanMertz:
Implements a `git_blame` tool backed by `git blame --porcelain`,
exposing per-line authorship over a bounded line range to the assistant
during pair-programming sessions.

Output is grouped by commit: consecutive lines that share the same
blamed commit are collapsed into a single hunk, so token cost scales
with the number of distinct commits rather than the number of lines.
Each hunk carries `hash`, `short_hash`, `author`, `date`, `summary`, and
— when present — a `previous` field containing the commit that owned
those lines before the current change. That lets the assistant drill
back through line history with a follow-up `git_show` or
`git_diff_commit` call without an extra blame invocation.

The range is capped at 200 lines per call. Validation rejects zero-based
start lines, inverted ranges, and oversized windows with a clear error
message before any subprocess is spawned. An optional `revision`
parameter allows blaming at any commit, tag, or branch;
`ignore_whitespace` passes `-w` through to git.

The tool is wired into the MCP tool registry and added to the
git-reading skill so it is available whenever that skill is active.

### DIFF
--- a/.jp/mcp/tools/git/blame.toml
+++ b/.jp/mcp/tools/git/blame.toml
@@ -38,9 +38,10 @@ Find who last touched a function body:
 ```
 
 Drill back one layer: blame the lines as they were before the current \
-change, using the `previous` hash from a prior call:
+change, using the `previous` hash from a prior call (porcelain's \
+`previous` is already the prior commit — do not append `^`):
 ```json
-{"path": "crates/jp_llm/src/stream.rs", "start_line": 42, "end_line": 58, "revision": "abc1234^"}
+{"path": "crates/jp_llm/src/stream.rs", "start_line": 42, "end_line": 58, "revision": "abc1234"}
 ```
 
 Ignore whitespace-only changes when assigning blame:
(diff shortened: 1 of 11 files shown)


.jp/mcp/tools/git/blame.toml
```diff
@@ -0,0 +1,78 @@
...
+examples = """
+Find who last touched a function body:
+```json
+{"path": "crates/jp_llm/src/stream.rs", "start_line": 42, "end_line": 58}
+```
+
+Drill back one layer: blame the lines as they were before the current \
+change, using the `previous` hash from a prior call:
+```json
+{"path": "crates/jp_llm/src/stream.rs", "start_line": 42, "end_line": 58, "revision": "abc1234^"}
```
JeanMertz: blocking: the prose says to use the `previous` hash, but this example passes `abc1234^`. The porcelain `previous` value is already the prior commit, so adding `^` walks one parent too far and blames the wrong revision. Please use the previous hash directly here (or reword the example if `abc1234` is meant to be the blamed commit instead).

PR #650, opened 2026-05-16
chore(tools, github): Add `repository`, `page` params + comment fetching

JeanMertz:
The `github_issues` and `github_pulls` tools were previously hardcoded to the project's own GitHub repository and returned no comments. This commit addresses both limitations.

Both tools now accept a `repository` parameter in `owner/repo` form, defaulting to the project's own repo when omitted. A `parse_repo()` helper centralises the parsing and default logic, replacing ad-hoc `split_once('/')` calls scattered across the tool modules.

A `page` parameter was added to both tools so that long comment threads can be walked incrementally. When a specific `number` is provided, the tool fetches that page of conversation comments (10 per page) and includes them in the response alongside a `comments_count` total. The page size is intentionally bounded to keep responses within LLM context limits.

The underlying `jp_github` crate gained a `list_comments()` method on `IssuesHandler`, backed by a new `IssueCommentListBuilder` and a `Comment` model. The `Issue` model also now carries a `comments` field (the total count) which GitHub exposes on both the list and detail endpoints.

Auth handling was also improved: `github_issues`, `github_pulls`, and `github_read_file` now use a new `auth_optional()` path that falls back to anonymous access when no token is configured. Read-only operations against public repos no longer require a token.

Finally, `web_fetch` now detects GitHub issue and PR URLs and returns a redirect error pointing at the correct dedicated tool, rather than silently returning near-empty HTML (GitHub renders comments client-side).

### DIFF
--- a/.config/jp/tools/src/github/issues.rs
+++ b/.config/jp/tools/src/github/issues.rs
@@ -1,23 +1,37 @@
 use chrono::{DateTime, Utc};
 use url::Url;
 
-use super::auth;
-use crate::{
-    Result,
-    github::{ORG, REPO, handle_404},
-    to_xml,
-};
+use super::{auth_optional, parse_repo};
+use crate::{Result, github::handle_404, to_xml};
 
-pub(crate) async fn github_issues(number: Option<u64>) -> Result<String> {
-    auth().await?;
+/// Comments-per-page when fetching a specific issue. Fixed at 10 to keep
+/// responses bounded; long threads are walked with the `page` parameter.
+const COMMENTS_PER_PAGE: u8 = 10;
+
+pub(crate) async fn github_issues(
+    repository: Option<String>,
+    number: Option<u64>,
+    page: Option<u64>,
+) -> Result<String> {
+    auth_optional().await?;
+
+    let (owner, repo) = parse_repo(repository)?;
+    let page = page.unwrap_or(1).max(1);
 
     match number {
-        Some(number) => get_issue(number).await,
-        None => get_issues().await,
+        Some(number) => get_issue(&owner, &repo, number, page).await,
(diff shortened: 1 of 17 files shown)


.config/jp/tools/src/github/issues.rs
```diff
@@ -66,10 +106,11 @@ async fn get_issues() -> Result<String> {
         created_at: DateTime<Utc>,
         closed_at: Option<DateTime<Utc>>,
         linked_pull_request: Option<Url>,
+        comments_count: u64,
     }
 
     let page = jp_github::instance()
-        .issues(ORG, REPO)
+        .issues(owner, repo)
```
JeanMertz: blocking: now that `repository` can point at any public repo, the list path is unbounded. `IssueListBuilder::send()` auto-paginates until an empty page, so `github_issues {"repository":"rust-lang/rust"}` will try to fetch every open issue/PR and can hit the 60 req/hour anonymous limit or produce a huge XML response. Keep this bounded too: accept a list `page`/`per_page`, or return only the first page instead of auto-paginating arbitrary repos.

.jp/mcp/tools/github/issues.toml
```diff
@@ -4,30 +4,60 @@ run = "unattended"
 
 source = "local"
 command = "just serve-tools {{context}} {{tool}}"
-summary = "Find one or more issues in the project's GitHub repository."
+summary = "Find one or more issues in a GitHub repository."
 
 examples = """
```
JeanMertz: nit: this says "all issues", but the GitHub repository issues endpoint defaults to open issues and `IssueListBuilder` does not set a `state` query parameter. Either call this "List open issues" or add a `state` parameter for issues too, otherwise the example over-promises what the tool returns.

PR #752, opened 2026-06-15
feat(llm, cli): Surface Anthropic model refusals to the user

JeanMertz:
Add a `FinishReason::Refused { category, explanation }` variant to the
`FinishReason` enum in `jp_llm`. This represents a terminal state where
a safety classifier declined the request rather than the model failing.

On the Anthropic provider side, `stop_reason: "refusal"` is now mapped
to this new variant, pulling `category` and `explanation` out of the
`stop_details` field when present. Both fields are `Option<String>` to
handle uncategorized refusals gracefully.

In the CLI coordinator, the `finish_notice` function now handles this
variant, producing a human-readable notice like:

    The model declined this request (cyber): This request was declined.

When neither category nor explanation is present, the notice falls back
to the simpler form:

    The model declined this request.

### DIFF
--- a/crates/jp_cli/src/cmd/query/turn/coordinator.rs
+++ b/crates/jp_cli/src/cmd/query/turn/coordinator.rs
@@ -590,6 +590,20 @@ fn finish_notice(reason: &FinishReason, dropped_tools: &[String]) -> Option<Stri
                 names.join(", ")
             ),
         }),
+        FinishReason::Refused {
+            category,
+            explanation,
+        } => {
+            let category = category
+                .as_deref()
+                .map_or_else(String::new, |c| format!(" ({c})"));
+            let explanation = explanation
+                .as_deref()
+                .map_or_else(|| ".".to_string(), |e| format!(": {e}"));
+            Some(format!(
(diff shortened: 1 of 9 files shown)


crates/jp_cli/src/cmd/query/turn/coordinator.rs
```diff
@@ -590,6 +590,20 @@ fn finish_notice(reason: &FinishReason, dropped_tools: &[String]) -> Option<Stri
                 names.join(", ")
             ),
         }),
+        FinishReason::Refused {
```
JeanMertz: blocking: `Refused` needs a different path than the generic `Event::Finished` handling. In `handle_event`, all finish reasons first call `self.event_builder.drain()` and push any buffered message/reasoning into the conversation stream before `finish_notice` sees the reason. Anthropic documents mid-stream refusals as incomplete output that must be discarded, and the new `FinishReason::Refused` docs say the same. As written, a refusal after partial text will still persist that partial assistant response. Please special-case `FinishReason::Refused` before draining, or add a discard path and a regression test for partial text followed by refusal.

PR #772, opened 2026-06-19
chore(tools): Fix non-UTF-8 bytes silently swallowing diff output

rxgrant:
`DuctProcessRunner` converted captured stdout/stderr with `String::from_utf8(...).unwrap_or_default()`. When a staged diff contained even a single non-UTF-8 byte — a latin-1 source file, a near-binary blob git still treats as text — the entire capture was discarded and replaced with an empty string. An empty stdout with a zero exit code fell through to the `"No changes."` branch in `git_diff_impl`, making a non-empty staged diff completely invisible.

Switching to `String::from_utf8_lossy(...).into_owned()` preserves the output and replaces only the invalid bytes with the Unicode replacement character, which is the correct behaviour for a tool that needs to display diff content rather than parse it precisely.

A regression test stages a file containing byte `0xe9` (no NUL, so git treats it as text and emits the raw byte in the diff) and asserts the result is neither `"No changes."` nor missing the filename.

### DIFF
--- a/.config/jp/tools/src/util/runner.rs
+++ b/.config/jp/tools/src/util/runner.rs
@@ -212,9 +212,13 @@ impl ProcessRunner for DuctProcessRunner {
 
         let output = command.run()?;
 
+        // Lossy conversion is deliberate: git diff output for text files can
+        // contain stray non-UTF-8 bytes (e.g. a latin-1 source file). A strict
+        // `from_utf8().unwrap_or_default()` would discard the entire capture on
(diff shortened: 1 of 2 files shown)


JeanMertz:
Good catch. Some of our non-official tools are more "vibe-coded" than most other code in JP. These mistakes are a clear sign of LLMs taking the easy route. We should probably spend some time verifying the other tools separately.

JeanMertz (approved):

PR #839, opened 2026-07-03
docs(rfd): Add RFD 094 for built-in `tell_user` tool

JeanMertz:
Propose a built-in `tell_user` tool that lets the assistant deliver a mid-turn message the user must see without ending the agentic loop: the message renders as chat content, the assistant gets a short acknowledgement, and the loop continues.

Most of the design activates mechanisms [RFD 058] defines but does not yet act on: honoring MCP `audience` annotations at JP's two view boundaries (provider payload vs. terminal/plugin display), rendering `text/markdown` content blocks through the `jp_md` pipeline, and a new `style.header` knob to suppress the "Calling tool" header line. `tell_user` is the first consumer of these mechanisms rather than a special case in jp-core.

Amend [RFD 058] itself to add the optional `mimeType` field and MCP-standard `annotations` to `ContentBlock::Text`, both referenced by RFD 094, and update the RFD summary index accordingly.


JeanMertz:
Note to self: style.header should be added to the list of options that the assistant is not allowed to modify.

PR #866, opened 2026-07-09
feat(cli, workspace): Add `jp workspace` and RFD 087 bootstrap

JeanMertz:
Adds `jp workspace` (alias `jp w`), with `use`, `ls`, and `show` subcommands, plus a global `--workspace` grammar (ID, path, `cwd`, `-` for stdin, and interactive keywords like `?`, `s`, `l`). Selecting a workspace with `jp w use` makes it the session's active workspace, so subsequent commands in that terminal session run against it from any directory, the way an active conversation already follows the session (RFD 020). Standing inside a different workspace than the session's active one now prompts to resolve the conflict (`c`/`C`/ `a`/`A`/`q`), with `A` pinning the session sticky to its active workspace across cwd changes.

A new pre-workspace bootstrap step in `jp_cli` resolves the launch cwd, the selected checkout root, and the working directory for spawned children once, before any `Workspace` is constructed. When JP operates on a workspace other than the launch cwd's own, MCP servers, plugins, and the `.jp.toml` config chain all run as if launched from the selected workspace root instead of the process cwd. Commands that only need to select or inspect a workspace (`jp w ls`, `jp w use`) no longer pay for workspace construction or conversation indexing.

Workspace checkouts are now tracked in a per-workspace roots registry (`workspace/<slug>-<id>/roots/`) instead of the single `storage` symlink, so a workspace with multiple live checkouts (e.g. git worktrees) can be targeted and listed by any of them; a one-time migration folds the legacy symlink into the registry. A new user-global `sessions/` store maps each terminal session to its active workspace, cleaned up on the same source-split liveness rules as the existing per-workspace session-to-conversation mapping (RFD 020).

`WorkspaceIdOrPath` and the ad hoc `-w <id>` directory scan are removed in favor of the shared `WorkspaceTarget` grammar and roots registry, and the "silo" terminology in docs and code comments is renamed to "user-workspace directory" for clarity.


crates/jp_cli/src/cmd/workspace/show.rs
```diff
@@ -0,0 +1,376 @@
...
+                    path.clone()
+                } else {
+                    env.launch_cwd.join(path)
+                };
+                let root = Workspace::find_root(base, DEFAULT_STORAGE_DIR)
+                    .ok_or_else(|| format!("No workspace found at `{path}`."))?;
+                let id = root_id(&root)
+                    .ok_or_else(|| format!("`{root}` has no readable workspace ID."))?;
+
+                Ok(Some(subject_for(env, id, "explicit target")))
```
JeanMertz: blocking: This discards the resolved `root` and builds the readout only from the roots registry. `jp w show /path/to/live/checkout` can report `(no live checkouts)` when that checkout has not been registered yet, which is likely when `show` is the first workspace command run against it because this path never calls `load_workspace`. Include the resolved root in `Subject` or register it before rendering. The `cwd` branch has the same issue.

PR #872, opened 2026-07-21
local issue store and GitHub issue conduit

rxgrant:
This RFD proposes a local-first store for issues, kept as git objects in the project's own `.git` under a dedicated ref namespace, synced between contributors with ordinary push and pull, including pulled from GitHub and processed into the local storage using a "conduit". 

Each issue is a set of per-writer, append-only operation logs whose union deterministically folds into issue state, so concurrent offline edits (as anticipated in later implementations but not proposed as a JP-centric workflow) never merge-conflict and never silently drop data.

GitHub is mirrored by a scraper using `jp_github`, and all issues for JP would continue to be input on GitHub.


docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,627 @@
...
+GitHub is a read-only interface where some issues get added.
+It is not the authority.
+The local store is the source of truth, designed from day one for a world
+where local edits are made concurrently on machines that do not know about
+each other.
+
+Three storage approaches fail the requirements:
+
+- **Checked-in files** cause merge conflicts for people working in worktrees,
+  and dirty the code's evolution with their own commit history.
```
parasyte, line 35: I could also add that checked-in issues follow code evolution itself. When you look at an old branch, you expect to see code as it once was but not old issues as they once were. Old issues may have since been closed or updated with new information. New issues created, and so on.

An argument could be made that such behavior is desired. I do not believe it to be desirable. The evolution of issues is not generally a concern. Resurrecting issues that were closed as WONT-FIX or INVALID have a conclusion and decision that gets erased by switching to an old branch. The chronology of issue management is nonsense.

This is not a criticism of the highlighted lines. Just adding an extra perspective on how checked-in issues fails a simple sanity check.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,627 @@
...
+- **A centralized per-user store** mixing all repositories (as radicle's
+  `/storage/` does) is rejected for security reasons - it would
+  require inventing a DSL for access control.
+- **Last-write-wins merging** (git-bug's approach, in elaborated form) hides
+  conflicts by silently dropping the losing write.
```
parasyte: A slightly altered form of LWW that maintains multiple values for client-side conflict resolution is a viable (though annoying and error-prone) solution to this problem. A lot of systems use multi-value registers for client-side conflict resolution. [Riak comes to mind](https://docs.riak.com/riak/kv/latest/developing/usage/conflict-resolution/index.html) (calling multi-values "siblings").

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,627 @@
...
+
+Storing issues as git objects in the repo's own `.git` — objects in the object
+database, refs under a dedicated namespace, nothing in the worktree — avoids
+all three.
```
parasyte, line 46: Not sure I agree with the statement. `git-bug` uses the git object store, and yet it chose LWW conflict resolution; one of the three "storage approaches" that using the git object store supposedly fixes.

1. LWW merging is not a storage approach.
2. `git-bug`'s storage approach *is* the git object store.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,627 @@
...
+Each write is one standard commit object, used as an envelope:
+
+- **tree**: a single blob, `ops.json` — the payload.
+  It carries the operations of this write, a `seen_heads` field (defined
+  below), and a format version.
+  Tools read and write `ops.json` and nothing else; its versioned schema is
+  the compatibility boundary between replicas, and changing it is a breaking
+  change for every clone that holds copies.
+- **parent**: the previous commit on this writer's chain (none for the first).
+  Exactly one parent, always: every ref is a strictly linear chain.
```
parasyte, line 125: I'm skeptical of the writer-id stuff (so far). But this fast-forward-only property is *very* appealing if it works.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,627 @@
...
+This relation is the only input to conflict resolution.
+
+**Resolution**, per the fold semantics in the op table:
+
+- **Multi-value register** (title, body, state, comment body, priority): the
+  register's state is the set of ops on it that do not happen-before another
+  op on the same register.
+  One op in that set yields a single value — the normal case.
+  Two or more ops in that set — possible only for concurrent writes — are
+  all retained and all displayed, ordered by op id.
```
parasyte: Surprise-surprise! Multi-value registers make an appearance. This is honestly the least-bad conflict resolution strategy that can be chosen without doing a great deal of work to detect and preserve intent from concurrent writes for text. With the exception of `priority`, all of these MVR's are text data types.

While it's least-bad, it also has some annoyances. Mostly in the form of showing all values to a client, who will most likely want to resolve the conflict by committing to one of the values or some amalgam of many values.

It is possible for many clients to attempt resolution concurrently by making different decisions, producing another conflict. Therefore, annoyance.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,627 @@
...
+- **Observed-remove set** (labels): a label is present when some `add-label`
+  op for it does not happen-before a `remove-label` op for it.
+  A remove affects only adds it acknowledged; a concurrent add survives.
```
parasyte: I'm familiar with this data type by the name ["Two-Phase Set (2P-set)"](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type#2P-Set_(Two-Phase_Set)).

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,627 @@
...
+  target from that point on; concurrent edits to the target stay in the log
+  but are not displayed.
+  Hiding is the entire mechanism; the bytes are never removed from the
+  store.
+- **Keep** (moderation): a `keep` op from any single moderator (1-of-n)
+  overrides a tombstone on the same target.
+  The kept content renders however the keeping moderator chose: full text,
+  a marker, or a moderator-written summary.
+  Without a `keep`, a single redact or delete suffices to hide (also
+  1-of-n).
```
parasyte: This is an unusual requirement! Naturally, I extrapolate this to "what about undoing a *keep*"? At some point, you have to admit that the best thing to do in situations where an item was deleted is to create a brand new one.

I think that applies here, and *keep* is YAGNI: When an item is deleted (tombstoned) that is a commitment for removal. If you **really* want to resurrect a deleted item, you should create a new one containing the original content. Because it, too, can be deleted using the same protocol without inventing "undo the undo of the undo" operations ad infinitum.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,627 @@
...
+
+### Sync
+
+`issue sync` is fetch plus push, nothing else:
+
+1. Fetch `refs/jp/issues/*` with a non-forcing refspec: every chain is
+   append-only, so every legitimate update is a fast-forward.
+   A non-fast-forward foreign ref means someone rewrote history; the tool
+   refuses the update, keeps its local copy, and reports the ref.
+2. Push the local writer's refs.
```
parasyte: IIUC, *all* refs need to be pushed. The local writer's refs can reference refs from other writers via `seen_heads`.

The push strategy defined here works for a single git remote as a centralized coordination point, but can fall apart in various ways with multiple remotes. I hope I'm wrong, but this sounds sketchy.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,627 @@
...
+git applies no special rules to `refs/jp/issues/*`: a pusher with write
+access can force-push a rewritten chain, and the server accepts what every
+jp tool would refuse.
+A `pre-receive` hook closes the gap by rejecting any update to
+`refs/jp/issues/*` that fails to fast-forward the ref's current value,
+contains a commit with more than one parent, or contains a commit whose
+signature does not match the key hash in the ref path — the same checks
+the fold applies at read time, run at push time.
+The hook is stock git tooling: any server whose operator can install
+`pre-receive` hooks can run it.
```
parasyte: I don't think this applies to GitHub.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,627 @@
...
+
+**Keeping deleted content.** A user's delete is not final: any single
+moderator may keep the deleted messages by recording a `keep` op in their
+own chain, naming the tombstoned target and the rendering they chose —
+full text, a marker, or a moderator-written summary.
+Nothing needs rescuing: the content is still in the author's chain, merely
+hidden, and the `keep` op changes how the fold renders it.
+
+**Capabilities.** Actions are authorized per signing key, evaluated at fold
+time.
```
parasyte: YAGNI?

A much simpler protocol can get you 95% of what is needed for issue management:

1. Every writer can do every operation to items they created.
2. A single "moderator/administrator/power user" kind of flag *configured locally and shared out-of-band amongst collaborators* allows writers to perform every operation on every item, regardless of provenance.

The moderator flag lives at the same web-of-trust level as allowed-signers.

Non-collaborators can still file issues and edit their own comments. Editing anything else is rejected at fold-time. BUT just saying that raises the YAGNI question: Are non-collaborators expected to file issues by pushing their own refs? I suspect not.

My recommendation is YAGNI: No capabilities and no moderator flags. Non-collaborators will use GitHub issues and all issues are managed by collaborators only, including through the scraper conduit.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,627 @@
...
+
+- `<number>` is the GitHub issue number, or — in the later local-creation
+  phase — a `jp_id`-formatted id for an issue born locally; the two forms
+  are syntactically disjoint.
+  The literal `meta` is reserved for the store-level metadata log.
+  The store holds the repo's own issues only; a cross-repo mirror would get a
+  sibling namespace and is out of scope.
+- `<writer-id>` identifies one worktree of one clone, held by one signer.
+  It is composed of three segments,
+  `<workspace-id>/<unique-hash-of-signing-key>-<avatar-nickname>/<worktree-id>`:
```
parasyte: Why is `<worktree-id>` included in the path for `<writer-id>`? Am I misunderstanding the definition of "worktree"?

My understanding of the word is that it refers to a working copy of whatever particular branch you have currently checked out. And in that case, `<worktree-id>` is the commit hash for HEAD? What does that buy? Down below this is a comment about each clone being a different worktree so it's a unique writer ID, but that doesn't sound right. In that case, the ID in question is not HEAD. I have no idea what it is! The hash of the repo's root CWD?

In fact, including the JP workspace ID in these paths is unusual, because (AFAIK) JP only has one workspace ID for each repo.

This whole thing is confusing. The only bit that makes any sense is `<unique-hash-of-signing-key>-<avatar-nickname>`.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,627 @@
...
+| `set-body` | issue | multi-value register |
+| `set-state` | issue | multi-value register |
+| `add-label` / `remove-label` | issue | observed-remove set |
+| `add-comment` | comment (by GitHub id) | grow-only set |
+| `set-comment-body` | comment (by GitHub id) | multi-value register |
+| `set-comment-visibility` | comment (by GitHub id) | multi-value register |
+| `set-priority` | issue | multi-value register |
+| `delete-issue` | issue | tombstone |
+| `delete-comment` | comment (by GitHub id) | tombstone |
+| `redact-op` | op (by op id) | tombstone |
```
parasyte: How is redact distinct from delete? What is the rationale behind it?

parasyte (commented):
General impression: Not too bad for a first draft. I did end up with about a dozen questions, comments, and suggestions. No explicit approval for this iteration.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: Adding a word about viewing current issues on branches.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,659 @@
+# RFD D24: Git Object Store for GitHub Issues
+
+- **Status**: Draft
+- **Category**: Design
```
JeanMertz, line 4: nit: "design" implies this is about a [feature/architectural change in JP](https://jp.computer/rfd/001-jp-rfd-process#templates), whereas this is more tied to the **Process** category.

I could even reason that this isn't tied to JP's process, but that this is a generic proposal for a new workflow/tool on how to manage issues in VCS/Git-backed projects, but given we already have the RFD workflow settled in JP, I do see value in exploring this here, instead of externally.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,659 @@
...
+- **Authors**: rgrant <rgrant@contract.design>
+- **Date**: 2026-07-19
+
+## Summary
+
+This RFD describes a local-first store for GitHub issues, kept as git objects in
+the project's own `.git` under a dedicated ref namespace.
```
JeanMertz, line 11: > dedicated ref namespace

Does this use [git namespaces](https://git-scm.com/docs/gitnamespaces), or does this refer to an [orphaned branch](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-orphan)?

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,659 @@
...
+A scraper tool mirrors issues and their comments from GitHub through
```
JeanMertz: mirroring implies bi-directional. Does this mean an issue I create outside of GH ends up on GH's issue board?

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,659 @@
...
+dropped data.
+
+## Motivation
+
+Contributors need offline access to the project's issues, and later the ability
+to edit them locally (`issue append`) and build views over them (a kanban tool).
+The consumers are our own jp-tools: `issue show` first, `issue append` and a
+kanban view later.
+
+GitHub is a read-only interface where some issues get added.
```
JeanMertz: I don't quite understand this line. It is read-only, but we still allow creating issues there? I *suspect* that if (when 😏) JP becomes popular, the _majority_ of issues are created via GitHub by external contributors and users so the "some issues get added" might hold up today, but I doubt it'll hold up in the long run.

And even so, I also feel like we are glossing over the added friction here, even for ourselves. Take an issue  like https://github.com/dcdpr/jp/issues/873:

- You (quickly, conveniently) create a two-line issue,
- You then easily add a one-paragraph comment
- I can see it, I get a notification in the GH user interface, and I can reply easily

If this were done outside of the UI, the UX (I assume, still reading) would become:

- You create a new branch
- You create the issue locally using the terminal
- You create a PR to push that branch containing the issue
- It has to be merged
- I have to pull it, potentially without knowing what new issues it contains, since these arent stored as plain text, right?
- I then have to manually see that there is a new issue, or you have to inform me, and I go through the same dance to add a reply that you can see

I don't know. This sounds great on a technical level, but UX-wise I think people might be put off by the added friction.

Unless the goal is for GH Issues to remain a first-class *input source* for issues, that are then auto-synced to our git-object-based-issue-store, but that seems not to be what is proposed here?

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,659 @@
...
+A contributor's responsibility is to pull; after that they have everything they
+need for offline work.
+
+## Design
+
+### What the user sees
+
+```sh
+# sync: fetch everyone's issue refs, push every issue ref you hold
+jp-tools issue sync
```
JeanMertz: I should note that `jp-tools` is not designed to be used in the way it is proposed here.

`jp-tools` is our "collection of tools built for JP development, exposed to LLMs via tool calls", it requires JSON objects as its first and second arguments, it does not have any conventional CLI arguments that it exposes.

This seems to be more appropriate for the (early, in draft) [**RFD D05: Internal Dev Plugin for RFD Workflows**](https://jp.computer/rfd/drafts/D05-internal-dev-plugin-for-rfd-workflows#general-purpose-jp-rfd-subcommand), which proposes a `jp dev ...` subcommand plugin used only for JP development (and not shipped/exposed to other users of JP). That would allow for `jp dev issue sync`, etc.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
```diff
@@ -0,0 +1,659 @@
...
+
+# read a folded issue, offline
+jp-tools issue show 42
+
+# poke at the raw store with stock git
+git for-each-ref refs/jp/issues/42/
+git log refs/jp/issues/42/<writer-id>
```
JeanMertz, line 63: I think ideally, this RFD is split into two smaller ones: first we experiment with this technique _without_ any specific JP tie-in, meaning, we document how this would work using purely `git`, and create some kind of test repository to play around with the workflows and how it all fits together.

Then, once we think this is the right direction, we shift our focus to integrating it back into this repository, and building any additional support tooling that makes for a nicer UX (such as `jp dev issue sync`, etc).

That way, we avoid building/writing code without first validating that this approach works well on a technical level, without worrying about the UX layer on top.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
JeanMertz: > In fact, including the JP workspace ID in these paths is unusual, because (AFAIK) JP only has one workspace ID for each repo.

Correct.

JP does support working in multiple Git worktrees concurrently for a single JP workspace (see: [RFD 031: Durable Conversation Storage with Workspace Projection](https://jp.computer/rfd/031-durable-conversation-storage-with-workspace-projection#store-conversations-at-the-bare-repo-level)), but that is irrelevant in this context.


> This whole thing is confusing. The only bit that makes any sense is `<unique-hash-of-signing-key>-<avatar-nickname>`.

I tend to agree with this. And it touches on my broader point I made in another comment: the need for an architectural/structural tie-in with JP escapes me. I can see how our _JP development workflow_ could benefit from the locally-stored issues, but I don't see how locally-stored issues need any knowledge or relation to JP-the-product.

In other words, this RFD could be written completely separate from JP, in its own repo, tool/project/repository-agnostic. Yes, we would have a second RFD in _this_ repo on how we would incorporate such a workflow into JP's development practices, but those are two separate things to me, one is a project/issue-tracking-workflow that anyone can use if they think it benefits them, the other is a description on how _we_ would use it in our way-of-working.

JeanMertz (commented):
Reviewed. My thoughts after the first round of draft/review cycle:

I am _very_ interested in this on a technical level, and would _love_ for someone to start experimenting with this, to see how viable it is, and how well we can integrate it into our workflow.

However, two things stick out to me in its current form:

1. I am having difficulties seeing this as replacing GitHub Issues UX (or Codeberg, or whatever hosting platform we eventually choose) for the majority of people interacting with us/our projects as users. So a statement such as "GitHub is a read-only interface where some issues get added" seems almost naive. If any of our projects (including JP) ever gain any popularity (which, of course they will), then I suspect 90+% of issues will be created by users, not us, and I suspect 99% of those will be created through whatever interface throws up the least amount of friction, which would be the Issues web interface that GitHub/Codeberg/... offers.

2.  While I get why the RFD is created in this repository for convenience sake, I don't see any needed tie-in with JP on a technical level. This RFD (should) describe(s) a generic issue tracking and management system built on top of Git, and us developers working on JP will adopt it in our workflow and we'll build tooling around it to make it convenient to use, but on a technical/foundational level, I see no need, and in fact would strongly advice against a direct relationship between the issue-tracking system and JP.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: My dear associate explains:

> Close relative, but not the same type: in a 2P-set a removed element can never
> be re-added, while here a later `add-label` re-adds a removed label (untag
> `wontfix`, re-tag it next month). The re-add behavior is why the spec says
> observed-remove set. 

Added a line to the draft making the re-add explicit.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: With the help of my associate:

> Agreed, and the design uses that mechanism. Title, body, state, comment body,
> and priority are all multi-value registers. Concurrent writes are all retained
> and displayed, and a later write that acknowledges every live value collapses
> the register back to one value. Riak's siblings are the same idea.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: Header fixed.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: With the help of my associate:

> The writer id is simpler now. The workspace id segment is gone, and the
> worktree id is renamed to replica id. The remaining form is
> `<key-hash>-<nickname>/<replica-id>`. The replica id exists to protect the
> fast-forward property you like: one signer works from several checkouts at
> once, and if the key alone named the writer, two of those checkouts would
> append to the same ref independently and one push would fail to fast-forward.
> Each checkout owning its own chain removes that case. The store-layout section
> now says this explicitly.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: Yeah, best we can do here.  Nothing in the fold treats resolution as a special case.  (Keep resolving.)

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: Dropped keep and made tombstones final.  We can rely on the collaborative environment aspects and, ultimately, repo write permission managment.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: With the help of my associate:

> `seen_heads` names op ids inside `ops.json`, not commit hashes, so a single
> pushed ref is self-contained at the git object level. The problem you smelled
> was one level up: with own-refs-only pushing, a remote could lack some
> writer's chain forever, and every clone from that remote would hit dangling
> acknowledgments. Sync now pushes every `refs/jp/issues/*` ref the replica
> holds, not just its own. The missing-acknowledgment error also no longer
> claims tampering outright: the draft now separates "not fetched yet" from
> "rewritten", and fetching from more remotes is what distinguishes them.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: Oh, true.  Rewritten to spots it applies.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: Agreed on YAGNI.  Removed, with no moderator.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: Simplifed to `<key-hash>-<nickname>/<replica-id>`, with replica-id being a checkout or git worktree thing.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: Removed.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
parasyte: I misunderstood the labeling semantics, and an OR-set does make more sense with the new context.

TIL: This type is also known as an [Add-Wins Set (AW-set)](https://crdt.tech/glossary).

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
parasyte: While true that `seen_heads` does not directly use git refs, it must do so indirectly to create the causal graph. This isn't just a "problem smell", it's the nature of causality.

Pushing only the user's local refs would work in a scenario where the user's local refs are fully self-contained without a happened-before relationship to refs pushed by other users. That just is not anticipated to be the scenario used in practice in a collaborative environment.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
parasyte: > The replica id exists to protect the fast-forward property you like

Somewhat surprising, but acceptable! I just don't do enough with multiple checkouts. This is a good rationalization.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: > read-only, but we still allow creating issues there 

Poorly worded, and changed.  The intent is to offer users a native interface into issues created on Github, similar to the simplicity of reading RFDs.  In this initial spec there is no discussion of how to write to Github, as that's an implementation detail, unrelated to data model.  

> If this were done outside of the UI

A few things here: these issues don't exist as files or in a branch, so there's no merge.  There is pulling and pushing, but the beauty of the scheme is how git-native it is.  No developer has to pull or push to any remote that they don't already have declared (ie. if git is set up, you can use it already - no new files to back up or lose, etc).  We don't expect anyone to create issues locally, until developers who opt-in really like the tools to do so, and the local-to-Github conduit sync (where every dev syncs their own changes) is nicely polished.

Overall the scheme creates a git-without-Github (including offline) possibility that improves everyone's experience even if they keep only using Github, in the sense that you can collaborate with people who don't use Github, and in the sense that your local-LLM always has the data it needs (and can index it with a custom indexer).

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: Agreed that they're not jp-tools for starters.  Also thanks for syntax check.  Made edits.  Disagree that deployed projects don't want an issue store that can pull from Github.  (Although I might agree that this should be a separate project outside JP, I'm continuing to fix it up for now.)

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
parasyte: One of the most important things needed (from my POV) is issue automation. But it is out scope for this proposal.

What GitHub does really well is not the UI, per se. It's features like these:

1. Issues can be [closed by referencing the issue number in commit comments](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests).
2. Issues automatically get back-links displayed, e.g., in the issue's thread *with the current status of that issue* (open, closed).
3. [Task lists](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/about-tasklists) are simple Markdown that automatically updates when the checkboxes are toggled.

These automation tools make for smooth workflows. The back-links are the important half of issue relationships (which are hard to do right: see how Jira messes this up).

I must also acknowledge that this RFD is for a simple CLI that pulls issues from GitHub and synchronizes it with the local git object store. (GitHub is read-only from the perspective of this CLI.) Issue management is out of scope for now, including automation, UIs, and Kanban/project boards. These can all be built on the git object store.

As a point of comparison, `git-bug` has a local web server you can run with an issue management interface. It doesn't have any of the cool features I mentioned above, apart from Markdown *rendering*. But it's a good bit of prior art to see where this could go.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: Neither.  Clarified in RFD.

parasyte: Git namespaces are written to git objects in `refs/namespaces/`. Issues are written to git objects in `refs/jp/issues/`. These are not branches and issues are not visible in the work-tree.

The git objects are transparent to git (you can use git commands like `git cat-file` with them) but the file contents are opaque to git (JSON-encoded CRDT operations). The relevant documentation is [git objects](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects).

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: Well, it specifies a data model, a wire format (`ops.json`), a sync protocol, and an attack analysis.  How we adopt this store into JP's development workflow sounds like the process RFD.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: One-way only for now.  Scraper reads from GitHub and writes to the local store, nothing
flows back. Motivation now states this.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: I can see data model separated from any UX or JP-workflow stuff, but I'm not going to separate the document in to a maybe-wont-work hedge.

rxgrant:
> 2. While I get why the RFD is created in this repository for convenience sake, I don't see any needed tie-in with JP on a technical level. 

Non-JP implementation does make sense.  I wanted to get it in so that we can get issues pulled locally and start using a kanban to communicate priorities that span RFDs and GitHub issues, which isn't easy.  

I wanted to preempt seeing prioritization or other tooling agglomerate on a worse data model.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
rxgrant: > No developer has to pull or push to any remote that they don't already have declared [...]

Slight addendum here, the tooling does propose adding a local git repository config setting to pull (and push) additional refs.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
JeanMertz: > Agreed that they're not jp-tools for starters. Also thanks for syntax check. Made edits. Disagree that deployed projects don't want an issue store that can pull from Github. (Although I might agree that this should be a separate project outside JP, I'm continuing to fix it up for now.)

Definitely. I don't disagree that this is a useful tool for many (which is why git-bug and git-issues are also used by many), I'm just saying the thing it does/solves is entirely orthogonal to what JP does/solves, so we shouldn't conflate them in a single product.

I am fine experimenting with it in this repository and building it out here, there is a reason we have `crates/contrib`, which contains many smaller projects that aren't really related to, but are used by, JP as a project. So if we want to start with it here, we can add `crates/contrib/git-tracker` (or whatever we name it), but that should not mean the actual implementation requires JP-specific code/crates, IMO.

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
JeanMertz: Yes, but the point is those RFDs are for JP-data-models, JP-wire-formats etc, not an unrelated crate that lives in `crates/contrib` for incubation reasons. The RFDs in this repo are RFDs related to JP's development, not third-party project development.

I can also see value in something similar to what Oxide does with their company-wide RFDs (see: https://rfd.shared.oxide.computer). I can even see us having some kind of aggregator that takes RFDs from different projects in their own repos, and displays them in a single DCD interface somewhere.

I should note that I am fine with whatever decision we make, just giving arguments for keeping the RFDs in this repo focused on JP, but I am also fine moving forward as-is (and then potentially later migrating it to a separate repository once the incubated project turns out to be valuable).

docs/rfd/drafts/D24-git-object-store-for-github-issues.md
JeanMertz: Right, this clarifies things for me. The most important pushback from me is that I want to avoid requiring outside contributors to having to jump through hoops to be able to report bugs (e.g. see open source projects that require CLA signing).

From what I now understand, that won't happen here, so the native UI on GItHub remains accessible and fully useable by outside contributors. Perhaps at some point we can point them to our _own_ UI that we host somewhere which integrates nicely with this data storage proposal, but that is out-of-scope for now.

parasyte: > The most important pushback from me is that I want to avoid requiring outside contributors to having to jump through hoops to be able to report bugs (e.g. see open source projects that require CLA signing).

I agree with your position. "GitHub as path of least resistance" is the same stance I held with @rxgrant when this whole thing began last week-ish.

I think there is a good compromise here, though. GitHub is still usable by collaborators. What it provides is extending collaboration to people who do not want to (or cannot) use GitHub. Forcing everyone to use GitHub is the problem @rxgrant is trying to solve.

> From what I now understand, that won't happen here, so the native UI on GItHub remains accessible and fully useable by outside contributors.

:100::exclamation:

PR #928, opened 2026-07-31
fix(config): Fix inquiry ceiling inheritance and `fields()` gaps

JeanMertz:
`conversation.inquiry.assistant.request.max_response_bytes` no longer silently disables the output ceiling when only a sibling field (e.g. `cache`) is set in that block. The inquiry `assistant` block now mirrors the full `assistant` config and inherits every unset field from it during resolution, rather than resolving unset fields to Rust's `Default`. Users can still disable the ceiling for inquiries alone with an explicit `0`/`"disabled"`, independent of the parent assistant's setting.

`max_response_bytes` itself is now a `MaxResponseBytes` value instead of a bare `u32`, accepting a byte count or `0`, `"disabled"`, `"off"`, `false` to turn the ceiling off, e.g.
`jp config set assistant.request.max_response_bytes disabled`. This removes the previous ambiguity where `0` had a different meaning depending on which config layer it appeared in.

`AppConfig::fields()` (and therefore `envs()`) now descends into `Option<T>` nested configs and into unions that describe one value written multiple ways (a shorthand plus its expanded table form), so keys like `style.reasoning.summary_model.id`,
`editor.cmd.program`/`args`/`shell`, and
`conversation.tools.*.enable.state` are reported and assignable via env vars, where before the containing block was reported as an opaque leaf. `editor.cmd` and `conversation.tools.*.enable` keep their shorthand form addressable too: writing `editor.cmd.args` after `editor.cmd = "code --wait"` now expands the shorthand and keeps the named program instead of being rejected.

Schema support for this (`schema_union_with`, `UnionType::expanded_index`) lives in `schematic_macros`/`schematic_types`, letting a derived struct schema be unioned with hand-declared shapes for types whose `Deserialize` accepts more than their fields describe.


crates/jp_config/src/lib.rs
```diff
@@ -333,7 +361,17 @@ impl AppConfig {
...
+    pub fn from_partial_with_defaults(mut partial: PartialAppConfig) -> Result<Self, ConfigError> {
+        // Inquiry settings inherit from the top-level assistant. This runs
+        // before the `#[setting(default)]` layer below, so an unset inquiry
+        // field takes the user's configured assistant value rather than the
+        // type's default.
+        partial.conversation.inquiry.assistant = partial
+            .conversation
+            .inquiry
+            .assistant
+            .fill_from(partial.assistant.clone());
```
JeanMertz: blocking: A user who sets `assistant.instructions` or `assistant.system_prompt_sections` and leaves the inquiry block unset loses those entries for inquiries. `MergeableVec::fill_from` deliberately keeps `self`'s items and copies only wrapper metadata; here `self` is the inquiry's empty vector, so this call does not copy the assistant's items. Defaults later fill the inquiry from its own schematic defaults, and `build_sections(inquiry_cfg, ...)` then omits the user's top-level sections. Please use inheritance semantics that can distinguish an unset mergeable vector from an explicit empty one, and add regression tests for both collection fields.

crates/jp_config/src/conversation.rs
```diff
@@ -229,11 +224,15 @@ impl ToPartial for ConversationConfig {
...
     ///
-    /// Unset fields fall back to the parent assistant config.
+    /// Accepts every key from the top-level `assistant` section.
+    /// Keys left unset here take the value from `assistant`, so
+    /// `conversation.inquiry.assistant.model.id` can point an inquiry at a
+    /// cheaper model while its system prompt and request settings stay whatever
+    /// the main assistant uses.
     #[setting(nested)]
-    pub assistant: AssistantOverrideConfig,
+    pub assistant: AssistantConfig,
```
JeanMertz: blocking: A user can set `conversation.inquiry.assistant.model.parameters.temperature = 0.1` or `conversation.inquiry.assistant.request.cache = "off"`, but the inquiry request silently keeps the main assistant's value. `LlmInquiryBackend::inquire` sends the cloned parent `ConversationStream`, and each provider reads model parameters and cache policy from `thread.events.config().assistant`; `build_inquiry_backend` extracts only the model ID, prompt, sections, and byte ceiling. Either carry the effective inquiry assistant config into that cloned stream before calling the provider, with a request-level regression test, or expose only the fields the inquiry path actually honors.

crates/jp_config/src/lib.rs
```diff
@@ -310,7 +326,19 @@ impl ToPartial for AppConfig {
...
+
+        // Inquiry settings are resolved by inheriting from the top-level
+        // assistant, so a field that merely equals the assistant's value holds
+        // no choice of the user's. Keeping only the differences is what lets a
+        // later layer that changes `assistant` still reach the inquiry: a
+        // recorded copy of the inherited value would pin it instead.
+        partial.conversation.inquiry.assistant = self
+            .assistant
+            .to_partial()
+            .delta(partial.conversation.inquiry.assistant);
```
JeanMertz: blocking: Equality does not tell us whether the inquiry value was inherited. For example, a user can explicitly set both `assistant.request.max_response_bytes` and `conversation.inquiry.assistant.request.max_response_bytes` to `4096` to pin the inquiry ceiling. This delta erases the inquiry setting when `ConversationStream::into_parts` serializes its resolved base config. If a later conversation delta raises the main ceiling, the inquiry ceiling follows it even though the user explicitly pinned it. Preserve explicit-versus-inherited state rather than inferring it from equal values, and test an equal explicit override across a stream save/load followed by an assistant-only change.

crates/jp_config/src/types/command.rs
```diff
@@ -82,6 +94,24 @@ impl AssignKeyValue for PartialCommandConfigOrString {
...
 }
 
+/// Expand the string shorthand into the table form it abbreviates.
+///
+/// Splits exactly as [`CommandConfigOrString::command`] does, so expanding
+/// before assigning a field cannot change which command ends up running.
+/// `shell` and any empty part are left unset so their defaults still apply, and
+/// so a later layer can still fill them.
+fn expand_shorthand(shorthand: &str) -> PartialCommandConfig {
+    let mut tokens = shlex::split(shorthand).unwrap_or_default().into_iter();
```
JeanMertz: nit: A command such as `just x {{ a | default('') }}` changes shape when a later layer assigns a nested field such as `shell`: `CommandConfigOrString::command` uses `split_command_words` so the template span stays one argument, while this raw `shlex::split` breaks it into several arguments. Tool commands then surface a template error instead of running, so the failure is visible and contained. Reuse `split_command_words` here and include a template span in `expanding_a_shorthand_matches_the_command_it_describes`.

crates/jp_config/src/assistant/request.rs
```diff
@@ -153,7 +154,9 @@ impl AssignKeyValue for PartialRequestConfig {
             "stream_idle_timeout_secs" => {
                 self.stream_idle_timeout_secs = kv.try_some_u32()?;
             }
-            "max_response_bytes" => self.max_response_bytes = kv.try_some_u32()?,
+            "max_response_bytes" => {
+                self.max_response_bytes = kv.try_some_number_or_from_str()?;
```
JeanMertz: nit: The documented `false` spelling fails through JP's JSON assignment path. For example, `--cfg assistant.request.max_response_bytes:=false`, or the same field inside a root JSON object, reaches this as `KvValue::Json(Value::Bool(false))`; `try_some_number_or_from_str` accepts only JSON numbers and strings and returns a type error. The error is visible and contained, but this makes an advertised value depend on assignment syntax. Use a bool-aware parser here and add a JSON-assignment test.

crates/contrib/schematic_macros/src/common/container.rs
```diff
@@ -192,6 +193,10 @@ fn generate_enum_schema(
                 default_index = Some(i);
             }
 
+            if v.args.expanded {
+                expanded_index = Some(i);
```
JeanMertz: theoretical: `i` is the index in the unfiltered variant list, but `variants_types` drops excluded variants below. If an excluded variant appears before the `expanded` variant, `expanded_index` points at the wrong schema or past the end. I found no current enum combining `exclude` and `expanded`, so there is no plausible JP config input that triggers this today. Derive the index from the filtered output position, or reject that attribute combination at macro expansion time.

crates/jp_config/src/conversation.rs
JeanMertz: Confirmed and fixed. I verified both halves: every provider reads the assistant config off the request's stream (`anthropic.rs:1136-1137`, `openai.rs:360-362`, `google.rs:196-197`, `cerebras.rs:433-434`, `llamacpp.rs:441-442`, `ollama.rs:272-273`, `openrouter.rs:744-745`), and `inquire` hands them `Thread { events, .. }` built from the cloned parent stream. `build_inquiry_backend` passed only provider, model details, system prompt, sections, and the byte ceiling, so everything else came from the parent.

This PR made it worse rather than inherited it: widening the block from `AssistantOverrideConfig` to the full `AssistantConfig`, plus the `fields()` change, means `conversation.inquiry.assistant.model.parameters.*` is now listed and assignable while still being ignored. Documented-and-inert is the failure mode I called the worse of the two when triaging #923, so this one is mine.

I took your first option. `InquiryConfig` gains `assistant: PartialAssistantConfig`, and `inquire` appends it to the request's stream as a config delta:

```rust
let mut delta = PartialAppConfig::empty();
delta.assistant = config.assistant.clone();
events.add_config_delta(delta);
```

A delta rather than a rebased base config, deliberately. `jp_llm::title::rebase_on_model` is the closest precedent and it collapses the conversation's effective config into a fresh base, which requires the stream to be projected first or compaction overlays are lost. The inquiry path doesn't project, so rebasing there would have quietly changed what the model sees. A delta applies on top of the conversation's own deltas without discarding anything.

Carrying it as a *partial* also made the per-question layer fall out for free: `per_q.clone().inherit_from(default_config.assistant.clone())` resolves that layer without a second partial-to-resolved pass. So `model.parameters` and `request.cache` now work at both the global and per-question layers, not just the one you named.

Request-level test as you asked: `llm_backend_applies_its_assistant_config_to_the_request` uses `MockProvider::capturing_requests`, sets `temperature = 0.125` and `cache = off` on the inquiry, and asserts what the provider actually received via `query.thread.events.config()`. Verified failing without the delta, reporting `None` where `Some(0.125)` was expected.</body>
</invoke>

crates/jp_config/src/lib.rs
JeanMertz: Correct, and the effect is worse than described. Fixed.

`MergeableVec::fill_from` (`types/vec.rs:207-231`) says it outright: "Items are always kept from self — only the `Merged` wrapper metadata is inherited." So the inquiry's lists never took the assistant's entries. Combined with this PR switching `build_sections(&cfg.assistant, …)` to `build_sections(inquiry_cfg, …)`, a user's `assistant.instructions` stopped reaching inquiries entirely.

Where I'd correct your description: the defaults don't fill the inquiry afterwards either. `MergeableVec::fill_from` only inherits metadata when the defaults side carries `strategy` or `dedup`, and `default_instructions` carries neither, so it falls through to `Self::Vec(items)` with `items` empty. I confirmed this by reverting the fix and running the new test: the inquiry ends up with **zero** instructions, not the type defaults. Every inquiry was running with an empty section list.

The fix is a distinct inheritance operation, since `fill_from`'s item semantics are right for a defaults layer and wrong for inheritance:

```rust
impl PartialAssistantConfig {
    pub fn inherit_from(mut self, parent: Self) -> Self {
        if self.instructions.is_empty() {
            self.instructions = parent.instructions.clone();
        }
        if self.system_prompt_sections.is_empty() {
            self.system_prompt_sections = parent.system_prompt_sections.clone();
        }
        self.fill_from(parent)
    }
}
```

On distinguishing unset from explicitly-empty, which you asked for: `MergeableVec::is_empty` already draws that line usefully. A `Merged` wrapper carrying only metadata is *not* empty, so `--cfg assistant.instructions.dedup=false` still inherits the parent's entries and applies its own strategy to them. What remains indistinguishable is a literal `instructions = []`, and there's no way to spell that distinctly today, so nothing expressible is lost.

Two tests, both fields, verified failing against plain `fill_from`: `inquiry_inherits_assistant_collections` and `an_explicit_inquiry_collection_wins_over_the_assistant`.</body>
</invoke>

crates/jp_config/src/lib.rs
JeanMertz: The diagnosis is right and I'm not fixing it. Reasoning, so you can overrule me.

Your scenario happens exactly as described: `delta` drops equal values, so an inquiry ceiling pinned to the same number as the assistant's is not recorded, and a later assistant-only change drags the inquiry along.

Where I disagree is that "preserve explicit-versus-inherited state" is available. A resolved `AppConfig` records values, not whether the user wrote them. Recovering that needs per-field presence to survive resolution, which means storing a partial inside the resolved config. That was considered and rejected in this branch as an antipattern, and I don't think this finding is enough to reopen it.

The alternative that needs no type change is to stop un-filling and record the inherited values verbatim. That's strictly worse: it breaks inheritance through *every* round-trip rather than only the equal-value case, so a conversation's stored config would pin the inquiry to whatever the assistant was at save time and no later `assistant` change would ever reach it. That's the failure `inquiry_inheritance_survives_a_partial_round_trip` exists to prevent, and it fires on a far more common path than two fields happening to hold the same number.

So both options are lossy and I've kept the one whose loss is narrower. What I've added instead:

- The reasoning is now in the code at the `to_partial` call site, naming equality as the only available signal and per-field presence as what a real fix needs.
- `a_same_valued_inquiry_pin_is_lost_on_a_round_trip` records the behavior as a characterization test, with a doc comment saying it pins a known limitation rather than desired behavior and pointing at the test whose guarantee it trades against.

I'd rather have it written down and tested than silently true. If you want the presence-tracking change, it's a `jp_config` type change worth its own PR and I'll file it.</body>
</invoke>

crates/jp_config/src/types/command.rs
JeanMertz: Taken. `expand_shorthand` now calls `split_command_words`.

This is a merge artifact worth naming: `split_command_words` arrived on `main` after I wrote `expand_shorthand`, and the merge left `command()` using the template-aware splitter while the expansion kept raw `shlex::split`. My doc comment on `expand_shorthand` claims "Splits exactly as `CommandConfigOrString::command` does", which was false from the moment the merge landed.

The part that bothers me more than the bug: `expanding_a_shorthand_matches_the_command_it_describes` was written specifically to enforce that invariant, and it passed anyway. Its five inputs contained no template span, and a template span is the *only* class of input where the two splitters disagree. So the test was pinning the invariant on precisely the cases that can't violate it.

Fixed by adding your case plus the other two span syntaxes, and noting in the test's doc why those inputs carry the weight:

```rust
"just x {{ a | default('') }}",
"echo {% if x %}on{% endif %} tail",
"echo {# a note #}",
```

Verified against the unfixed code:

```
expanding "just x {{ a | default('') }}" changed the command
  left:  args: ["x", "{{", "a", "|", "default()", "}}"]
  right: args: ["x", "{{ a | default('') }}"]
```</body>
</invoke>

File: crates/jp_config/src/assistant/request.rs (hunk 5 above)
Comment (JeanMertz) [reply crates/jp_config/src/assistant/request.rs]: Confirmed and fixed. `try_number_or_from_str` matches `Json(Number)` and `Json(String) | String` and type-errors on everything else, so a JSON bool never reaches `MaxResponseBytes`, which implements both `From<bool>` and `FromStr`.

Added `try_bool_number_or_from_str` (and its `try_some_` wrapper) alongside the existing `try_bool_or_from_str` and `try_number_or_from_str`, bounded `T: From<bool> + FromStr`, and switched the field to it. New helper rather than widening `try_number_or_from_str`, since that one is used by types like `Color` that have no `From<bool>` and adding the bound would force it on them.

One narrowing of the scope, for the record: `--cfg assistant.request.max_response_bytes=false` already worked, because that arrives as `KvValue::String` and goes through `FromStr`. TOML and JSON config files also worked, via `Deserialize::visit_bool`. Only the explicit-JSON spellings (`:=false`, or the field inside a root JSON object) failed. So the reachability is narrower than the comment implies, but you're right that an advertised value shouldn't depend on which syntax the user reaches for.

`max_response_bytes_accepts_a_json_bool` covers the bool, number, and quoted-string arms through `:=`. Verified failing against the old helper with `Type { value: "false", need: ["number", "string"] }`.</body>
</invoke>

File: crates/contrib/schematic_macros/src/common/container.rs (hunk 6 above)
Comment (JeanMertz) [reply crates/contrib/schematic_macros/src/common/container.rs]: Fixed, and extended to a second instance you didn't mention.

The bug is real: `i` came from `.enumerate()` over all variants while `variants_types` was built by a `filter_map` that drops excluded ones. Your reachability call is also right, and I checked it rather than taking it on faith: `expanded` appears exactly once in the tree (`CommandConfigOrString::Config`) and there is no `#[setting(exclude)]` anywhere in `jp_config`.

Worth adding: `default_index` has the identical bug on the line above and predates this PR. Fixing only `expanded_index` would have left the same trap one line away, so I did both.

I took your first option rather than rejecting the attribute combination, and structured it so the indices can't diverge again:

```rust
for variant in variants {
    if variant.is_excluded() {
        continue;
    }

    let index = variants_types.len();

    if variant.is_default() { default_index = Some(index); }
    if variant.args.expanded { expanded_index = Some(index); }

    variants_types.push(variant.generate_schema_type(is_all_unit_enum));
}
```

Both indices are now taken from the length of the vector they address, at push time, so there's no arithmetic left to get wrong. Rejecting the combination at expansion time would have prevented the same failure, but it also forecloses a combination that's perfectly meaningful once the index is right.

No test, deliberately: exercising it needs a fixture enum combining `exclude` and `expanded` that exists solely for the test, and after this change correctness is structural rather than something a test would be guarding. I'd rather not add a permanent fixture for a combination the codebase doesn't use. Say the word if you'd prefer one anyway.

Behavior is unchanged today, as expected: the `AppConfig::fields()` snapshot is untouched.</body>
</invoke>

PR #929, opened 2026-08-02
RFD for Interactive Compaction 

rxgrant:
RFD D46 adds an interactive mode to `jp conversation compact`.  Instead of applying a summary blind, the user reviews and revises a proposed compaction by chatting with the model in a forked "compaction conversation", then applies the agreed result to the source with `--continue` (or discards it with `--abort`). 

The review negotiates every field of an RFD 064 `Compaction` proposal (turn range, summary text, reasoning and tool-call policies), with each revision recorded as a call to a new built-in `propose_compaction` tool that JP validates against the source conversation as it is made.  Compaction does not need other tools from the jp-tools kit to be active in the conversation; for instance, it does not need to give the LLM permission to read any files.

### DIFF
--- a/docs/rfd/drafts/D46-interactive-compaction-review.md
+++ b/docs/rfd/drafts/D46-interactive-compaction-review.md
@@ -1,0 +1,494 @@
+# RFD D46: Interactive Compaction Review
+
+- **Status**: Draft
+- **Category**: Design
+- **Authors**: rgrant <rgrant@contract.design>
+- **Date**: 2026-08-02
+- **Extends**: [RFD 064]
+
+## Summary
+
+This RFD adds an interactive mode to `jp conversation compact`.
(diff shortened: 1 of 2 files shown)


docs/rfd/drafts/D46-interactive-compaction-review.md
```diff
@@ -0,0 +1,494 @@
...
+produces different text.
+So the user has two choices, neither good.
+Apply the compaction blind and inspect it afterwards, or skip summarization and
+lose its benefit.
+
+The failure is not rare.
+A summary that drops the current task state, or that covers a range ending in
+the middle of an unfinished topic, misleads the model worse than the long
+conversation it replaced.
+`compact --reset` removes every compaction event, not just the bad one, so
```
JeanMertz: Good point! Fixed in https://github.com/dcdpr/jp/pull/930.

PR #935, opened 2026-08-04
fix(cli): Make `-u` force a disabled tool for a single query

JeanMertz:
`jp query -u NAME` forces the assistant to call a specific tool, but it refused
any tool that was not already enabled. A tool that is off by default therefore
had to be named twice, once to enable it and once to select it:

    jp q -t my_tool -u my_tool "..."

`-u NAME` now runs the named tool whether or not it is enabled:

    jp q -u my_tool "..."

Nothing is enabled to make that work. `tool_definitions` has kept a forced tool
in the tool list since RFD 081 even while that tool is disabled, so the choice
alone is enough — and leaving the config untouched is what lets the force expire
with the query.

That split is now the rule across all four flags, and each says so in its help:

- `-u NAME` forces a tool for this query, running it even when disabled. Not
  persisted.
- `-U` suppresses tools for this query. Not persisted.
- `-t NAME` enables a tool for this query and every later one; `-T` is the undo.
- `-T NAME` disables a tool for this query and every later one; `-t` is the undo.

The asymmetry is deliberate. `-t` has to be sticky, because editing a config
file is otherwise the only way to turn a tool on for a conversation, and `-T`
makes that cheap to undo. There is no matching use for "force this tool on every
future turn", so `-u` binds only the query it was typed on.

Getting there fixed a pre-existing bug. `-u` and `-U` wrote
`assistant.tool_choice` into the partial config, which `get_config_delta_from_cli`
then persisted as a conversation delta — so a one-shot `jp q -u my_tool` left the
tool forced on every later bare query, contradicting the flag's own help ("not
persisted across queries"). Both flags are now read from the invocation at query
time and never reach the config.

`-T my_tool -u my_tool` is no longer a conflict: the disable persists, the force
applies to this query, and both hold.

A forced tool backed by an MCP server now starts that server even while the tool
is disabled. Without it the tool would be dropped from the tool list again as
unreachable, and the forced choice could not be satisfied.

A tool whose policy locks it off is still refused, and the refusal names the lock
(``cannot enable `my_tool`: this tool is configured as locked-off``) instead of
the misleading `tool choice 'my_tool' does not match any enabled tools`. An
unknown tool reports `does not match any known tools`.


crates/jp_cli/src/cmd/query.rs
```diff
@@ -2118,35 +2124,39 @@ fn apply_tool_use(
...
+        .unwrap_or_default();
+
+    let Some(tool) = partial.conversation.tools.tools.get_mut(name) else {
+        return Err(format!("tool choice '{name}' does not match any known tools").into());
     };
 
+    // A tool already in the desired state is a no-op, so a locked-*on* tool
+    // (e.g. a builtin) stays selectable. A locked-off one is refused with the
+    // lock-aware message.
+    apply_directive_to_tool(name, tool, &defaults, ToggleScope::Named, true)?;
```
JeanMertz: blocking: On a continuing conversation where `my_tool` is configured off, `jp q -u my_tool "..."` writes `state = true` into the same partial that `get_config_delta_from_cli` persists at the start of the turn. The next bare query silently inherits the enabled tool. A fresh conversation also stores this state in its base config. That contradicts `-u`'s “not persisted across queries” help and `-t`'s “duration of the query” help, and lets a one-shot forced tool remain available on later turns. Please apply the implied enable as an invocation-scoped overlay, or remove it from both fresh-conversation bases and continuing-conversation deltas before persistence. A two-query regression test should cover both paths.

crates/jp_cli/src/cmd/query.rs
```diff
@@ -2106,8 +2106,14 @@ fn apply_directive_to_tool(
...
 /// Apply the CLI tool use configuration to the partial configuration.
 ///
-/// NOTE: This has to run *after* `apply_enable_tools` because it will return an
-/// error if the tool of choice is not enabled.
+/// `-u NAME` implies `-t NAME`: forcing a specific tool also selects it, so a
+/// tool that is off by default never has to be named twice.
+/// A tool whose `allow_toggle` policy forbids a named enable is still refused,
+/// with the lock-aware message from [`apply_directive_to_tool`].
+///
+/// NOTE: This has to run *after* `apply_enable_tools`, so that `-u NAME` wins
```
JeanMertz: nit: The plausible conflict `jp q -T my_tool -u my_tool "..."` is a stated behavior of this PR, but the added tests never combine these flags. Please add a regression test that starts with an enabled tool, disables it through `tool_directives`, and verifies that `tool_use` re-enables and selects it. Otherwise reversing these two calls would break the precedence contract while every current test still passes.

crates/jp_cli/src/cmd/query_tests.rs
```diff
@@ -678,6 +678,99 @@ fn test_tool_use_accepts_locked_on_builtin() {
...
+            ..Default::default()
+        },
+        None,
+        make_partial_with_tools(),
+        None,
+    )
+    .unwrap_err()
+    .to_string();
+
+    assert!(err.contains("no_such_tool"), "unexpected error: {err}");
```
JeanMertz: nit: For `jp q -u no_such_tool`, this assertion also passes with the old “does not match any enabled tools” diagnostic, so it does not test the user-visible error change claimed by the PR. Assert the static full message, or at least require `does not match any known tools`, so the test distinguishes unknown tools from configured-but-disabled tools.

crates/jp_cli/src/cmd/query.rs
JeanMertz: Fixed, and the fix went further than the report in two directions.

Your diagnosis was right: the implied enable was written into the partial with `get_mut` and flowed through `get_config_delta_from_cli` into the conversation. What the report understates is that `-t NAME` already did exactly the same thing through the same call, so the behaviour was pre-existing and not specific to `-u`.

That mattered for the fix. Scoping it to `-u` alone would have made `-u NAME` behave differently from `-t NAME`, which is what the PR set out to unify. So we settled the question at the design level instead: the two flags are now deliberately asymmetric, and each says so in its help.

- `-t`/`-T` are sticky. Editing a config file is otherwise the only way to turn a tool on for a conversation, and `-T` makes it cheap to undo.
- `-u`/`-U` bind one query. There is no use for "force this tool on every future turn".

`-u` needs no config mutation at all to work. `tool_definitions` has kept a forced tool in the list since RFD 081 even while it is disabled (`jp_llm/src/tool.rs:1237-1241`), so `tool_choice` alone is enough. `apply_tool_use` became `validate_tool_use`: it takes `&PartialAppConfig`, writes nothing, and probes a clone through `apply_directive_to_tool` purely to keep the locked-off refusal. The choice is resolved at query time by `Query::effective_tool_choice`.

Two things beyond what you reported:

`assistant.tool_choice` itself was persisting too, which made `-u`'s own help line — *"this setting is **not** persisted across queries"* — false before this PR. Same fix covers it.

Removing the mutation broke MCP-backed tools, silently: `configure_active_mcp_servers` skips `!is_enabled()`, so the server would not start and `tool_definitions` would then drop the tool as unreachable. It now takes the forced tool name and starts its server regardless of enable state.

Guarded by `test_tool_use_does_not_persist_into_partial_config`, `test_no_tool_use_does_not_persist_into_partial_config`, and `test_tool_directive_does_persist_into_partial_config` — the last one pins the asymmetry on purpose, so a future "let's make these uniform" pass has to argue with a test. I confirmed the first goes red by re-adding the write: `left: Some(Function("explicitly_disabled_tool")), right: None`.

crates/jp_cli/src/cmd/query.rs
JeanMertz: Added as `test_tool_use_forces_a_tool_a_directive_disabled`, though it asserts the opposite of what you asked for — because the behaviour it was covering no longer exists.

Your reasoning was right: the precedence was stated in the PR description and in a doc-comment NOTE, enforced by nothing but the call order at `apply_cli_config`, and every added test left `tool_directives` at its default. Swapping the two calls would have passed the whole suite.

What changed is that `-u` no longer mutates the config, so the two flags stopped competing. `-T my_tool` disables durably, `-u my_tool` forces for this query through `tool_definitions`' forced-tool path, and both hold at once. There is no precedence left to get wrong, and the ordering NOTE is gone with it.

The test now pins both halves:

```rust
assert!(
    !effective(&partial, "implicitly_enabled_tool").state,
    "`-T` still persists the disable"
);
assert_eq!(
    query.effective_tool_choice(&AppConfig::new_test()),
    ToolChoice::Function("implicitly_enabled_tool".into()),
    "`-u` still forces the tool for this query"
);
```

The PR description has been rewritten; the old "`-u` wins over a `-T` in the same invocation" line was the thing your comment was testing against, and it is no longer accurate.

crates/jp_cli/src/cmd/query_tests.rs
JeanMertz: Fixed, and it turned up a third site.

Confirmed the assertion was vacuous with respect to the change: the old message was `tool choice '{v}' does not match any enabled tools` and the new one is `tool choice '{name}' does not match any known tools`, so `contains("no_such_tool")` held either way and the diagnostic this PR advertises went untested.

Now exact:

```rust
assert_eq!(
    err,
    "tool choice 'no_such_tool' does not match any known tools"
);
```

`test_tool_use_of_locked_off_tool_errors` had the same weakness in its two `contains` checks, so it is exact too. That surfaced a pre-existing one: `test_named_enable_of_locked_off_tool_errors` covers `-t` on a locked-off tool and asserts against the identical string, since both route through `apply_directive_to_tool`. Tightened it in the same pass — same assertion, same message, strictly more coverage. Happy to back that one out if you would rather keep the diff to `-u`'s own tests.

All three now pin the backtick quoting and the verb as well, which the `contains` form left free.

PR #936, opened 2026-08-04
fix(cli): Merge built-in tool config under user config

JeanMertz:
Built-in tool configs (currently `describe_tools`) were injected with `or_insert`, so any user block under the same tool name replaced them wholesale instead of layering on top. Setting one field was enough to lose everything else the built-in declares:

    [conversation.tools.describe_tools]
    result = "ask"

That left the tool with no `source`, so it no longer resolved as a built-in, and with no `enable`, `style`, or parameter schema.

The built-in block is now the lower-priority side of a partial merge, so user fields win per field and everything they do not mention is kept. Merging happens in place, which preserves each tool's position in the map, and that position is the order tools are presented to the provider.

Per-field merging also changes what a partial `enable` override means: `enable = { state = false }` on a built-in now keeps the built-in's `allow_toggle` rather than resetting it, so the tool becomes locked off rather than freely toggleable. The `enable = false` shorthand still resets `allow_toggle` to `any`, because it sets the field explicitly.

The fix is generic across every entry in `builtins::all()`, not specific to `describe_tools`.


crates/jp_cli/src/cmd/query.rs
```diff
@@ -2142,6 +2136,35 @@ fn apply_directive_to_tool(
...
+/// than replacing the whole block with a one-field partial.
+///
+/// Merging in place preserves each tool's position in the map, which is the
+/// order tools are presented to the provider.
+fn inject_builtin_tools(partial: &mut PartialAppConfig) -> BoxedResult<()> {
+    for (name, defaults) in tool::builtins::all() {
+        let tools = &mut partial.conversation.tools.tools;
+        match tools.get_mut(&name) {
+            Some(user) => {
+                let mut merged = defaults;
```
JeanMertz: blocking: A user can plausibly define a local or MCP tool named `describe_tools`, for example with `source = "local"` and a `command`. This branch merges the built-in's locked `enable`, hidden style, description, and `tools` parameter schema into that unrelated tool solely because the map key matches. The explicit `source` still wins, so JP executes the user's tool with built-in policy and schema applied silently. Preserve replacement behavior when the entry's source resolves to a different implementation, and merge these defaults only when `source` is absent or resolves to this built-in. Please cover an explicit local or MCP collision in a test.

crates/jp_cli/src/cmd/query_tests.rs
```diff
@@ -893,6 +893,143 @@ fn test_tool_use_forces_a_tool_a_directive_disabled() {
     );
 }
 
+#[test]
+fn test_builtin_config_merges_under_user_config() {
+    // A user overriding one field of a builtin keeps the rest of the builtin's
+    // block. Before the merge-under fix, the one-field partial replaced the
+    // whole entry and the tool lost its `source`, `enable`, and parameters.
```
JeanMertz: nit: These new test comments narrate patch history (`Before the merge-under fix`; later, `only became reachable once both halves landed`) and the final test refers to “the above.” That context goes stale and makes the tests depend on PR or RFD history. Please state each current invariant directly and keep each test's rationale self-contained, for example: “A one-field user override keeps the built-in source, enable policy, and parameters.”

crates/jp_cli/src/cmd/query_tests.rs
```diff
@@ -893,6 +893,143 @@ fn test_tool_use_forces_a_tool_a_directive_disabled() {
...
+        IntoPartialAppConfig::apply_cli_config(&Query::default(), None, partial, None).unwrap();
+
+    let names: Vec<&str> = partial
+        .conversation
+        .tools
+        .tools
+        .keys()
+        .map(String::as_str)
+        .collect();
+    assert_eq!(names, vec!["describe_tools", "zzz_last"]);
```
JeanMertz: nit: This exact-vector assertion makes the order test fail whenever `builtins::all()` gains another entry, even if `describe_tools` still stays before `zzz_last`. Since the injection helper is deliberately generic and more built-ins are planned, assert the relative indices of these two names, or filter `names` to the pair under test.

crates/jp_cli/src/cmd/query.rs
JeanMertz: The mechanics are right, and there's a sharper version of them than the one described: the built-in also sets `run: Some(RunMode::Unattended)` (`tool/builtins.rs:34`) while `RunMode::default()` is `Ask`, so a shadowing local command would run without a confirmation prompt. `parameters` merges per key (`merge_nested_indexmap`, `jp_config/src/util.rs:438`), so the required `tools` array lands in the user's schema too.

Not taking the fix, though: this is the specified behavior, not an oversight. RFD 083 Phase 1 step 3 is the design this PR implements, and it names exactly this case:

> Users *may* also override structural fields including `source`, `parameters`, and `command` — changing `source` shadows the built-in executor and routes through the selected source path instead. This is unsupported in the sense that the user owns the consequences [...] but it is not blocked. The reserved capability lives in the layering itself: built-in defaults always merge under, never replace.
>
> A partial override of `source` does not implicitly clear the built-in's other structural fields: the user's executor still runs against the built-in's `parameters` schema, `questions`, `style`, and `run`/`result` defaults unless those are explicitly overridden as well.

A source check would invert that decision and also special-case built-in injection against the rest of config layering, where the tools map is already merged key-wise with no source check. The RFD is still in Discussion, so the decision isn't ratified — but reversing it here would leave the code and the plan disagreeing; that argument belongs in 083.

Taking the test request, as a characterization test rather than a regression test: `test_builtin_config_merges_under_a_shadowing_local_tool` inserts `source = "local"` plus a `command` at the `describe_tools` key and pins that the user's `source` wins while `run`, `parameters`, and `enable` come from the built-in. It fails under the old `or_insert` behavior, so it's discriminating in both directions. Added a paragraph to the `inject_builtin_tools` doc comment stating the same, so the next reader doesn't have to derive it from the merge call.

crates/jp_cli/src/cmd/query_tests.rs
JeanMertz: Agreed, and applied. `test_builtin_config_merges_under_user_config` now reads "A one-field user override keeps the built-in source, enable policy, and parameters." — the suggested wording, verbatim.

Two phrases from the same batch still narrate history and want the same treatment: "only became reachable once both halves landed" on `test_table_disabled_builtin_cannot_be_forced`, and "The escape hatch for the above" on `test_bool_disabled_builtin_can_still_be_forced`. Replacements:

```rust
/// A table override sets only `state`, so the built-in's `allow_toggle = never`
/// survives and the tool is locked *off* rather than merely disabled. Forcing it
/// with `-u` is refused, which is the documented meaning of the table form
/// (RFD 083).
```

```rust
/// The `enable = false` shorthand writes `allow_toggle = any` explicitly, so it
/// overrides the built-in's toggle policy as well as its state and the tool
/// stays forceable with `-u`.
```

Keeping the RFD 083 pointer. Your comment groups "PR or RFD history" together, but 083 is the spec for what the table form means, not a note about what changed when — a reader who wants to know why `{ state = false }` behaves differently from `false` needs somewhere to go.

crates/jp_cli/src/cmd/query_tests.rs
JeanMertz: Correct, and the churn is concrete rather than hypothetical: `builtins::all()` returns one entry today, and RFD 083 Phase 2 adds `ask_user`, which would append after `zzz_last` and break the vector while the ordering invariant still held.

Took the filter over the relative-index version — one line, and the expected value stays a static vector, which is what the project's testing guidance asks for:

```rust
    .keys()
    .map(String::as_str)
    .filter(|name| matches!(*name, "describe_tools" | "zzz_last"))
    .collect();
assert_eq!(names, vec!["describe_tools", "zzz_last"]);
```

PR #944, opened 2026-08-05
fix(cli): Tailor no-target help to the failing command's grammar

JeanMertz:
When a conversation target can't be resolved, the error now reports
only the keywords and flags the failing command actually accepts.
`jp conversation unarchive` no longer suggests session keywords or
`--new` it doesn't support, and a single-target command no longer
advertises multi-select syntax like `+l, +live`. Declining the
interactive picker (Escape, or an empty multi-select) is now reported
as "No conversation selected." and is treated as an expected outcome
rather than a failure, since the picker appeared and the user made a
deliberate choice.

`Error::NoConversationTarget` gains `session`, `multi`, and
`allow_new` fields, and a new `TargetGrammar` type reads these
directly off each command's target argument so the help text can't
drift from what the command supports. Every picker helper
(`resolve_picker`, `resolve_archived_picker`, `resolve_multi_picker`,
etc.) now takes a `TargetGrammar` and threads it through to the error
path instead of returning one generic variant. Piping nothing into a
command that reads conversation IDs from stdin (`-`) now reports a
clearer, dedicated error instead of the generic no-target message.

The ubiquitous-language docs for the Conversation cluster gain
entries for the Live, Archived, and Active conversation states, which
the CLI error and help text already relied on.

### DIFF
--- a/crates/jp_cli/src/cmd/target.rs
+++ b/crates/jp_cli/src/cmd/target.rs
@@ -173,6 +173,49 @@ impl ConversationLoadRequest {
     }
 }
 
+/// What the command being resolved accepts as a target.
+///
+/// Carried down into the picker paths so that a resolution failure can print
+/// keyword help for that command, rather than for the union of every command's
+/// grammar.
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct TargetGrammar {
+    /// Session-based keywords (`s`, `+s`, `?s`, `.`) are accepted.
+    pub session: bool,
+
+    /// Several conversations are accepted, and `?` opens a multi-select.
+    pub multi: bool,
+
+    /// The command can start a fresh conversation, so the picker may offer it
+    /// and `--new` is worth suggesting when resolution fails.
+    pub allow_new: bool,
+}
+
+impl TargetGrammar {
+    /// Read the grammar off a command's target argument.
+    ///
+    /// `allow_new` is passed separately because the argument type can't know
+    /// it: whether a fresh conversation can be created is a property of the
+    /// command, not of what it accepts as a target.
+    pub fn from_args(args: &dyn ConversationIds, allow_new: bool) -> Self {
+        Self {
+            session: args.supports_session(),
+            multi: args.is_multi(),
+            allow_new,
+        }
+    }
+}
+
+impl From<TargetGrammar> for Error {
+    fn from(grammar: TargetGrammar) -> Self {
+        Self::NoConversationTarget {
+            session: grammar.session,
+            multi: grammar.multi,
+            allow_new: grammar.allow_new,
+        }
+    }
+}
+
 /// The outcome of resolving a [`ConversationLoadRequest`].
 #[derive(Default)]
 pub(crate) struct ResolveOutcome {
@@ -207,15 +250,13 @@ pub(crate) fn resolve_request(
         return Ok(ResolveOutcome::default());
     };
 
-    let (ids, start_new) = resolve_targets(
-        targets,
-        workspace,
-        session,
-        default_id,
-        request.multi,
-        request.session,
-        allow_picker_new,
-    )?;
+    let grammar = TargetGrammar {
+        session: request.session,
+        multi: request.multi,
+        allow_new: allow_picker_new,
+    };
+
+    let (ids, start_new) = resolve_targets(targets, workspace, session, default_id, grammar)?;
 
     let handles = ids
         .iter()
@@ -555,7 +596,10 @@ fn non_empty(ids: Vec<ConversationId>, empty: &'static str) -> Result<Vec<Conver
 /// Errors when stdin is a terminal (nothing piped) or when no IDs are provided.
 fn read_stdin_ids() -> Result<Vec<ConversationId>> {
     if io::stdin().is_terminal() {
-        return Err(Error::NoConversationTarget);
+        return Err(Error::NotFound(
+            "conversation",
+            "the '-' target reads IDs from stdin, but nothing is piped to it".into(),
+        ));
     }
 
     let mut ids = Vec::new();
@@ -588,12 +632,10 @@ fn resolve_targets(
     workspace: &Workspace,
     session: Option<&Session>,
     default_id: DefaultConversationId,
-    multi: bool,
-    supports_session: bool,
-    allow_new: bool,
+    grammar: TargetGrammar,
 ) -> Result<(Vec<ConversationId>, bool)> {
     if targets.is_empty() {
-        return resolve_from_session_or_picker(workspace, session, default_id, allow_new);
+        return resolve_from_session_or_picker(workspace, session, default_id, grammar);
     }
 
     let all_ids = targets
@@ -626,8 +668,8 @@ fn resolve_targets(
     for target in targets {
         if matches!(target, ConversationTarget::Help) {
             return Err(Error::TargetHelp {
-                session: supports_session,
-                multi,
+                session: grammar.session,
+                multi: grammar.multi,
             });
         }
 
@@ -640,19 +682,21 @@ fn resolve_targets(
 
                 // Archived picker draws from the archive partition.
                 if filter.archived {
-                    return if multi {
-                        resolve_archived_multi_picker(workspace, filter).map(|ids| (ids, false))
+                    return if grammar.multi {
+                        resolve_archived_multi_picker(workspace, filter, grammar)
+                            .map(|ids| (ids, false))
                     } else {
-                        resolve_archived_picker(workspace, filter).map(|id| (vec![id], false))
+                        resolve_archived_picker(workspace, filter, grammar)
+                            .map(|id| (vec![id], false))
                     };
                 }
 
-                if multi {
-                    return resolve_multi_picker(workspace, session, filter)
+                if grammar.multi {
+                    return resolve_multi_picker(workspace, session, filter, grammar)
                         .map(|ids| (ids, false));
                 }
 
-                return resolve_single_picker(workspace, session, filter, allow_new);
+                return resolve_single_picker(workspace, session, filter, grammar);
             }
             Ok(v) => return Ok((v, false)),
             Err(e) => last_err = Some(e),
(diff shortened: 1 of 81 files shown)


crates/jp_cli/src/cmd/target.rs
```diff
@@ -207,15 +250,13 @@ pub(crate) fn resolve_request(
...
-        targets,
-        workspace,
-        session,
-        default_id,
-        request.multi,
-        request.session,
-        allow_picker_new,
-    )?;
+    let grammar = TargetGrammar {
+        session: request.session,
```
JeanMertz: blocking: A normal `printf 'hello' | jp query` in a workspace with no active or default conversation still gets the wrong grammar here. Because `explicit_or_session_with_config()` delegates to `from_session_with_config()` when `ids()` is empty, `request.session` is hard-coded to `true` even though query uses `FlagIds<false, false>`. The resulting help advertises `.`, `s`, and `?s`, which query's parser rejects. The same empty-argument path resets `multi` to `false` for multi-target commands, so their help silently omits accepted `+...` forms. Please preserve `supports_session()` and `is_multi()` in the request even when the arguments are empty, and test the request path rather than only `TargetGrammar::from_args` in isolation.

crates/jp_cli/src/cmd/target.rs
```diff
@@ -733,28 +777,30 @@ pub(crate) fn resolve_picker(
...
 ) -> Result<ConversationId> {
     if !io::stdin().is_terminal() {
-        return Err(Error::NoConversationTarget);
+        return Err(grammar.into());
     }
 
     pick_conversation(workspace, session, filter, false)
         .and_then(PickSelection::existing)
-        .ok_or(Error::NoConversationTarget)
+        .ok_or(Error::NoConversationSelected)
```
JeanMertz: nit: `jp c show ?pinned` when no conversations are pinned reaches this branch without showing a picker: `pick_conversation` returns `None` immediately when its item list is empty. Reporting that as “No conversation selected” says the user declined a prompt they never saw and marks the run expected. Keep “no candidates” distinct from prompt cancellation, for example by returning `NotFound` for the filtered set and reserving `NoConversationSelected` for an actual cancelled or empty submission.

crates/jp_cli/src/cmd/target.rs
```diff
@@ -1001,21 +1048,22 @@ pub(crate) fn resolve_archived_picker(
     }
     let selected = prompt
         .prompt_with_writer(&mut writer)
-        .map_err(|_| Error::NoConversationTarget)?;
+        .map_err(|_| Error::NoConversationSelected)?;
```
JeanMertz: blocking: If the terminal disconnects while this picker is active, `prompt_with_writer` can return an I/O error. This discards that source, reports a deliberate selection decline, and marks the run expected, which suppresses the trace-log diagnostic for the actual failure. The inquiry coordinator already treats only `OperationCanceled` and `OperationInterrupted` as user-initiated; picker resolution should make the same distinction and propagate other errors through `Error::Inquire`. The live and multi-picker helpers also need to retain the prompt error rather than erasing it through `Option` or an empty `Vec`.

docs/architecture/ubiquitous-language/conversation.md
```diff
@@ -2,11 +2,14 @@
...
 The Conversation cluster covers JP's central abstraction: the persistent record
 of "talking to the assistant" and the event log that backs it.
-Six terms work together here — a **Conversation** is the stored entity, a
+Six terms describe its structure — a **Conversation** is the stored entity, a
 **Turn** is one slice of it, an **Event** is the atomic unit inside a Turn,
 **Tool Calls** and **Inquiries** are specific event kinds, and a **Thread** is
 the projection of a Conversation that gets sent to an LLM provider.
 
+Three more name a Conversation's *state*: **Live** and **Archived** are where it
+sits, and **Active** is the one a session is working on.
```
JeanMertz: nit: This repeats the state introduction at lines 17–20, which already explains `Active`, `Live`, `Archived`, and `Label`. Keeping both makes the cluster opening introduce the same axis twice. Please consolidate them into one paragraph, including `Label` there if the structural summary is meant to cover the whole cluster.

docs/features/conversation-search.md
```diff
@@ -178,7 +178,8 @@ jp c grep --scope user,assistant 'retry'
 Searching `title` alone never reads the event streams, so it stays fast across a
 large workspace.
 
-Every conversation in the workspace is searched unless `--id` narrows it:
+Every live conversation — everything outside the archive — is searched unless
+`--id` narrows it:
```
JeanMertz: nit: The Conversation glossary defines **Live Conversation** as a conversation in the default storage partition and explicitly says to avoid defining it by negation. “Everything outside the archive” is the same negated definition and becomes wrong if another location is added. “Every live conversation is searched unless `--id` narrows it” is enough here, or use “every conversation in the default storage partition” if a gloss is needed.

crates/jp_cli/src/cmd/target.rs
JeanMertz: Confirmed, and it's wider than the two constructors you named — all four `explicit_or_*` helpers drop the args in their empty branch, because each one delegates to a constructor that hard-codes `session: true, multi: false`. `explicit_or_previous` does it too, via `Self::explicit(vec![SessionPrevious, Picker])`.

Traced your exact case: `query` is `FlagIds<false, false>` (`query.rs:176`) and uses `explicit_or_session_with_config` (`query.rs:757`), so a bare invocation produced `NoConversationTarget { session: true, multi: false }` and the help offered `.`, `active`, `s`, `session`, `?s`, `?session` — every one rejected by `TargetParser<false>` via `requires_session()`. The multi half bites `jp c print` (`PositionalIds<true, true>`), which lost its whole Multi-Target Keywords block despite accepting `+p`, `+s`, and `-`.

Fixed by hoisting the assignment out of the branch in all four:

```rust
let mut req = if args.ids().is_empty() {
    Self::from_session()
} else {
    Self::explicit(args.ids().to_vec())
};

req.multi = args.is_multi();
req.session = args.supports_session();
req
```

On testing the request path: I asserted on the **request** rather than driving `resolve_request`. Reaching the failure needs the picker to be unreachable, and `stdin` may still be a terminal in a test binary — `inquire` would block on a real prompt rather than fail, which is a hang instead of a red. `an_empty_target_list_keeps_the_commands_grammar` builds an empty `FlagIds<false, false>` and an empty `PositionalIds<true, true>`, runs both through all four constructors, and asserts the flags survive. I watched it fail on your scenario first ("invented session keyword support"). `set_tests.rs:321` and `archive_tests.rs:106` already covered `conversation_load_request()` but never these two fields.

crates/jp_cli/src/cmd/target.rs
JeanMertz: Right on the substance, and the archived siblings already did it your way — `resolve_archived_picker` builds items, checks `is_empty()`, and returns `NotFound` before prompting. So this was an inconsistency inside one file rather than a new idea.

One correction to the example: `jp c show` is `PositionalIds<true, true>` (`show.rs:18`), so `?pinned` routes through `grammar.multi` to `resolve_multi_picker`, not to the line you anchored on. The reachable case here is a single-target command — `jp c use ?pinned` (`PositionalIds<true, false>`). The defect was real at both, though: `pick_conversations` returned an empty `Vec` for an empty candidate set, which `resolve_multi_picker` then read as a decline.

`pick_conversation` and `pick_conversations` now return `Result` and error with `NotFound` when the filter matches nothing, naming the set that came up empty:

```rust
fn no_candidates(filter: &PickerFilter) -> Error {
    let missing = if filter.pinned {
        "no pinned conversations"
    } else if filter.session {
        "no conversations in this session"
    } else if filter.candidate_ids.is_some() {
        "no conversations match"
    } else {
        "no conversations exist"
    };

    Error::NotFound("conversation", missing.into())
}
```

I dropped a `query`-based arm I'd first written: `build_picker_items` filters on `candidate_ids`, `pinned`, and `session` only — `filter.query` is just the prompt's starting input and never affects emptiness, so that branch would have been dead.

`NoConversationSelected` is now reserved for a cancelled prompt or an empty submission, which is the split you asked for. Tested directly against both helpers (they return before constructing a prompt, so no terminal is involved) with an exact-message assertion, so the `pinned` branch is pinned rather than the fallback.

crates/jp_cli/src/cmd/target.rs
JeanMertz: Accepted. This was the one place in the PR where the damage genuinely hid: a real I/O fault rendered as "No conversation selected." with `expected: true`, and by that field's own contract that keeps failure-only diagnostics quiet.

Four sites discarded the error, not one: this `map_err(|_| ...)`, `pick_conversation`'s `.ok()?`, `pick_conversations`' `let Ok(..) else { return vec![] }`, and `resolve_archived_multi_picker`'s `let Ok(..) else`. All four now go through one classifier modelled on the `prompt_cancellation_reason` precedent you pointed at:

```rust
fn declined_or_failed(error: InquireError) -> Error {
    match error {
        InquireError::OperationCanceled | InquireError::OperationInterrupted => {
            Error::NoConversationSelected
        }
        error => Error::Inquire(error),
    }
}
```

`Error::Inquire(#[from] InquireError)` already existed (`error.rs:155`), so the cause propagates and the run reports as broken.

Your second point — retaining the error through `Option` / empty `Vec` — needed `pick_conversation` and `pick_conversations` to return `Result`, which is the same change comment #3851958004 needed to separate "no candidates" from "declined". Landed as one refactor.

Test mirrors `coordinator_tests.rs:601-613` in shape. I checked it was load-bearing by reverting `declined_or_failed` to the old blanket mapping and watching the `NotTTY` case fail.

docs/architecture/ubiquitous-language/conversation.md
JeanMertz: Correct, and it's a rebase artifact — my paragraph predates the one now at lines 14-17, and both survived.

Consolidated by deleting mine rather than merging them. The existing paragraph is the better of the two: it names **Label**, which mine omitted, and it draws the where-it-sits versus how-it's-annotated split explicitly.

Kept the line 5 rewording ("Six terms describe its structure"). It's more accurate with only one state paragraph following, since it scopes "six terms" to structure and leaves state to that paragraph — which is what made the duplication visible in the first place.

docs/features/conversation-search.md
JeanMertz: Fair — the PR contradicted a glossary entry it adds in the same diff. "Everything outside the archive" is exactly the negation that the **Live Conversation** entry's `**Avoid.**` line rules out, for the reason it gives: it stops working as more locations appear.

Took your first suggestion:

```md
Every live conversation is searched unless `--id` narrows it:
```

No gloss needed at all, as it turns out — eleven lines below, the same section already says "`+l` (long form `+live`) is every live conversation" without one.

PR #949, opened 2026-08-07
build(justfile): Add `git-merge` recipe for AI-assisted conflicts

JeanMertz:
Add `just git-merge REF` to drive a merge and, when it stops on conflicts, hand them to `jp` for resolution. A clean merge just reports and exits. On conflicts, it prompts to resolve in a `[p]icked` conversation, a `[n]ew` one (seeded with `personas/dev`), or `[q]uit`, then sends the merge output and `git status -sb` as the query. The conflicted files are left in the working tree for `jp` to edit; the recipe never commits or runs `git merge --abort`.

Re-running `just git-merge` while a merge is already in progress skips the merge step and hands the outstanding conflicts over again, so a resolution that went sideways can be retried in a fresh conversation.

Example usage:

    just git-merge main
    just git-merge origin/main --edit

### DIFF
--- a/justfile
+++ b/justfile
@@ -128,11 +128,17 @@ stage-and-commit: _install-jp
 # prompts where to resolve them (a [p]icked conversation, a [n]ew one, or
 # [q]uit) and sends the merge output plus `git status -sb` as the query. The
 # conflicted files stay in the working tree; jp edits them in place, and does
-# not commit or abort the merge.
+# not commit or abort the merge. Exits nonzero, listing what is left, when
+# conflicts remain after the query.
 #
 # Re-running while a merge is already in progress skips the merge itself and
 # hands the outstanding conflicts over again, so a resolution that went
-# sideways can be retried in another conversation.
+# sideways can be retried in another conversation. That merge's source is read
+# from MERGE_HEAD, not from REF, so the retry needs no arguments.
+#
+# Refuses to start when unmerged paths exist without a merge in progress: a
+# rebase, cherry-pick, or revert stopped on conflicts is not this recipe's to
+# resolve.
 #
 # A picked conversation keeps its own config; it usually already holds the
 # branch's context. A new one gets `personas/dev`, since the workspace default
@@ -148,18 +154,27 @@ git-merge REF="main" *ARGS: _install-jp _install-tools
     #!/usr/bin/env sh
     set -eu
 
+    ref="$1"
     shift # remove REF from positional params
 
     if git rev-parse -q --verify MERGE_HEAD >/dev/null 2>&1; then
-        merge_out="A merge is already in progress; 'git merge {{REF}}' was not re-run."
+        # This invocation's REF says nothing about the merge already underway;
+        # MERGE_HEAD is the commit actually being merged.
+        ref=$(git name-rev --name-only --always MERGE_HEAD)
+        merge_block="(a merge with ${ref} was already in progress; no new merge was started)"
         merge_exit=0
-        printf '%s\n' "$merge_out" >&2
+        printf '%s\n' "$merge_block" >&2
+    elif [ -n "$(git diff --name-only --diff-filter=U)" ]; then
+        echo "Unmerged paths exist but no merge is in progress." >&2
+        echo "Finish or abort the rebase/cherry-pick/revert that stopped first." >&2
+        exit 1
     else
         set +e
-        merge_out=$(git merge "{{REF}}" 2>&1)
+        merge_out=$(git merge "$ref" 2>&1)
         merge_exit=$?
         set -e
         printf '%s\n' "$merge_out"
+        merge_block=$(printf '$ git merge %s\n%s' "$ref" "$merge_out")
     fi
 
     # Unmerged paths, not the exit status, decide whether there's work for jp:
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -122,6 +122,86 @@ stage-and-commit: _install-jp
...
+        merge_out=$(git merge "{{REF}}" 2>&1)
+        merge_exit=$?
+        set -e
+        printf '%s\n' "$merge_out"
+    fi
+
+    # Unmerged paths, not the exit status, decide whether there's work for jp:
+    # a merge can also fail for reasons it can't help with (dirty tree, unknown
+    # ref), and an in-progress merge may already be fully resolved.
+    conflicts=$(git diff --name-only --diff-filter=U)
```
JeanMertz: blocking: If the user runs `just git-merge main` while a rebase or cherry-pick is stopped on conflicts, `git merge` fails without creating `MERGE_HEAD`, but those pre-existing `U` entries make this branch continue. JP is then told those files are conflicts with `main` and can rewrite an unrelated operation under false context. Check for unmerged entries before invoking `git merge` and fail unless `MERGE_HEAD` was already present; only conflicts created by this invocation should enter the resolver.

justfile
```diff
@@ -122,6 +122,86 @@ stage-and-commit: _install-jp
...
+[group('jp')]
+[positional-arguments]
+git-merge REF="main" *ARGS: _install-jp _install-tools
+    #!/usr/bin/env sh
+    set -eu
+
+    shift # remove REF from positional params
+
+    if git rev-parse -q --verify MERGE_HEAD >/dev/null 2>&1; then
+        merge_out="A merge is already in progress; 'git merge {{REF}}' was not re-run."
```
JeanMertz: blocking: The documented retry path can identify the merge incorrectly. After `just git-merge origin/main`, rerunning plain `just git-merge` supplies the default `main`; `MERGE_HEAD` skips the new merge, but the output, title, and prompt all claim the existing merge is with `main`. JP can then compare against the wrong branch and write a plausible but incorrect resolution. Derive the source of the existing operation if that can be done reliably, or use source-neutral wording on this path rather than treating this invocation's `REF` as the existing merge's identity.

justfile
```diff
@@ -122,6 +122,86 @@ stage-and-commit: _install-jp
...
+    preamble="Please resolve the merge conflicts with {{REF}}. Read each conflicted file, \
+    work out what both sides were after, and write the resolution that keeps both intents. \
+    Where the two sides genuinely disagree, stop and tell me instead of picking one. Leave \
+    the merge uncommitted and do not run 'git merge --abort'."
+
+    printf '%s\n\n```sh\n$ git merge %s\n%s\n\n$ git status -sb\n%s\n```\n' \
+        "$preamble" "{{REF}}" "$merge_out" "$(git status -sb)" \
+        | jp query $target $persona "$@"
+
+    printf "\nReview the resolution, then 'git commit' to conclude the merge.\n" >&2
```
JeanMertz: nit: When JP follows the prompt's instruction to stop on a genuine disagreement, `jp query` can exit successfully while `U` paths remain. The recipe then prints this commit instruction and exits zero even though `git commit` will refuse. The failure is visible and stays in this merge, but the final status is misleading. Recheck `git diff --name-only --diff-filter=U` after the query, report any remaining paths, and return nonzero; print this line only when the index is clear.

justfile
```diff
@@ -122,6 +122,86 @@ stage-and-commit: _install-jp
...
+
+    shift # remove REF from positional params
+
+    if git rev-parse -q --verify MERGE_HEAD >/dev/null 2>&1; then
+        merge_out="A merge is already in progress; 'git merge {{REF}}' was not re-run."
+        merge_exit=0
+        printf '%s\n' "$merge_out" >&2
+    else
+        set +e
+        merge_out=$(git merge "{{REF}}" 2>&1)
```
JeanMertz: theoretical: I could not name an ordinary ref likely to trigger this, but `REF` is arbitrary text and valid Git refnames may contain double quotes, dollar signs, or backticks. Just substitutes `{{REF}}` before the shell parses this script, so the surrounding quotes do not preserve such a value; a crafted ref can break the command or execute shell syntax. Capture `ref=$1` before `shift` and use `"$ref"` in the command and generated text, matching the positional-argument approach used elsewhere in this justfile.

justfile
JeanMertz: Confirmed and fixed. `{{REF}}` was used in five places and nothing on the resume path re-derived it, so `just git-merge` after `just git-merge origin/main` produced a prompt, a fenced `$ git merge` header, and a `merge:main-into-<branch>` title that all named a branch nobody merged.

There was a second defect on the same path: the fenced block presented `merge_out` as the output of `git merge`, but on the resume path `merge_out` was the recipe's own sentence, so the transcript handed to JP was fabricated.

`MERGE_HEAD` is the commit actually being merged, so the identity is derivable:

```sh
ref=$(git name-rev --name-only --always MERGE_HEAD)
merge_block="(a merge with ${ref} was already in progress; no new merge was started)"
```

`--name-only --always` yields `main`, `remotes/origin/main`, or a short sha as fallback, so there is no `undefined` case to handle. `$ref` now feeds the preamble and the title, and `$merge_block` replaced the invented `$ git merge %s\n%s` portion of the query, so the resume path states what it is instead of imitating a command that never ran.

justfile
JeanMertz: Applied, though not on the injection reasoning. The mechanism you describe is right (just substitutes before the shell parses, and `check-ref-format` does permit `"`, `$`, and backticks), but the threat model here is the user's own repo and their own branch names, so as an attack it is self-inflicted and would not gate the merge on its own. Your `theoretical:` label is the correct read.

It rides along anyway: the `MERGE_HEAD` fix above needs a single variable to carry the derived identity, and that variable has to be quoted. So `ref="$1"` before `shift` plus `"$ref"` at all five sites is now in.

One thing I checked before taking it, since it would have broken the suggestion: `Evaluator::evaluate_parameters` in `casey/just` evaluates a missing argument's `default` and pushes the result into `positional`, so `just git-merge` with no arguments gives `$1 == "main"` and the existing `shift` is safe.

The same pass fixed a closer relative of your point that the diff had elsewhere: `$target` and `$persona` were expanded unquoted into `jp query`, and `name-rev` output can contain `~` or `^`. Both variables are gone in favour of prepending to the positional params, which keeps every argument quoted:

```sh
n|N) set -- --new --title "merge:${ref}-into-${branch}" --cfg=personas/dev "$@" ;;
```

justfile
JeanMertz: Agreed, and this was the one I'd have held the merge on. A conflicted rebase sets `REBASE_HEAD`, cherry-pick sets `CHERRY_PICK_HEAD`, revert sets `REVERT_HEAD`; none set `MERGE_HEAD`. `git merge` then refuses without creating it, so the guard read the other operation's unmerged entries and proceeded. `git stash pop` and `git checkout -m` conflicts reach the same branch.

The damage also spreads rather than staying put: JP rewrites files inside a rebase whose conflict sides are not the ones it was told about, and `git rebase --continue` then commits whatever it wrote onto the wrong commit.

Checking before the merge, as you suggested:

```sh
elif [ -n "$(git diff --name-only --diff-filter=U)" ]; then
    echo "Unmerged paths exist but no merge is in progress." >&2
    echo "Finish or abort the rebase/cherry-pick/revert that stopped first." >&2
    exit 1
```

Phrased against unmerged entries rather than enumerating operations, so it covers every stopped operation instead of the two named. With this in place, conflicts reaching the resolver on the fresh path are necessarily the ones this invocation created.

justfile
JeanMertz: Taken, and your severity read is right: the damage is contained (the merge was already stopped, nothing new is corrupted) and visible (JP says it stopped, `git commit` refuses loudly), so this is a suggestion rather than a gate. Worth doing for two reasons that sit slightly beside the misleading-status one: the recipe couldn't be chained honestly, since `just git-merge main && ...` reported success on an unresolved merge, and listing the remaining paths is useful output in its own right.

```sh
remaining=$(git diff --name-only --diff-filter=U)
if [ -n "$remaining" ]; then
    printf "\nStill conflicted:\n" >&2
    printf '%s\n' "$remaining" | sed 's/^/  /' >&2
    exit 1
fi
```

One point in the check's favour that isn't obvious: `--diff-filter=U` reports index-level unmerged entries, so a file JP edited but never staged still counts as conflicted, which is exactly what `git commit` refuses on. The check and the message stay aligned.

I left `[no-exit-message]` off, so just will print its own `error: Recipe 'git-merge' failed...` line on this path. Suppressing it would also mute genuine failures; say the word if you'd rather have the quieter version.

PR #953, opened 2026-08-10
fix(workspace, storage): Keep unwritten conversations across a reload

JeanMertz:
`load_conversation_index` rebuilds the index from a store scan, and a scan reports only what the store holds. A conversation created in memory has been written nowhere, so the scan cannot see it and the rebuild drops it. A long-running host reloads on every request, which makes this reachable: a conversation created to serve one request is gone by the next.

Keeping everything the scan omits is not the answer, because a conversation another process deleted is equally absent, and that one has to be forgotten. The two are only distinguishable if the difference was recorded while the conversation was still in hand, so creation records `StoragePresence::Unwritten` rather than the durable presence its projection maps to. Absent and unwritten is kept; absent and durable is dropped.

`Unwritten` carries the `Projection` it was created with, since there is no stored copy to derive one from, and that intent is what decides whether a first write lands in the workspace, in user-local storage, or both.

`jp conversation ls` and `jp conversation show` decided whether a conversation was local by comparing its presence to `UserLocalOnly`. They go through `Projection` instead, so a conversation not yet on disk still reports the root it is destined for.

### DIFF
--- a/crates/jp_cli/src/cmd/conversation/ls.rs
+++ b/crates/jp_cli/src/cmd/conversation/ls.rs
@@ -2,7 +2,7 @@ use chrono::{DateTime, FixedOffset, Local, Utc};
 use comfy_table::{Cell, CellAlignment, Row};
 use crossterm::style::{Color, Stylize as _};
 use jp_conversation::{Conversation, ConversationId};
-use jp_storage::backend::StoragePresence;
+use jp_storage::backend::{Projection, StoragePresence};
 use jp_term::{
     osc::hyperlink,
     table::list,
@@ -185,7 +185,10 @@ impl Ls {
             workspace
                 .archived_conversations()
                 .filter_map(|(id, c, presence)| {
-                    let local = presence == StoragePresence::UserLocalOnly;
+                    // Derived through `Projection` rather than compared: a
+                    // conversation not yet written is still destined for one root
+                    // or both, and the indicator should say which.
+                    let local = Projection::from(presence) == Projection::LocalOnly;
                     let external = presence == StoragePresence::WorkspaceOnly;
                     matches_filters(&id, &c, local).then(|| to_details(id, &c, local, external))
                 })
@@ -195,7 +198,7 @@ impl Ls {
                 .conversations()
                 .filter_map(|(id, c)| {
                     let presence = workspace.conversation_presence(id);
-                    let local = presence == Some(StoragePresence::UserLocalOnly);
+                    let local = presence.map(Projection::from) == Some(Projection::LocalOnly);
                     let external = presence == Some(StoragePresence::WorkspaceOnly);
                     matches_filters(id, &c, local).then(|| to_details(*id, &c, local, external))
                 })
--- a/crates/jp_cli/src/cmd/conversation/ls.rs
+++ b/crates/jp_cli/src/cmd/conversation/ls.rs
@@ -2,7 +2,7 @@ use chrono::{DateTime, FixedOffset, Local, Utc};
 use comfy_table::{Cell, CellAlignment, Row};
 use crossterm::style::{Color, Stylize as _};
 use jp_conversation::{Conversation, ConversationId};
-use jp_storage::backend::{Projection, StoragePresence};
+use jp_storage::backend::StoragePresence;
 use jp_term::{
     osc::hyperlink,
     table::list,
@@ -185,10 +185,7 @@ impl Ls {
             workspace
                 .archived_conversations()
                 .filter_map(|(id, c, presence)| {
-                    // Derived through `Projection` rather than compared: a
-                    // conversation not yet written is still destined for one root
-                    // or both, and the indicator should say which.
-                    let local = Projection::from(presence) == Projection::LocalOnly;
+                    let local = presence == StoragePresence::UserLocalOnly;
                     let external = presence == StoragePresence::WorkspaceOnly;
                     matches_filters(&id, &c, local).then(|| to_details(id, &c, local, external))
                 })
@@ -198,7 +195,7 @@ impl Ls {
                 .conversations()
                 .filter_map(|(id, c)| {
                     let presence = workspace.conversation_presence(id);
-                    let local = presence.map(Projection::from) == Some(Projection::LocalOnly);
+                    let local = presence == Some(StoragePresence::UserLocalOnly);
                     let external = presence == Some(StoragePresence::WorkspaceOnly);
                     matches_filters(id, &c, local).then(|| to_details(*id, &c, local, external))
                 })
--- a/crates/jp_storage/src/backend/load.rs
+++ b/crates/jp_storage/src/backend/load.rs
@@ -5,6 +5,7 @@ use std::fmt::Debug;
 use chrono::{DateTime, Utc};
 use jp_conversation::{Conversation, ConversationId, ConversationStream};
 
+use super::Projection;
 use crate::{LoadError, validate::ValidationError};
 
 /// Controls which storage partition to scan.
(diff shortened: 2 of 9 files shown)


crates/jp_workspace/src/lib.rs
```diff
@@ -197,21 +198,56 @@ impl Workspace {
...
+        // found.
+        //
+        // Keeping them is safe because the distinction is decidable: the presence
+        // recorded at creation says which of the two an absent conversation is,
+        // and a scan that *does* find one is proof it has since been written.
+        let unwritten: Vec<ConversationId> = self
+            .state
+            .presence
+            .iter()
+            .filter(|(_, presence)| !presence.is_durable())
```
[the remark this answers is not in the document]
JeanMertz: Fixed, and you traced it exactly.

Confirmed before changing anything: `Unwritten` was written in one place (`create_conversation_with_projection`) and cleared in one place (`load_conversation_index` rebuilding from scan entries). Nothing on the write path touched it, and `ConversationMut` has no route back to the `Workspace` — it holds `metadata`, `events`, `writer`, `projection` and a non-shared `dirty` flag.

So the comment I wrote at the filter — "a scan that *does* find one is proof it has since been written" — is true but only fires if a scan happens between the flush and the delete. Nothing guarantees that. My reasoning covered create→scan and never enumerated create→flush→delete→scan.

Worth noting the window is wider than "until the next reload": `refresh_conversations` runs per plugin request, so for an idle `serve-web` the marking persists until the next request arrives. A delete landing in that gap resurrects the conversation, and a later mutation flushes it back to disk — the user's delete silently undone.

The fix records durability where the write happens. `State` gains `written: HashMap<ConversationId, Arc<AtomicBool>>`, created `false` at conversation creation and handed to every lock; `ConversationMut::flush` sets it *after* a successful `writer.write(...)`, so a failed write leaves it unwritten. The reload filter asks that instead of the presence variant. This keeps the lock owning itself (RFD 020) — no `&mut Workspace` needed, which the delegated-turn path depends on.

Also fixed the reload to carry the cell across for scanned entries, so a lock still held on one keeps writing to the cell the workspace reads.

The test is replaced rather than kept alongside, since you're right that the manual `presence.insert(id, Projected)` asserted the merge while skipping the path that records the state — which is exactly where the bug was. `an_index_reload_forgets_a_written_conversation_that_is_gone` now runs the real sequence: create → lock → mutate → `flush()` → `fs.remove(&id)` as another process would → reload → assert gone.

I checked it fails without the fix before keeping it — commenting out the one `written.store(true, ...)` line reproduces the resurrection:

```
thread '...forgets_a_written_conversation_that_is_gone' panicked:
a conversation this process wrote and someone else deleted is forgotten
```

`jp_workspace` 108/108, `jp_storage` 136/136, `jp_cli` 1029/1029.

crates/jp_storage/src/backend/load.rs
```diff
@@ -29,6 +30,21 @@ pub enum StoragePresence {
...
+    /// time a scan finds it, which is also the first moment it is known to be
+    /// durable.
+    ///
+    /// It carries the projection it was created with, so the write intent
+    /// survives until there is a stored presence to derive one from.
+    ///
+    /// Distinguishing this from a durable presence is what lets an index reload
+    /// tell "created here, not saved yet" apart from "deleted by someone else":
+    /// both are absent from the scan, and only one should be forgotten.
+    Unwritten(Projection),
```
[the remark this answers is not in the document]
JeanMertz: Agreed, and `jp_storage` is now untouched by this PR — `load.rs` and `persist.rs` are back to their original blobs.

You were quoting RFD 031 accurately, and my own doc comment on the variant conceded the contradiction outright: "Never produced by a scan: a store cannot report something it does not hold." A variant unreachable through the type's only producer is a boundary in the wrong place, and `Projection::from` had grown an arm existing solely to service it.

The reason it resolved itself rather than needing its own refactor: fixing the blocking issue above required a durability flag the workspace owns and the lock writes, and once that exists `Unwritten` has no remaining job. So creation records `StoragePresence::from(projection)` again, the variant and `is_durable()` are gone, and "have we written this" lives in `jp_workspace` where the answer is known. Two symptoms of the same misplacement, one change — doing them separately would have left two representations of the same fact to drift apart.

I went with the shared flag rather than your `Unwritten(Projection) | Stored(StoragePresence)` shape because it needs no second enum and no second `Projection` derivation path, but both satisfy the objection and I'd have taken yours if the flag hadn't fallen out of the other fix.

Two follow-ons worth flagging since they're no longer strictly required:

- `ls.rs` and `show.rs` deriving `local` via `Projection::from` rather than comparing to `UserLocalOnly` — I've kept it deliberately, it reads better and is robust to future variants, but say the word if you'd rather this PR not carry it.
- `conversation_presence_reflects_creation_intent` is back to its original assertions.

crates/jp_workspace/src/conversation_lock.rs
```diff
@@ -467,6 +483,11 @@ impl ConversationMut {
         self.writer.write(&self.id, &meta, &evts, self.projection)?;
         self.dirty.store(false, Ordering::Relaxed);
 
+        // After the write, never before: a failed write leaves the conversation
+        // as unwritten as it was, and claiming otherwise would have a later
+        // index reload treat its absence as somebody else's deletion.
+        self.written.store(true, Ordering::Relaxed);
```
JeanMertz: blocking: A user running a long-lived host with `--no-persist` can create and mutate an ephemeral conversation, then lose it on the next index reload. `NullPersistBackend::write` deliberately returns `Ok(())` while discarding the write, so this sets `written = true`; the filesystem scan remains empty, and `load_conversation_index` drops the only in-memory copy. The same issue exists in the drop path at line 555. Please make the write result distinguish persisted from discarded, or keep the flag false when persistence is disabled, and add a `NullPersistBackend -> mutate -> reload` regression test.

crates/jp_workspace/src/lib.rs
```diff
@@ -353,25 +357,83 @@ impl Workspace {
...
+        //
+        // Keeping them is safe because the distinction is decidable, but only by
+        // asking whether *we* have written it: a conversation that was written and
+        // then deleted by another process is also absent, and that one has to be
+        // forgotten. The flag is set by whatever wrote it, under the lock.
+        let unwritten: Vec<ConversationId> = self
+            .state
+            .written
+            .iter()
+            .filter(|(_, written)| !written.load(atomic::Ordering::Relaxed))
```
JeanMertz: nit: A long-lived host can still omit a new conversation for one request when its first flush races this reload. The scan above can return before the write creates its files; if `flush()` then sets the flag before this load, the ID is absent from `entries` but fails the unwritten filter, so the rebuild drops it even though the write succeeded. `ConversationMut` is `Send + Sync` and owns independent `Arc`s, so this interleaving is allowed. Snapshot the false candidates before the scan, or synchronize the scan and rebuild with first writes; a blocking backend can make the regression test deterministic.

crates/jp_cli/src/cmd/conversation/ls.rs
```diff
@@ -185,7 +185,10 @@ impl Ls {
             workspace
                 .archived_conversations()
                 .filter_map(|(id, c, presence)| {
-                    let local = presence == StoragePresence::UserLocalOnly;
+                    // Derived through `Projection` rather than compared: a
+                    // conversation not yet written is still destined for one root
+                    // or both, and the indicator should say which.
+                    let local = Projection::from(presence) == Projection::LocalOnly;
```
JeanMertz: theoretical: I cannot name a current input that changes the output because the three `StoragePresence` variants make this equivalent to `presence == StoragePresence::UserLocalOnly`. The conversion still crosses the contract defined by RFD 031: `StoragePresence` is the read-side filesystem fact that defines `local`, while `Projection` is future write intent. An archived entry always comes from a scan, so the preceding “not yet written” explanation is impossible here. Keep the direct presence comparison in `ls` and `show`; a future storage variant should decide locality explicitly rather than inherit it through write projection.

PR #983, opened 2026-08-15
fix(cli, storage, workspace): Keep sessions attached across `-!`

JeanMertz:
A `--no-persist` run swapped the persist and lock backends for their Null variants but left the session backend pointing at real storage, while `jp query` records the session's active conversation unconditionally. An ephemeral query therefore wrote a conversation it never persisted to the front of the session history, and the next `jp q` in that terminal opened the conversation picker instead of resuming.

Wrapping the session backend in `ReadOnlySessionBackend` closes the leak at its source: `jp -! q` still reads which conversation the session is on, but can no longer record one.

`session_active_conversation` also read `history[0]` and nothing else, so a single unresolvable entry detached the session even with the real conversation directly behind it. It now walks the history and returns the first entry that still resolves, skipping conversations that exist in neither the live index nor the archive. The walk stops at an archived entry so archiving stays a deliberate act — the session is left without a target and the caller opens the picker, rather than being moved somewhere the user did not ask to go.

Two nearby defects in `cleanup_stale_files` fall out of the same mismatch. It read lock state from the workspace's lock backend, which reports nothing locked under `--no-persist`, letting an ephemeral run delete another process's session mapping while its conversation was mid-write; it now reads from the filesystem backend it is already handed. And it scanned only the live partition, so archiving a conversation pruned it from every session's history and unarchiving restored a conversation the session had forgotten; the archive partition now counts as live for pruning.


crates/jp_workspace/src/session_mapping.rs
```diff
@@ -94,25 +89,73 @@ impl Workspace {
...
+    /// See [`resolvable_history`] for which entries count.
+    ///
+    /// [`resolvable_history`]: Self::resolvable_history
     #[must_use]
     pub fn session_active_conversation(&self, session: &Session) -> Option<ConversationId> {
-        self.load_session_mapping(session)
-            .and_then(|m| m.active_conversation_id())
-            .filter(|id| self.state.conversations.contains_key(id))
+        let mapping = self.load_session_mapping(session)?;
+        self.resolvable_history(&mapping).first().copied()
```
JeanMertz: blocking: A user upgrading with the leaked history this PR repairs, for example `[ghost, expired --tmp conversation]`, can still lose the real conversation on the next non-activating command such as `jp c ls`. This method resolves the second entry as active, but teardown calls `all_active_conversation_ids()`, which still reads raw `history[0]`. Ephemeral cleanup then protects the ghost and silently deletes the expired conversation even though the session is attached to it. Please make `all_active_conversation_ids()` use the same resolution rule and cover this interaction with a regression test.

crates/jp_storage/src/backend/null.rs
```diff
@@ -74,3 +81,39 @@ impl LockBackend for NullLockBackend {
...
+    }
+}
+
+impl SessionBackend for ReadOnlySessionBackend {
+    fn load_session(&self, session_key: &str) -> Result<Option<Value>> {
+        self.0.load_session(session_key)
+    }
+
+    fn save_session(&self, _session_key: &str, _data: &Value) -> Result<()> {
+        Ok(())
```
JeanMertz: blocking: A user with a surviving pre-source-prefix session file and a live conversation can lose the mapping by running any `jp -! ...` command. Teardown still runs `cleanup_stale_files()`: legacy migration calls this method, treats `Ok(())` as a durable destination write, and then removes the old file. This backend discarded the destination, so the mapping is gone. Please route cleanup's maintenance write through the filesystem backend, or otherwise prevent source removal unless the destination was actually written, and add a ReadOnly-plus-migration regression test.

crates/jp_workspace/src/session_mapping.rs
```diff
@@ -242,22 +285,38 @@ impl Workspace {
...
+        //
+        // Archived conversations count too: archiving is reversible, so
+        // dropping the entry would lose the session's place in a conversation
+        // the user can restore.
+        let mut conversation_ids: HashSet<_> = self
             .loader
             .load_conversation_ids(ConversationFilter::default())
             .into_iter()
             .collect();
+        conversation_ids.extend(self.archived_conversation_ids());
```
JeanMertz: blocking: Two concurrent CLI runs can still make cleanup forget an existing conversation. If one terminal runs `jp c use archived` and moves the directory from archive to live after another run's live scan above but before this archive scan, the ID appears in neither snapshot. `unarchive_conversation()` takes no conversation lock, so the later `lock_info()` check does not preserve the entry; cleanup can prune it or remove an env-backed mapping while the unarchive command reports success. Please make the partition transition visible to cleanup, for example by locking unarchive and revalidating IDs absent from both scans before pruning.

crates/jp_storage/src/backend/null.rs
JeanMertz: Confirmed and fixed in `f6a3c957`, though not the way you suggested.

The mechanism is exactly as you describe: `save_session` at `session_mapping.rs:430` returns `Ok(())` having written nothing, and the migration reads that as durable before `fs::remove_file`. A test that writes a legacy bare-value file against a `ReadOnlySessionBackend` and runs cleanup ends with **zero** session files on the unfixed code.

I didn't route the maintenance write through the filesystem backend. That would put `jp -!` back to writing session files on disk, which is the leak this PR set out to close. The narrower "don't remove the source unless the destination was written" also leaves most of the hole open: the `should_remove` path (`:362`) and the duplicate-legacy path (`:379`) call `fs::remove_file` directly, bypassing the session backend entirely, so an ephemeral run could still delete mappings the read-only wrapper was meant to protect.

Instead `SessionBackend` gained `is_read_only()` (default `false`, `true` on `ReadOnlySessionBackend`), and `cleanup_stale_files` returns before the session pass when it's set. A run that cannot write session state doesn't maintain it either. Orphaned lock-file reaping still runs — that's guarded by the actual flock and unrelated to session storage.

Regression test: `cleanup_skips_session_maintenance_when_session_storage_is_read_only`, observed red first (0 files, expected 1).

crates/jp_workspace/src/session_mapping.rs
JeanMertz: Correct, and taken as proposed — fixed in `f6a3c957`.

`all_active_conversation_ids` (`session_mapping.rs:177`) read `mapping.active_conversation_id()`, raw `history[0]`, while `session_active_conversation` resolved past it. The list feeds `remove_ephemeral_conversations` at `lib.rs:573` and removal goes through `self.persist.remove`, so the disagreement is a real deletion of the conversation the session is attached to. This repo's own `just issue-bug` / `issue-feat` use `--tmp=1h`, so expired conversations aren't hypothetical.

`all_active_conversation_ids` now resolves through `resolvable_history`, same rule as `session_active_conversation`.

Test: `ephemeral_cleanup_protects_the_conversation_the_session_resolves_to` — expired `--tmp` conversation on disk with a ghost in front of it, asserting the conversation survives `remove_ephemeral_conversations`. It runs against a workspace with persistence *enabled*, since a Null persist backend would make the assertion hold regardless of which ids are protected. Observed red first.

One scoping note for anyone reading later: this isn't reachable from a freshly created ghost, since the `ReadOnlySessionBackend` change stops those at the source. It needs a history that was already dirty before the upgrade, or a crash before cleanup — which is the case you named, and it's the state this PR exists to repair, so repairing resolution while leaving protection on the old rule was the wrong half.

crates/jp_workspace/src/session_mapping.rs
JeanMertz: The mechanics check out and it's fixed in `f6a3c957`, but with a much smaller change than you proposed, and I'd push back on the severity framing.

Verified: `unarchive_conversation` takes no lock at any of its three call sites (`unarchive.rs:41`, `use_.rs:163`, `use_.rs:248`), so the `fs.lock_info` fallback at `:398` won't hold the entry.

On severity — this isn't a regression. Before this PR an archived conversation was absent from the scan *unconditionally*, so its history entry was pruned every time, no race needed. The two-scan version narrows "always" to a specific interleaving between two `readdir` calls in two concurrent processes. That's an improvement, and the residual case is closer to a construction than a named input.

I declined "lock unarchive and revalidate IDs absent from both scans": adding a lock to the unarchive path changes the conversation-locking protocol and belongs in its own change, and revalidation is a second TOCTOU wearing a hat. A conversation lives in exactly one partition, so re-reading the live partition *after* the archive scan closes the window for a single move in either direction — a conversation missed at its source is already at its destination by the time the destination is read. Three lines, one extra directory listing per teardown.

It does not survive multiple concurrent transitions, and it does not address the missing unarchive lock. It also has no test: reproducing the interleaving needs two processes racing between directory reads, so this rests on the ordering argument rather than a reproduction. Happy to file a ticket for the unarchive lock if you want that pursued properly.

crates/jp_workspace/src/session_mapping.rs
```diff
@@ -142,7 +182,7 @@ impl Workspace {
             .filter_map(|key| {
                 let value = self.sessions.load_session(&key).ok()??;
                 let mapping = mapping_from_value(value, &key)?;
-                mapping.active_conversation_id()
+                self.resolvable_history(&mapping).first().copied()
```
JeanMertz: blocking: A second terminal can create `jp q --new --tmp ...` after this process loaded its index but before teardown. That conversation is persisted and session-active, but absent from this workspace's stale `state`, so `resolvable_history` skips it here. `remove_ephemeral_conversations()` then rescans disk, sees the bare `--tmp` conversation as expired, and deletes it while the other session is using it. This cross-session scan must also protect conversations created after startup. Please resolve against freshly loaded live and archive IDs rather than `self.state`, and add a regression test for an active ephemeral conversation written after `load_conversation_index()`.

JeanMertz: You're describing a regression I introduced two commits ago, and it's worse than the bug it replaced. Fixed in `9fa64487`.

`git diff f6a3c957` on this method shows the swap: `mapping.active_conversation_id()` (raw `history[0]`, read fresh from the session file) became `resolvable_history(&mapping)`, which gates on `self.state.conversations` — the index `load_conversation_index()` populated at startup (`jp_cli/src/lib.rs:500`). The removal side never changed: `remove_ephemeral_conversations` rescans disk via `load_expired_conversation_ids`. So the fix for the sibling comment left a fresh removal set filtered by a stale protection set, and the old raw read had been protecting the other terminal's conversation correctly all along.

One detail sharper than you framed it. `expires_in_duration` (`query.rs:1023-1027`) maps bare `--tmp` to `Duration::new(0, 0)`, and `create_new_conversation` sets `expires_at = id.timestamp() + 0` (`:794-802`). Such a conversation is expired at birth — the flag's own docs say it is "persisted, but only until the conversation is no longer marked as active". Appearing in this list isn't a safety net for it, it's the only thing keeping it on disk. Two terminals is enough, no long-running command required.

`cleanup_stale_files` carries a comment (`:290-295`) warning about exactly this trap. I reintroduced it one function away.

Resolved as you asked. `resolvable_history` takes the live-conversation test as a parameter: `session_active_conversation` passes the workspace index, since opening a conversation requires it to be there; `all_active_conversation_ids` passes a set scanned from the store. The archive lookup stays lazy, so the common path costs no extra I/O.

Test: `ephemeral_cleanup_protects_a_conversation_created_after_the_index_was_loaded` — `load_conversation_index()` on an empty store, then an expires-immediately conversation plus its session mapping written straight through `fs` to stand in for the other process, asserting it survives. Observed red first.

It also turned up that `all_active_conversation_ids_across_sessions` had been covering this method with conversations that only ever existed in memory, so no store scan could see them. Its premise was wrong rather than the change; it writes them through the backend now, and it discriminates between the two resolution sources where before it was indifferent.

Filed **T0003** for the missing unarchive lock from the other thread.

PR #984, opened 2026-08-15
fix(term, cli): Give details views a declared JSON shape

JeanMertz:
A details view is rendered from a list of rows, each of which may or may not carry a label. `details_json` keyed its object by that label, so a row without one had no key to file its value under and was emitted as `{"<value>": ""}` — the data ended up in the object's key space, where a consumer has to parse it back out. `jp attachment ls` listed URLs that way, and any listing built on the same renderer would have too.

Deriving the shape from the rows instead is worse: it makes the output depend on what an invocation happened to produce. A listing holding one row would be an array and the same listing holding none an object, so every consumer needs a fallback for every field.

The shape is now a property of the view rather than of its data. `Details` is either `Fields`, a record of named rows that renders as a JSON object, or `Items`, a sequence that renders as a JSON array. A record cannot hold an unlabelled row and a listing cannot hold a labelled one, so neither can drift into the other's shape, and an empty view keeps its variant's form.

Conversation details are a record and read as they always did. Attachment listings are a sequence and now emit an array of values.

BREAKING CHANGE: `jp attachment ls --format=json` emits an array

The `details` field of a listing was an object whose keys were the listed values (`{"file:///x": ""}`) and is now an array of those values (`["file:///x"]`). A consumer reading keys should read elements instead. `jp conversation show` is unaffected: its rows are all named, so `details` remains an object keyed by field name.

### DIFF
--- a/crates/jp_cli/src/cmd/attachment/ls.rs
+++ b/crates/jp_cli/src/cmd/attachment/ls.rs
@@ -10,19 +10,20 @@ impl Ls {
     pub(crate) fn run(self, ctx: &mut Ctx) -> Output {
         let uris = &ctx.config().conversation.attachments;
 
-        if uris.is_empty() {
-            ctx.printer.println("No attachments in current context.");
-            return Ok(());
-        }
-
-        let title = Some("Attachments".to_owned());
-
         let mut items = vec![];
         for uri in uris {
             items.push(DetailItem::plain(uri.to_url()?.to_string()));
         }
 
-        print_details(&ctx.printer, title.as_deref(), Details::Items(items));
+        // The text views swap an empty listing for a sentence; JSON keeps the
+        // array, so a script sees one shape whether or not anything was
+        // attached.
+        if items.is_empty() && !ctx.printer.format().is_json() {
+            ctx.printer.println("No attachments in current context.");
+            return Ok(());
+        }
+
+        print_details(&ctx.printer, Some("Attachments"), Details::Items(items));
         Ok(())
     }
 }
--- a/crates/jp_cli/src/cmd/attachment/ls.rs
+++ b/crates/jp_cli/src/cmd/attachment/ls.rs
@@ -1,4 +1,4 @@
-use jp_term::table::DetailRow;
+use jp_term::table::{DetailItem, Details};
 
 use crate::{cmd::Output, ctx::Ctx, output::print_details};
 
@@ -17,12 +17,12 @@ impl Ls {
 
         let title = Some("Attachments".to_owned());
 
-        let mut rows = vec![];
+        let mut items = vec![];
         for uri in uris {
-            rows.push(DetailRow::bare(uri.to_url()?));
+            items.push(DetailItem::plain(uri.to_url()?.to_string()));
         }
 
-        print_details(&ctx.printer, title.as_deref(), rows);
+        print_details(&ctx.printer, title.as_deref(), Details::Items(items));
         Ok(())
     }
 }
--- a/crates/jp_cli/src/cmd/conversation/label.rs
+++ b/crates/jp_cli/src/cmd/conversation/label.rs
@@ -8,7 +8,7 @@
 use crossterm::style::Stylize as _;
 use jp_conversation::{Conversation, ConversationId};
 use jp_inquire::prompt::TerminalPromptBackend;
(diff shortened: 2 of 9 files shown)


crates/jp_cli/src/cmd/attachment/ls.rs
```diff
@@ -17,12 +17,12 @@ impl Ls {
...
 
-        let mut rows = vec![];
+        let mut items = vec![];
         for uri in uris {
-            rows.push(DetailRow::bare(uri.to_url()?));
+            items.push(DetailItem::plain(uri.to_url()?.to_string()));
         }
 
-        print_details(&ctx.printer, title.as_deref(), rows);
+        print_details(&ctx.printer, title.as_deref(), Details::Items(items));
```
JeanMertz: blocking: A user running `jp attachment ls --format=json` with no configured attachments never reaches this call. The early return on lines 13 to 15 uses `Printer::println`, which wraps the prose as `{"message":"No attachments in current context."}`, while a non-empty invocation emits `{"title":"Attachments","details":[...]}`. The machine-output shape therefore still depends on the data, and the command exits successfully with a valid but incompatible object. Keep the prose shortcut for text formats, but send JSON formats through `print_details(..., Details::Items(vec![]))`; a command-level regression test should assert the same `details` array shape for empty and non-empty contexts.

JeanMertz: Confirmed and fixed.

You're right that this lands on the PR's own claim rather than on a tangent. The commit message says the shape is "a property of the view rather than of its data" and that "an empty view keeps its variant's form" — and `jp attachment ls` is the only command whose output this PR changes, so leaving it shape-unstable made that claim false for its single affected case.

Verified the mechanism end to end: `println` routes through `wrap_json` (`printer.rs:267-276`), which emits `{"message": ...}`, so the early return and `print_details` produced two different shapes for one command.

`Ls::run` now builds the items first and lets only the text formats take the prose shortcut:

```rust
if items.is_empty() && !ctx.printer.format().is_json() {
    ctx.printer.println("No attachments in current context.");
    return Ok(());
}

print_details(&ctx.printer, Some("Attachments"), Details::Items(items));
```

Text output is unchanged in both cases; JSON gains `{"title":"Attachments","details":[]}` for the empty one.

I used the format check rather than `print_details_with_json`, since that helper exists for commands whose JSON needs a shape the rows can't express (`jp c label ls` nests under `conversation`), and `Details::Items(vec![])` already produces the right value here.

Test added as `an_empty_listing_is_still_an_array_in_json`, using the `Ctx` construction already in `attachment_tests.rs` — no harness work needed. It asserts `details == []` **and** that no `message` key is present; the second assertion is the one that actually catches a regression to `println`, since `details == []` alone would pass against any future shape carrying an empty array. I reverted the fix and watched it go red (`left: Null, right: Array []`) before restoring, so it isn't vacuous.

The non-empty half of your request is already pinned one layer down by `json_details_shape_follows_the_variant_not_the_data` in `jp_term`, which asserts `Items` renders as an array both empty and populated — so the genuinely new coverage here is that the command reaches `print_details` at all.

The sibling commands with this same pattern (`jp c archive`, `jp c rm`, `jp c unarchive`, and nine others) stay with T0004, which already names the empty-results decision explicitly.

crates/jp_cli/src/cmd/conversation/label.rs
```diff
@@ -310,9 +310,24 @@ fn list(ctx: &Ctx, handles: &[ConversationHandle]) -> Output {
     }
 
     for handle in handles {
+        let id = handle.id();
         let conversation = ctx.workspace.metadata(handle)?;
-        let rows = label_rows(&conversation);
-        print_details(&ctx.printer, Some(&handle.id().to_string()), rows);
+
+        let json = json!({
```
JeanMertz: nit: A user running `jp conversation label ls --format=json` against a labelled conversation gets a second breaking schema change here. The base branch emitted `{"title":"...","details":{"foo":""}}`, while this emits `{"conversation":{...},"labels":[{"key":"foo","value":""}]}`. The new shape is clearer, but the PR's `BREAKING CHANGE` section only names `jp attachment ls`, so label-list consumers get no migration notice. Please document this output change too, and preferably pin the populated and empty forms with command-level tests.

PR #986, opened 2026-08-16
chore(ticket): Replace the id counter with time-ordered random ids

JeanMertz:
`docs/ticket/.counter` is a coordination primitive, and ticket creation has no coordination point. Parallel git worktrees each read the same counter value and hand out the same number; once testers file tickets from their own clones, the same happens across machines.

Renumbering at merge is not a fix. An id that has reached a pushed commit message cannot be rewritten, and the assistant references ticket ids in commits and pull request descriptions routinely. An id must be final the moment it is created, in a checkout that cannot see its siblings, which rules out any local-only allocator.

Ids become `T-02wt0kx`: five base-32 characters holding a five-second bucket since 2026-08-10, then two of randomness. Every position is fixed width and the alphabet's ASCII order matches its digit order, so comparing two ids as bytes orders them by creation time and `ls docs/ticket/` still reads chronologically.

Two things follow from dropping the counter. Ids are collision-resistant rather than never-reused: nothing records an id once its file is gone. And two files can claim one id, which leaves every reference to it ambiguous, so `store::list` and the docs build both refuse rather than pick. The build already aborted on duplicate RFD ids.

`jp ticket refresh <path>` is the repair.

`jp ticket migrate` converts tickets left in the old `NNNN-slug.md` shape on a branch cut before this change. The four existing tickets move to `T-005zd00` through `T-005zd03`.


crates/internal/ticket/src/store.rs
```diff
@@ -290,8 +329,15 @@ pub fn close(dir: &Utf8Path, id: TicketId) -> Result<(Utf8PathBuf, Status)> {
 /// Read every ticket in `dir`, ordered by id.
 ///
 /// A directory that doesn't exist yet holds no tickets.
+/// Two files claiming one id is an error: nothing downstream can tell which of
+/// them a reference means, so the board refuses to render rather than pick.
 pub fn list(dir: &Utf8Path) -> Result<Vec<Entry>> {
-    files(dir)?
+    let files = files(dir)?;
+    if let Some(error) = duplicate(&files) {
```
JeanMertz: blocking: A plausible input is a checkout containing two merged files with the same id. `list` refuses that state here, but id-based writes such as `append_comment`, `close`, `delete`, `edit`, and `promote` go through `locate`, which uses `find` and silently picks one file. A user can therefore mutate or delete the wrong ticket without knowing. Move this ambiguity check into the shared resolver so every operation by id refuses duplicates.

crates/plugins/command/ticket/src/main.rs
```diff
@@ -366,6 +397,188 @@ fn with_email(name: String) -> String {
...
+fn refresh(dir: &Utf8Path, path: &Utf8Path, base: &str) -> Result<Output, String> {
+    let bucket = created_bucket(dir, path)?;
+    let changed = branch_files(dir, base)?;
+
+    let done = store::reassign(dir, path, bucket).map_err(|error| error.to_string())?;
+    let mut output = Output::from(format!("{} -> {} at {}\n", done.old, done.new, done.path));
+
+    let new = done.new.to_string();
+    for file in changed {
+        if file == path {
```
JeanMertz: blocking: A losing ticket that already has a reply contains `- **Re**: T-02wt0kx#1`. `store::reassign` changes only the heading, and this skip means the renamed ticket is never passed through `rewrite`, so that reply keeps pointing at the winning ticket's id. Rewrite the known self-reference fields in the reassigned ticket, and add a refresh test with at least one reply.

crates/plugins/command/ticket/src/main.rs
```diff
@@ -366,6 +397,188 @@ fn with_email(name: String) -> String {
...
+            continue;
+        }
+        if rewrite(&file, &done.old, &new)? {
+            output.text.push_str(&format!("  rewrote {file}\n"));
+        }
+    }
+
+    let board = dir.join(".board.json");
+    if rewrite(&board, &done.old, &new)? {
```
JeanMertz: blocking: Changed-file membership does not tell us which occurrence the branch introduced. For example, if the branch edits an unrelated line in a file that already referenced the winning ticket on `base`, line 419 rewrites that existing reference too. The unconditional board rewrite has the same problem when `.board.json` already ranks the winner. Compare against the base version and rewrite only branch-added occurrences, or skip and report any file whose base content already contains the old id for manual repair.

crates/plugins/command/ticket/src/main.rs
```diff
@@ -366,6 +397,188 @@ fn with_email(name: String) -> String {
...
+        return Err(format!("No file at {path}."));
+    }
+
+    let added = git(dir, &[
+        "log",
+        "--diff-filter=A",
+        "--format=%at",
+        "-1",
+        "--",
+        path.as_str(),
```
JeanMertz: blocking: The documented input `docs/ticket/02wt0kx-some-bug.md` is workspace-relative, but this `git` command runs with `current_dir(dir)`, so Git interprets it relative to `docs/ticket/` and finds no add commit. The silent fallback then assigns the current bucket, making refreshed tickets jump in ordering. Normalize the input once, pass Git a path relative to its chosen working directory (or run Git at the repository root), and test the workspace-relative spelling exposed by the recipe.

crates/internal/ticket/src/store.rs
```diff
@@ -349,24 +395,142 @@ fn files(dir: &Utf8Path) -> Result<Vec<(TicketId, Utf8PathBuf)>> {
...
+
+/// Give the ticket at `path` a new id, renaming its file and rewriting its
+/// heading.
+///
+/// `bucket` places the new id in time; pass the one the ticket was created in
+/// so it keeps its position relative to everything around it.
+/// The slug is kept, and references held by other files are the caller's to fix
+/// — nothing here knows which of them meant this ticket.
+pub fn reassign(dir: &Utf8Path, path: &Utf8Path, bucket: u32) -> Result<Reassigned> {
+    let source = fs::read_to_string(path)?;
```
JeanMertz: blocking: `path` is unvalidated CLI input. A typo such as `jp ticket refresh docs/rfd/102-collision-resistant-ticket-identifiers.md` passes this read and `set_id`, gets copied into `docs/ticket/`, and is then deleted from `docs/rfd/`. Before any write, require the normalized path to be a direct child of `dir` with a current or legacy ticket filename, and verify that its heading id matches the filename.

crates/internal/ticket/src/store.rs
```diff
@@ -349,24 +395,142 @@ fn files(dir: &Utf8Path) -> Result<Vec<(TicketId, Utf8PathBuf)>> {
...
+            Some(id) => match id.tail() + 1 {
+                // The bucket is full, which takes 1,024 tickets in five
+                // seconds. Move to the next one rather than reusing a tail.
+                next if next >= TAIL_SPACE => {
+                    bucket += 1;
+                    continue;
+                }
+                next => next,
+            },
+            None => random_tail(),
```
JeanMertz: blocking: Two parallel creates in the same checkout can both observe an empty bucket and draw the same tail here. If their titles produce the same slug, the later `fs::write` silently overwrites the first ticket, which is hidden data loss on the exact parallel-creation path this change targets. Claim the destination with `create_new(true)` and retry allocation when it already exists; the existence check and write need to be one atomic filesystem operation.

crates/plugins/command/ticket/src/main.rs
```diff
@@ -366,6 +397,188 @@ fn with_email(name: String) -> String {
...
+
+    if legacy.is_empty() {
+        return Ok("No tickets to migrate.\n".to_owned().into());
+    }
+
+    let mut output = Output::default();
+    let mut renamed = vec![];
+    for path in &legacy {
+        let bucket = created_bucket(dir, path)?;
+        let done = store::reassign(dir, path, bucket).map_err(|error| error.to_string())?;
```
JeanMertz: blocking: Migration mutates each file before all legacy files have been validated. With two legacy files where the second has a malformed heading, the first is renamed here, the command exits on the second, and reference rewriting is never reached. A retry no longer includes the first file in `renamed`, so its old references remain dangling permanently. Preflight every source and bucket before mutating, or persist/apply each old-to-new mapping before another fallible conversion.

docs/rfd/100-in-repo-ticket-tracking.md
```diff
@@ -59,21 +60,25 @@ without pushing task work back into RFDs.
...
+> Ids become collision-resistant rather than never-reused.
+
 ## Ticket Format
 
-A ticket is a single markdown file at `docs/ticket/NNNN-slug.md`, where `NNNN`
-is a zero-padded sequential number.
-The canonical reference form is `T0042`; tooling accepts `42`, `042`, and `T42`.
+A ticket is a single markdown file at `docs/ticket/<id>-slug.md`.
+The canonical reference form is `T-02wt0kx`; tooling accepts `T02wt0kx`, the
+bare body, any case, and unique prefixes.
```
JeanMertz: nit: A user entering the documented unique prefix `T-02wt` gets a parse error. `TicketId::normalize` permits a short body, but `FromStr` requires exactly seven characters and no lookup caller resolves prefixes. Either implement ambiguity-checked prefix lookup at the command/tool boundary or remove this accepted-input claim.

justfile
```diff
@@ -1990,7 +1990,7 @@ ticket-add KIND *TITLE: _install-ticket
 
 # Append a comment to a ticket.
 #
-# NNN is 42, 042, or T0042; append `#N` to reply to the Nth comment. The body
+# NNN is T-02wt0kx or 02wt0kx; append `#N` to reply to the Nth comment. The body
```
JeanMertz: nit: The examples immediately below this still use `42` and `42#1`. Copying any of them fails because the new parser rejects numeric ids. Update the example invocations to the new body or canonical form too.

docs/rfd/100-in-repo-ticket-tracking.md
```diff
@@ -59,21 +60,25 @@ without pushing task work back into RFDs.
...
+> checkouts allocate without coordinating.
+> Ids become collision-resistant rather than never-reused.
+
 ## Ticket Format
 
-A ticket is a single markdown file at `docs/ticket/NNNN-slug.md`, where `NNNN`
-is a zero-padded sequential number.
-The canonical reference form is `T0042`; tooling accepts `42`, `042`, and `T42`.
+A ticket is a single markdown file at `docs/ticket/<id>-slug.md`.
+The canonical reference form is `T-02wt0kx`; tooling accepts `T02wt0kx`, the
```
JeanMertz: nit: The generated `jp ticket` help still advertises `42`, `042`, and `T0042` for the comment, close, edit, delete, show, and promote arguments, and the delete help still says the counter retires numbers. A user following that help gets rejected input and the wrong deletion semantics. Update the Clap doc comments in `crates/plugins/command/ticket/src/main.rs` as part of this format migration.

crates/plugins/command/ticket/src/main.rs
```diff
@@ -366,6 +397,188 @@ fn with_email(name: String) -> String {
...
+/// The new id lands in the bucket the ticket was created in, so it keeps its
+/// place in time rather than jumping to now.
+/// References are rewritten only in what the branch introduced; the same token
+/// on `base` belongs to the ticket that won the id, and rewriting it there
+/// would redirect it.
+fn refresh(dir: &Utf8Path, path: &Utf8Path, base: &str) -> Result<Output, String> {
+    let bucket = created_bucket(dir, path)?;
+    let changed = branch_files(dir, base)?;
+
+    let done = store::reassign(dir, path, bucket).map_err(|error| error.to_string())?;
```
JeanMertz: blocking: A plausible failure is a branch-changed file that is readable but not writable. The ticket is renamed here, a later `rewrite` returns an error, and the caller receives no `old -> new` output even though some mutations already landed; retrying with the old path then fails. Make the repair recoverable across partial I/O failure, for example by preflighting targets and returning the reassignment mapping even when a later rewrite fails, or by staging all writes before replacing originals.

crates/internal/ticket/src/id.rs
```diff
@@ -0,0 +1,178 @@
+//! Ticket identifiers.
+//!
+//! An id is seven Crockford base-32 characters: five encoding a five-second
+//! bucket since the epoch, two carrying randomness.
+//! The canonical form prefixes them with `T-`, as in `T-02wt0kx`.
+//!
+//! Every position is fixed width and the alphabet's ASCII order matches its
+//! digit order, so comparing two ids as bytes orders them by creation time.
```
JeanMertz: nit: Byte order is chronological only to the five-second bucket. Two checkouts creating tickets in the same bucket use unrelated random tails, so the later ticket can sort first. Match the RFD wording here (`orders by bucket, then tail`) so this doc comment does not promise strict creation-time ordering to callers of `Ord`.

JeanMertz: I thought we had implemented such that the random tail ALSO guaranteed ordering?

docs/rfd/100-in-repo-ticket-tracking.md
JeanMertz: I would remove this accepted input claim

crates/internal/ticket/src/id.rs
JeanMertz: We did, but the guarantee only reaches as far as the allocator can see.

`allocate_in` finds the highest id **in `dir`** and increments its tail. So sequential creations that share a directory are ordered — that's the `ticket_create`-five-times-in-one-turn case, and it holds. Two checkouts that can't see each other's files each call `random_tail()` independently, so their tails are unrelated and between those two the order is arbitrary.

The Format table in RFD 102 says this correctly ("to the bucket; sequential local creations are always ordered"). The `id.rs` prose didn't — it claimed byte order equals creation time with no qualifier. Fixed to say "by bucket, then tail", with the local-vs-cross-checkout distinction spelled out.

crates/internal/ticket/src/store.rs
JeanMertz: Accepted. `locate` now collects every file claiming the id and returns `Error::Duplicate` when there is more than one, so `close`, `append_comment`, `delete`, `edit`, `set_field`, and `promote` all refuse instead of taking whichever the directory yielded first.

One thing worth recording, because it constrains the fix: the check has to go in `locate`, **not** in `files`. `reassign` reaches `files` through `allocate_in`, so erroring there would make `jp ticket refresh` unable to repair the duplicate it exists for. Id-based lookup refuses; path-based operations stay permissive. That's also the answer to your earlier point about a hard error making lookup unavailable while the duplicate exists — it is, but only for the operations that genuinely can't proceed.

Test: `a_write_by_a_duplicated_id_is_refused` in `store_tests.rs`.

crates/internal/ticket/src/store.rs
JeanMertz: Accepted, with the remedy scoped tighter than proposed.

`create` now allocates, then claims the path with `OpenOptions::create_new(true)`, and on `AlreadyExists` re-runs allocation and retries (bounded at 16, then `Error::Contended`). Each retry sees the file it lost to, so the tail advances rather than redrawing into the same clash. That closes the silent-overwrite half — the one that loses data.

I'd decline the lock-file half of the suggestion. `create_new` *is* the atomic reservation; the filesystem provides it, and a lock adds machinery that still wouldn't help across checkouts, which is where the real concurrency in this project lives.

Being straight about the gap: there's no test. A deterministic reproduction needs two threads racing a directory, and a timing-based one would pass whether or not the mechanism fired — the vacuous-test shape. The change is defensive against a race I can't force, and I'd rather say so than ship a green test that proves nothing.

Also worth noting the blast radius is narrower than the PR's motivating case: this needs two `jp` processes in **one** worktree inside the same five-second bucket. The parallelism that motivated RFD 102 is across worktrees, which are separate directories.

crates/internal/ticket/src/store.rs
JeanMertz: Accepted — and it reproduces exactly as described. RFD 102's first line is `# RFD 102: Collision-Resistant Ticket Identifiers`, which `set_id` splits on `:` happily; the slug falls out of `102-collision-resistant-ticket-identifiers`; the file lands in `docs/ticket/` and the RFD is removed.

`reassign` now requires the path to be a ticket file directly inside `dir` before it reads anything, via `ticket_filename`, returning `Error::NotATicketFile`. Both id shapes pass, since converting a pre-RFD-102 ticket is the whole point of `migrate`.

One thing the test surfaced that's worth flagging: the containment check can't be a string comparison. `git rev-parse --show-toplevel` returns the canonical path, so on macOS a repo under `/var/folders/...` comes back as `/private/var/folders/...` and a legitimate ticket gets rejected. Any repo behind a symlinked path hits it, not just temp dirs. `same_directory` compares strings first and falls back to canonicalising both sides.

Test: `reassign_refuses_a_path_outside_the_ticket_directory`, which asserts the source file is still there afterwards.

crates/plugins/command/ticket/src/main.rs
JeanMertz: Accepted. Confirmed: `changed` holds pre-rename paths, and `done.path` is a filename git has never seen, so the renamed ticket never reached `rewrite` at all — the skip was incidental to that.

`refresh` now runs `rewrite(&done.path, ...)` immediately after `reassign`, before the loop.

Worth noting `migrate` never had this bug: it rewrites across `store::list(dir)` after renaming, which picks up the new files.

Test as requested: `refresh_rewrites_a_reference_inside_the_ticket_it_renames` builds a losing ticket with two comments where the second carries `- **Re**: T-02wt0kx#1`, and asserts the `Re` field names the new id. I checked it fails without the fix rather than assuming — it reports `left: Some("T-02wt0kx#1")` against `right: Some("T-03j98x7#1")`.

crates/plugins/command/ticket/src/main.rs
JeanMertz: Finding accepted, remedy amended — I took the second of your two suggestions rather than the first.

`refresh` now checks each candidate with `git show <base>:<file>`. If the base version already contains the old id, the file is skipped and reported in a warning naming it. Only files where the token is absent on base get rewritten — those are unambiguously the branch's. `.board.json` goes through the same check instead of the unconditional rewrite it had.

I declined the "compare against the base version and rewrite only branch-added occurrences" option deliberately. Doing that correctly means per-hunk attribution, and a partially-correct implementation is worse than none here: it would look authoritative while still redirecting the occasional reference. Skip-and-report converts a hidden wrong-rewrite into a visible "these three files need your eyes", which is the failure mode I'd rather ship.

Two tests, because the behaviour cuts both ways:
- `refresh_reports_a_file_that_named_the_id_on_base` — a file on `main` naming the winner, edited by the branch for an unrelated reason. Asserts the content is byte-identical afterwards and that it appears in exactly one warning.
- `refresh_rewrites_the_branch_and_leaves_the_base_alone` — updated to assert `warnings` is now *empty*, since `NOTES.md` was never in `git diff` and so was never a candidate. That assertion caught the behaviour change when I made it.

crates/plugins/command/ticket/src/main.rs
JeanMertz: Accepted. The `path.exists()` guard was actively hiding this: it resolves against the *process* CWD, so the workspace-relative spelling passed the check, then git — running in `docs/ticket/` — looked for `docs/ticket/docs/ticket/…`, found nothing, and fell through to `current_bucket()`.

`refresh` now resolves the path first via `resolve_ticket_path`: absolute is taken as given; relative is tried against the repository root (the spelling `just ticket-refresh` documents) and then against the ticket directory, so naming a file from inside `docs/ticket/` works too. Everything downstream gets an absolute path. The `exists()` check is gone — resolution proves existence, and failing to resolve is now an error rather than a silent fallback.

Test: `refresh_accepts_a_repository_relative_path` passes `docs/ticket/02wt0kx-filed-earlier.md` against a commit pinned to `2026-08-14T00:00:00Z` and asserts bucket 69,120. A silent fallback puts it at the current bucket, so that assertion can only hold if git actually resolved the path.

My existing tests missed this because both passed `dir.join(...)`, which is always absolute. That's the fixture being too convenient, and it's the same reason the item above went unnoticed.

docs/rfd/100-in-repo-ticket-tracking.md
JeanMertz: Accepted. Six occurrences of `/// Ticket id: \`42\`, \`042\`, or \`T0042\`.` on `Comment`, `Close`, `Edit`, `Delete`, `Show`, and `Promote`, all now `\`T-02wt0kx\`, \`T02wt0kx\`, or \`02wt0kx\`.`

The delete help was worse than stale — "Its number stays retired: the counter never goes backwards" describes a guarantee this PR removes, and RFD 102's change table records the reversal explicitly. It now reads: "Its id is not retired: a later creation in the same time bucket can draw it again."

docs/rfd/100-in-repo-ticket-tracking.md
JeanMertz: Done — RFD 100 now reads "tooling accepts `T02wt0kx`, the bare body, and any case."

For the record on why removal is the right call rather than building the lookup: prefix resolution was proposed back when the alphabet was base-36 and `0`/`o` were confusable, as the answer to hand-typing friction. Switching to Crockford base-32 solved that problem directly, so the feature lost its justification. Seven characters isn't long enough to need abbreviating, and every tool emits the canonical form, so ids get pasted rather than typed.

This was the last place the claim survived — RFD 102 and the three tool TOMLs were already corrected.

justfile
JeanMertz: Accepted. All four updated to `T-02wt0kx` / `T-02wt0kx#1`. I'd fixed the `NNN is …` line directly above and left the examples underneath it, which is the more copy-pasted half.

crates/plugins/command/ticket/src/main.rs
```diff
@@ -366,6 +398,243 @@ fn with_email(name: String) -> String {
...
+        .map(|line| root.join(line))
+        .collect())
+}
+
+/// Replace every `old` with `new` in `file`, reporting whether anything moved.
+///
+/// A file that isn't there, or isn't text, is skipped rather than failing the
+/// run: the caller is working through a list it didn't curate.
+fn rewrite(file: &Utf8Path, old: &str, new: &str) -> Result<bool, String> {
+    let Ok(source) = std::fs::read_to_string(file) else {
```
JeanMertz: blocking: A plausible input is a branch-changed text file that contains the old id but is unreadable because of its permissions. This turns every read error into `Ok(false)`, so `refresh` exits successfully while leaving that reference pointed at the winning ticket. Skip only expected cases such as `NotFound` for a deleted path and invalid UTF-8 for a binary file; propagate or at least warn on other I/O errors so an incomplete repair cannot look successful.

crates/plugins/command/ticket/src/main.rs
```diff
@@ -366,6 +398,243 @@ fn with_email(name: String) -> String {
...
+
+    let mut legacy: Vec<Utf8PathBuf> = vec![];
+    for entry in entries {
+        let path = entry.map_err(|error| error.to_string())?.path();
+        let Ok(path) = Utf8PathBuf::try_from(path) else {
+            continue;
+        };
+        if path
+            .file_name()
+            .is_some_and(|name| name.len() > 5 && name[..4].bytes().all(|b| b.is_ascii_digit()))
```
JeanMertz: nit: A non-ticket Markdown file whose fourth byte is not a UTF-8 boundary, for example `aéé.md`, panics this command at `name[..4]` before it can report that there is nothing to migrate. Check the first four bytes through `name.as_bytes().get(..4)` (and the following `-`) rather than slicing the string by a byte offset.

docs/rfd/100-in-repo-ticket-tracking.md
```diff
@@ -59,21 +60,25 @@ without pushing task work back into RFDs.
...
 Unlike RFDs, tickets can be deleted — a ticket carrying false claims or
 imported spam is removed outright, so that nothing reads it as true.
-Ticket numbers are never reused: the next id comes from a monotonic counter, not
-from the highest file on disk.
 References to a deleted ticket render as dangling.
 
+> [!TIP]
+> [RFD 102] replaces the counter with a time-ordered random id, so that two
+> checkouts allocate without coordinating.
+> Ids become collision-resistant rather than never-reused.
```
JeanMertz: nit: The deletion UI still promises the opposite contract: `docs/.vitepress/theme/TicketIndex.vue` says “The number stays retired,” and `store::delete` still says the counter never goes backwards. A user deleting from the site is told the id cannot return when this change explicitly allows reuse. Update those stale descriptions with the format migration.

crates/internal/ticket/src/store.rs
JeanMertz: blocking: The directory and filename validation closes the destructive wrong-path case, but the heading still is not checked against the filename id. A conflict-resolved file such as `02wt0kx-loser.md` headed `# T-03abcde: Loser` passes `ticket_filename`; `reassign` then reports and rewrites `T-03abcde` while references to the collided `T-02wt0kx` remain untouched. Parse the heading before allocation and require it to match the filename for current-format tickets (legacy filenames should similarly match `TNNNN`).

crates/plugins/command/ticket/src/main.rs
JeanMertz: Correcting my earlier reply on this thread: the fix I described is gone, and so is the bug.

Tickets no longer carry their own id anywhere inside the file. The heading is `# Title`, and a reply is `- **Re**: #1` rather than `- **Re**: T-02wt0kx#1` — a reply always targets a comment on the same ticket, so the position alone says everything. With no structural self-reference left, `store::reassign` is a plain `fs::rename`.

So `rewrite(&done.path, ...)` and `refresh_rewrites_a_reference_inside_the_ticket_it_renames` have both been removed. What replaces them:

- `refresh_leaves_a_reply_alone` — asserts the renamed file is byte-identical afterwards and its `Re` field still reads `#1`.
- `refresh_reports_prose_naming_the_old_id_in_the_renamed_ticket` — the renamed file now goes through the same ambiguity treatment as every other candidate.

That second one is more correct than what I originally shipped. Prose inside the loser saying `duplicate of T-02wt0kx` most likely means the ticket that *kept* the id, and the unconditional rewrite I added would have silently redirected it — the exact failure mode you flagged one thread over. It's now reported instead.

Your finding stands; it just turned out to be a symptom of storing the id twice.

crates/internal/ticket/src/store.rs
JeanMertz: One correction to my earlier reply on this thread: the guard stands, but the mechanism I described has changed.

`reassign` no longer reads the file, rewrites the heading, writes to the new path and deletes the old one. It is now `ticket_filename` validation plus `fs::rename`, because the document no longer carries the id and there is nothing inside to update.

The guard is still needed and unchanged in substance: a rename would move an RFD into `docs/ticket/` just as the write-then-delete would have. What changes is the shape of the damage, not whether it can happen — so `Error::NotATicketFile`, the `ticket_filename` check, `same_directory`'s canonicalisation fallback, and `reassign_refuses_a_path_outside_the_ticket_directory` all stay as described.

JeanMertz: Correct against the code as it stood, and now moot rather than fixed.

Rather than cross-checking two copies of the id, the second copy is gone: the heading is `# Title` and the filename is the only place a ticket names itself. `Reassigned.old` now comes from `ticket_filename`, which is the same id `files()`, `locate()`, and the duplicate check all key on — so a refresh can only ever report and rewrite the id the collision is actually on. There is no `Error::IdMismatch` and no cross-check, because the mismatch is no longer representable.

Two details worth recording:

Pre-RFD-102 tickets genuinely did carry the id in the heading (`# T0005: Old ticket`). Since `reassign` only renames, `jp ticket migrate` strips it with `render::strip_heading_id`, which matches the id exactly — so a title that merely contains a colon, `# Fix: the wrap calculation`, is left alone. There's a test for that.

The cost, for the record: `git grep T-02wt0kx` now finds references to a ticket rather than the ticket itself (use `docs/ticket/02wt0kx-*`), and a ticket body pasted elsewhere carries no id. Both are written into RFD 102 alongside why this diverges from RFDs, which do keep the number in both places.

docs/rfd/100-in-repo-ticket-tracking.md
JeanMertz: Accepted, both confirmed stale and both fixed.

`docs/.vitepress/theme/TicketIndex.vue:105` now reads "This removes the file. A later ticket in the same time bucket can draw the id again." That was the worse of the two: it's a promise made to the user in a confirmation dialog at the moment they decide to delete, and this PR makes it false.

`store::delete`'s doc comment dropped "the counter never goes backwards" for the same wording as the plugin's `Delete` help.

Same class as the clap-doc thread above — I fixed the plugin's copy there and missed these two.

crates/plugins/command/ticket/src/main.rs
JeanMertz: Accepted. Extracted as `legacy_filename`, which walks the bytes instead of slicing the string:

```rust
fn legacy_filename(name: &str) -> bool {
    let bytes = name.as_bytes();

    bytes.get(..4).is_some_and(|id| id.iter().all(u8::is_ascii_digit))
        && bytes.get(4) == Some(&b'-')
}
```

I reproduced it before fixing rather than taking it on faith — `aéé.md` panics with `end byte index 4 is not a char boundary; it is inside 'é' (bytes 3..5 of string)`, exactly as you described. Test: `migrate_skips_a_non_ascii_filename_without_panicking`.

I also took the `-` check you mentioned. Without it `12345-foo.md` passed the scan and then failed hard in `ticket_filename`, which errors rather than corrupting anything, but the scan should be exact. Second test: `migrate_ignores_a_filename_that_only_starts_with_digits`.

On severity, for the record: I'd have called this contained and visible rather than blocking — the panic happens while collecting candidates, before any `reassign` runs, so nothing is written and a backtrace is impossible to miss. It's still worth fixing; a directory scan shouldn't ever be able to crash.

crates/plugins/command/ticket/src/main.rs
JeanMertz: Accepted, though not quite for the reason given — worth saying so plainly.

The permissions scenario is thin: a file in your own repo that git lists as branch-changed but you can't read is close to hypothetical. What actually justifies the change is that the doc comment two lines above already claimed the narrow behaviour — "A file that isn't there, or isn't text, is skipped" — while `let Ok(source) = … else` caught everything. The code didn't do what it said, and that gap is the kind that stays invisible until it bites.

Now:

```rust
let source = match std::fs::read_to_string(file) {
    Ok(source) => source,
    Err(error) if matches!(
        error.kind(),
        io::ErrorKind::NotFound | io::ErrorKind::InvalidData
    ) => return Ok(false),
    Err(error) => return Err(format!("{file}: {error}")),
};
```

`NotFound` and `InvalidData` are the two the catch-all was written for and both are expected: `git diff --name-only` lists files the branch *deleted*, and binary files aren't text. Everything else now fails the run rather than letting `refresh` report success on an incomplete repair.

Two tests, and I checked the first fails against the old code (it returned `Ok(false)` where it now errors):
- `rewriting_an_unreadable_file_is_an_error` — reads a directory, which fails with neither of the two expected kinds.
- `rewriting_a_missing_or_binary_file_is_skipped` — pins both expected cases so the narrowing can't quietly become a widening again.

crates/plugins/command/ticket/src/main.rs
```diff
@@ -658,7 +960,9 @@ fn show(dir: &Utf8Path, id: TicketId, json: bool) -> Result<Output, String> {
         return Ok(format!("{json}\n").into());
     }
 
-    let mut out = format!("# {}: {}\n\n", ticket.id, ticket.title);
+    // The rendered view names the ticket even though the file doesn't: the id
+    // is what a reply or a `Blocked by` has to quote.
+    let mut out = format!("# {}: {}\n\n", entry.id, ticket.title);
```
JeanMertz: blocking: The normal input `just ticket-promote T-02wt0kx` depends on `jp ticket show --json` returning `.id`, but removing `Ticket::id` also removes that field from `Detail` because it only flattens `ticket` and adds `path`. The recipe reads `null`, creates the RFD draft with “Promoted from ticket null,” then `jp ticket promote null` fails, leaving a partial promotion. Add `id: TicketId` to `Detail`, populate it from `entry.id`, and pin the JSON contract with a show test.

crates/plugins/command/ticket/src/main.rs
```diff
@@ -366,6 +398,272 @@ fn with_email(name: String) -> String {
...
+    let mut output = Output::default();
+    let mut renamed = vec![];
+    for path in &legacy {
+        let bucket = created_bucket(dir, path)?;
+        let done = store::reassign(dir, path, bucket).map_err(|error| error.to_string())?;
+
+        // A pre-RFD-102 ticket carried its id in the heading as well as the
+        // filename. `reassign` only renames, so the heading is stripped here.
+        let source = std::fs::read_to_string(&done.path).map_err(|error| error.to_string())?;
+        if let Some(stripped) = render::strip_heading_id(&source, &done.old) {
```
JeanMertz: blocking: A legacy ticket with a reply stores `- **Re**: T0005#1`. This strips the heading id, but the later generic rewrite changes that field to `T-<new>#1` rather than the new `#1` form, so the migrated document still names itself and violates the invariant this refactor relies on. A later collision refresh will treat that structural self-reference as ambiguous and leave it pointing at the winning ticket. Normalize same-ticket `Re` fields while migrating, and add a legacy-ticket test with a reply.

crates/plugins/command/ticket/src/main.rs
```diff
@@ -366,6 +398,272 @@ fn with_email(name: String) -> String {
...
+    let done = store::reassign(dir, &path, bucket).map_err(|error| error.to_string())?;
+    let mut output = Output::from(format!("{} -> {} at {}\n", done.old, done.new, done.path));
+
+    let new = done.new.to_string();
+    let mut ambiguous = vec![];
+
+    // A ticket names itself nowhere structurally, so anything left inside the
+    // renamed file is prose — and prose naming the old id is ambiguous the same
+    // way any other file's is: it may mean the ticket that kept the id.
+    if std::fs::read_to_string(&done.path).is_ok_and(|source| source.contains(&done.old)) {
```
JeanMertz: blocking: A losing ticket can be renamed even when its file is unreadable, because rename only needs directory permissions. `is_ok_and` then turns that read failure into “contains no old id,” and `refresh` reports success while any old-id prose remains unreported. Propagate this read error, as `rewrite` does for unexpected I/O failures, so the repair cannot silently skip the file it just moved.

crates/internal/ticket/src/store.rs
```diff
@@ -349,24 +449,181 @@ fn files(dir: &Utf8Path) -> Result<Vec<(TicketId, Utf8PathBuf)>> {
...
+        .map_err(|_| Error::ClockBeforeEpoch)?
+        .as_secs();
+
+    bucket_at(now)
+}
+
+/// A random tail.
+///
+/// `RandomState` is seeded by the OS once per thread and advances per call, so
+/// two processes starting in the same second still draw different values.
```
JeanMertz: nit: `RandomState` makes the draws independent; it does not guarantee different values. Two processes can still choose the same ten-bit tail, which is the collision this PR detects and repairs. Say “draw independently” here so the comment matches the actual guarantee.

JeanMertz: Accepted. The comment claimed a guarantee the mechanism doesn't provide, and it contradicted the design it sits in — RFD 102's whole argument is that collisions are possible but rare, so the allocator shouldn't read as though they can't happen.

Now:

> `RandomState` is seeded by the OS once per thread and advances per call, so two processes starting in the same second draw independently. Independent is not distinct: ten bits leave a 1-in-1,024 chance they land on the same tail, which is the collision `store::list` and the docs build exist to catch.

crates/plugins/command/ticket/src/main.rs
JeanMertz: Accepted, though the reason I'd give differs from yours.

The unreadable-file scenario is thin — same class as the permissions case in the `rewrite` thread, and you're right that `fs::rename` only needs directory permissions so the two can come apart. What actually decides it is that this contradicts the fix that landed two commits ago in the same function: `rewrite` now propagates unexpected read errors precisely so an incomplete repair can't report success, and four lines earlier `is_ok_and` was doing the opposite.

```rust
match std::fs::read_to_string(&done.path) {
    Ok(source) if source.contains(&done.old) => ambiguous.push(done.path.clone()),
    Ok(_) => {}
    Err(error) => return Err(format!("{}: {error}", done.path)),
}
```

No `NotFound`/`InvalidData` tolerance here, unlike `rewrite`: the rename put this file in place one line earlier, so any read failure is unexpected rather than expected.

Worth being straight about severity and coverage. The damage is narrower than blocking implies — the file is renamed correctly and left unmodified, so this was a missing *report*, not corruption. And there's no test: forcing a file that renames but won't read needs permission games that don't port across the platforms CI runs. I'd rather say that than add something that passes either way.

crates/plugins/command/ticket/src/main.rs
JeanMertz: Accepted — this is the best find of the four, and I traced it exactly as you describe.

`targets` comes from `store::list(dir)`, which includes the ticket just renamed, and `renamed` holds `("T0005", "T-<new>")`. So the cross-file rewrite reached `- **Re**: T0005#1` and produced `- **Re**: T-<new>#1`: a migrated document naming itself, which is the invariant this PR just spent a refactor establishing. A later `refresh` would then treat it as ambiguous prose and leave it pointing at a dead id.

`render::strip_heading_id` is now `render::strip_ids`, doing both legacy conversions — heading and same-ticket reply targets — and `migrate` calls it before the cross-file rewrite, so there's nothing left for that pass to mangle. The `Re` match goes through `parse::meta_line` and requires the value to start with `{old}#`, so a reply naming a *different* ticket survives and is handled by the cross-file pass instead, which is the correct split.

Three tests:
- `stripping_ids_converts_the_heading_and_its_own_replies` — the unit case, exact expected document.
- `stripping_ids_leaves_unrelated_colons_and_replies_alone` — `# Fix: the wrap calculation` and `- **Re**: T0009#2` both survive.
- `migrate_converts_a_legacy_reply_to_the_positional_form` — end to end, asserting the `Re` field reads `#1` and that the file contains neither the old id nor the new one.

I checked the last one fails without the fix rather than assuming: `left: Some("T-04d3xma#1")` against `right: Some("#1")`.

crates/plugins/command/ticket/src/main.rs
JeanMertz: Accepted, and confirmed end to end. `Detail` flattens `Ticket` and adds `path`, so dropping `Ticket::id` silently removed the key; `justfile:2102` does `jq -r '.id'`, which yields the string `null` for a missing key. Under `set -eu` the run then creates and seeds the draft, writes "Promoted from ticket null.", and dies at `jp ticket promote null` — draft on disk, ticket neither closed nor linked.

`Detail` now carries `id: TicketId` from `entry.id`, ordered first so the JSON reads naturally. `.metadata.promoted_to` was unaffected, since `metadata` is a nested field rather than a flattened one.

Test: `show_json_carries_the_id` runs `show --json` and asserts `id`, `title`, and `description`. I confirmed it fails without the field — `left: Null` against `right: "T-04d3xfp"`, which is exactly the `null` the recipe was reading.

The more useful half of this finding is why it slipped. The only `Show` tests were argument-parsing ones; nothing executed `show --json`, so a serialization contract that a `just` recipe depends on had no coverage at all. That gap is what let removing a field break a command two files away.

PR #987, opened 2026-08-17
build(tools): Add Markdown formatting tool

JeanMertz:
Authors can format Markdown files through `markdown_format`, which uses the same `comfort` options enforced by CI. The tool accepts Markdown files or directories, rejects unsupported paths, and reports the formatted files.

Writing and RFD workflows enable the tool so Markdown is formatted before RFD hashes are generated.

### DIFF
--- a/.config/jp/tools/src/markdown.rs
+++ b/.config/jp/tools/src/markdown.rs
@@ -1,7 +1,7 @@
 //! The `markdown_format` tool: reflow standalone Markdown files with `comfort`.
 //!
-//! Covers the `.md` files that live outside Rust sources — RFDs, tickets,
-//! READMEs, the docs site.
+//! Covers the standalone Markdown files (`.md` and `.markdown`) that live
+//! outside Rust sources — RFDs, tickets, READMEs, the docs site.
 //! Doc comments inside `.rs` files are `cargo_format`'s territory.
 //!
 //! The `comfort` flags match the `fmt-markdown-ci` recipe in the justfile, so a
@@ -21,7 +21,7 @@ use crate::{
     },
 };
 
-/// Cap for the formatted-file listing embedded in a tool result.
+/// Cap for each chunk of `comfort` output embedded in a tool result.
 ///
 /// A workspace-wide run after a broad edit can touch hundreds of files, and the
 /// head of the list is what the reader needs.
@@ -47,9 +47,21 @@ fn markdown_format_impl<R: ProcessRunner>(
     paths: Option<OneOrMany<String>>,
     runner: &R,
 ) -> ToolResult {
-    let selected = match select_paths(ctx, paths.unwrap_or_default().as_slice()) {
-        Ok(selected) => selected,
-        Err(message) => return error(message),
+    // An omitted `paths` means the whole workspace; an empty list means the
+    // caller selected nothing, and silently widening that to every Markdown
+    // file in the workspace rewrites files nobody asked about.
+    let selected = match paths {
+        None => vec![],
+        Some(paths) if paths.as_slice().is_empty() => {
+            return error(
+                "`paths` is empty. Name the files or directories to format, or omit `paths` \
+                 entirely to format every Markdown file in the workspace.",
+            );
+        }
+        Some(paths) => match select_paths(ctx, paths.as_slice()) {
+            Ok(selected) => selected,
+            Err(message) => return error(message),
+        },
     };
 
     let mut args = vec![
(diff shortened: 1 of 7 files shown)


.config/jp/tools/src/markdown.rs
```diff
@@ -0,0 +1,136 @@
...
+fn markdown_format(ctx: &Context, paths: Option<OneOrMany<String>>) -> ToolResult {
+    markdown_format_impl(ctx, paths, &DuctProcessRunner)
+}
+
+fn markdown_format_impl<R: ProcessRunner>(
+    ctx: &Context,
+    paths: Option<OneOrMany<String>>,
+    runner: &R,
+) -> ToolResult {
+    let selected = match select_paths(ctx, paths.unwrap_or_default().as_slice()) {
```
JeanMertz: blocking: A call with `{"paths":[]}` is valid under this schema and deserializes to `Some(OneOrMany::Many(vec![]))`, but this collapses it with an omitted argument; line 63 then runs `comfort --workspace`. That input can unexpectedly rewrite every Markdown file when the caller selected no files. The tool description promises workspace scope only when `paths` is omitted, and the other path-scoped tools reject empty arrays for the same reason. Preserve `None` versus `Some(empty)` and return an error or no-op for the latter, with a regression test.

.config/jp/tools/src/markdown.rs
```diff
@@ -0,0 +1,136 @@
...
+        args.extend(selected.iter().map(String::as_str));
+    }
+
+    let ProcessOutput {
+        stdout,
+        stderr,
+        status,
+    } = runner.run("comfort", &args, &ctx.root)?;
+
+    if !status.is_success() {
```
JeanMertz: blocking: A call such as `{"paths":["docs/a.md","docs/read-only.md"]}` can rewrite `a.md` and then fail on the second file. `comfort` processes files sequentially and emits its changed-path list to stdout, but this branch discards that list and reports only stderr. The tool therefore leaves partial edits behind without telling the user which earlier files changed. Include bounded stdout in the failure result with a warning that files may already have been modified, or make the operation transactional; add a failure test with non-empty stdout.

.config/jp/tools/src/markdown.rs
```diff
@@ -0,0 +1,136 @@
...
+/// whole-workspace run.
+///
+/// `comfort` walks a directory and filters it down to Markdown extensions, but
+/// takes a named file as-is whatever its extension — so a file that isn't
+/// Markdown is refused here rather than rewritten as Markdown.
+fn select_paths(ctx: &Context, paths: &[String]) -> Result<Vec<String>, String> {
+    paths
+        .iter()
+        .map(|path| {
+            let resolved = resolve_workspace_path(&ctx.root, path, ctx.access.as_ref())?;
```
JeanMertz: blocking: A user can give `markdown_format` a root write grant and a more-specific read-only rule for `docs/generated`, then call `{"paths":["docs"]}`. `resolve_workspace_path` only validates confinement and approved external mounts; it does not check `Capability::Update`. `comfort` can therefore rewrite Markdown under the read-only subtree. Omitted paths bypass even this resolver. Enforce update access for every file `comfort` may modify, including files discovered by directory and workspace scopes, and cover the carve-out case in a regression test.

.config/jp/tools/src/markdown.rs
```diff
@@ -0,0 +1,136 @@
...
+    // Workspace runs report absolute paths; strip the root so the listing reads
+    // the way the rest of the tools spell a path.
+    let mut files: BTreeSet<String> = BTreeSet::new();
+    for line in stdout.lines() {
+        let trimmed = line.trim();
+        if !trimmed.is_empty() {
+            files.insert(
+                trimmed
+                    .trim_start_matches(ctx.root.as_str())
+                    .trim_start_matches('/')
```
JeanMertz: nit: On Windows, a workspace run can print `C:\repo\docs\usage.md`. Stripping `ctx.root` leaves `\docs\usage.md`, and trimming only `/` keeps that leading separator, so the reported path is not workspace-relative. Use path-aware `strip_prefix` or handle both separators, and add a Windows-shaped output test.

.jp/mcp/tools/markdown/format.toml
```diff
@@ -0,0 +1,41 @@
...
+enable = false
+run = "unattended"
+
+source = "local"
+command = "just serve-tools {{context}} {{tool}}"
+summary = "Format standalone Markdown files: reflow paragraphs, canonicalize structure, and normalize reference links."
+description = """
+Runs `comfort` over Markdown files with the same flags CI enforces, so a \
+formatted file is one CI accepts.
+
```
JeanMertz: nit: A user editing `notes.markdown` will read this as unsupported, but `select_paths` and `comfort` both accept `.markdown`, and CI includes that extension. Describe the scope as `.md` and `.markdown` files, or as standalone Markdown files. The writing-skill instruction has the same `.md`-only wording.

.config/jp/tools/src/markdown.rs
JeanMertz: Fixed. `Tool::opt` only maps `None`/`null` to `None`, so `{"paths":[]}` did arrive as `Some(Many([]))` and `unwrap_or_default()` flattened it into the same empty slice as an omitted argument. That contradicts the parameter doc ("If omitted, every Markdown file in the workspace is formatted") and it spreads past the request: a workspace run reflows every RFD, staling the stored hashes in `rfd-summaries.json` for RFDs the caller never touched.

`markdown_format_impl` now matches on `paths` before flattening: `None` is the workspace run, `Some([])` returns an error naming the fix ("omit `paths` entirely to format every Markdown file in the workspace"). Regression test `an_empty_path_list_is_refused_without_running_comfort` uses `MockProcessRunner::never_called()`, so it fails if the empty list reaches the subprocess again.

One correction to the supporting argument: the other path-scoped tools don't reject empty arrays, they no-op. `fs_grep_files` documents it explicitly (`grep_files.rs:27-29`: "`None` searches the whole workspace, `Some([])` searches nothing") and `fs_list_files` walks zero prefixes. `cargo_update` does error, but there `packages` is required rather than optional. The precedent is *distinguish* the two cases, not reject the empty one. I went with an error rather than a silent no-op because "No files to format." is indistinguishable from "everything was already formatted", which hides the caller's mistake on a tool that writes.

.config/jp/tools/src/markdown.rs
JeanMertz: Took the first half, skipped the second.

The mechanism is as described: `comfort::run::run_files` (`crates/contrib/comfort/src/run.rs:125-148`) prints each changed path to stdout, writes the file, and returns `Err` on the first read/write failure. So a mid-run failure leaves the earlier files reformatted on disk with their paths sitting in the stdout this branch was throwing away.

The failure branch now appends that list: `comfort failed: <stderr>` followed by "Already reformatted before the failure:" and the workspace-relative paths, both chunks bounded by `truncate`. The root-stripping and dedup moved into a `changed_files` helper so the success and failure paths report identically. Test `a_failure_names_the_files_already_reformatted` pins the exact message against a non-zero exit with non-empty stdout.

Not doing the transactional alternative. Staging every file and rolling back on failure is a change to `comfort` itself, and it's out of proportion to the actual damage here: the already-written changes are the same reflow `fmt-markdown-ci` enforces, and they show up in `git status`. What was genuinely lost was the caller's ability to know which files moved, and that's what the listing restores.

.config/jp/tools/src/markdown.rs
JeanMertz: Not taking this one, though the observation is correct.

Scope is narrower than it looks: explicit paths are passed to `comfort` as workspace-relative strings and come back the same way, so only the `--workspace` branch prints absolute paths and only that branch is affected. The consequence is a listing that reads `\docs\usage.md` instead of `docs/usage.md`. Nothing downstream consumes the listing, so the damage is contained to the rendered result and visible to whoever reads it.

It's also a copy of the existing `cargo_format` logic (`cargo/format.rs:87-91`), which does the same `trim_start_matches(root).trim_start_matches('/')`. Fixing it here alone would leave the two formatter tools reporting paths differently on the same platform. If it's worth doing, it's a shared `strip_root` helper applied to both call sites, which is a separate change from this PR.

.config/jp/tools/src/markdown.rs
JeanMertz: Declining, on the strength of a decision the project has already made.

The factual half is right: there's no `authorize(..., Capability::Update, ...)` call here, unlike `fs_create_file` / `fs_modify_file` / `fs_delete_file` / `fs_move_file`. But RFD 076 puts subprocess-spawned file access explicitly out of scope for cooperative `access.fs` enforcement (`docs/rfd/076-tool-access-grants.md:862-870`): "per-argument path analysis fights the subprocess model", with full enforcement handed to RFD 075's OS-level sandbox. `markdown_format` is exactly that shape, and its nearest sibling `cargo_format` rewrites every `.rs` file in the workspace with no policy check at all.

The proposed fix also wouldn't cover the scenario it names. Authorizing the requested path `docs` matches the root grant and passes; catching the `docs/generated` carve-out would mean the tool enumerating the walk itself and duplicating `comfort`'s discovery. And the partial version (named paths only) is worse than none, because it reads as a guarantee that the directory and workspace scopes don't honour.

Worth noting for anyone reading this later: nothing under `.jp/mcp/tools` declares an `access` key, so `ctx.access` is `None` and the policy is unrestricted for every one of these tools today. If we want formatter tools inside the policy, that's a cross-cutting call covering `cargo_format` too, and it belongs on RFD 075 rather than in this PR.

.jp/mcp/tools/markdown/format.toml
JeanMertz: Correct on all three sides, and fixed. `select_paths` matches `Some("md" | "markdown")`, comfort's walk filter matches `(Language::Markdown, Some("md" | "markdown"))` (`crates/contrib/comfort/src/walk.rs:88`), and `fmt-markdown-ci` runs that same walk, so `.markdown` is both accepted and CI-checked. The refusal message already spelled both extensions, which made the description the only place that was wrong.

Updated the tool description to "standalone Markdown files (`.md` and `.markdown`)", the writing-skill line to "after editing a Markdown file", and the module doc in `markdown.rs`, which carried the same `.md`-only phrasing.

PR #988, opened 2026-08-17
fix(cli, rfd): Escalate Ctrl-C only on unanswered presses

JeanMertz:
`EscalationState` reset the Ctrl-C press count on elapsed time alone, so a press an interrupt menu had already answered stayed on the escalation ladder. Answering the streaming menu and interrupting again inside the two-second cooldown counted as a second press, which bypasses the handler stack and requests a graceful shutdown, so the run ended with no menu and no message. Answering the menu quickly made the next Ctrl-C fatal; taking longer than the cooldown left it working.

A delivered press is now an `InterruptNotice` that the notified event loop resolves. `handled()` clears the count, `decline()` hands the press to the next handler down with the count intact, and dropping it leaves the press on the ladder, which is the safe default for a handler that was notified but produced nothing visible. Only presses that produced nothing escalate.

Carrying the reset on the delivered value rather than on the router means an event loop cannot decide what a press did without holding the value that records the decision, so the ladder behaves the same at every handler site. `lock.rs` and `plugin/dispatch.rs` had the same defect and now resolve their presses too.

`inline_select` returning `Err` mapped every error to an escalation, conflating the user cancelling the menu with the menu being unable to run at all. The latter now yields `PromptFailed`, which leaves the turn untouched and the press on the ladder, so a second Ctrl-C gets the user out through the router instead of through a prompt that is not working.

The interrupt path also logs the resolved action, the decision the handler reached, and the `InquireError` behind a prompt that returned without a selection. That error was discarded as `Err(_)`, and its absence is why diagnosing this took several rounds of guessing.

RFD 045 is updated to match: the ladder counts unanswered presses, and `SignalRouter::decline` is replaced by resolving the notice. T0014 stays open, since the traces that opened it showed a different signature and are no longer available.

### DIFF
--- a/crates/jp_cli/src/cmd/query/turn_loop.rs
+++ b/crates/jp_cli/src/cmd/query/turn_loop.rs
@@ -69,14 +69,15 @@ use crate::{
     editor::build_editor_backend,
     error::Error,
     render::metadata::set_rendered_arguments,
-    signals::SignalRouter,
+    signals::{InterruptNotice, SignalRouter},
     timer::LineTimer,
 };
 
 /// Events produced by the merged streaming loop sources.
 enum StreamingLoopEvent {
-    /// A Ctrl-C interrupt notification from the signal router.
-    Interrupt,
+    /// A Ctrl-C press delivered by the signal router, carried as the notice the
+    /// loop resolves once it has decided what the press did.
+    Interrupt(InterruptNotice),
     /// An event from the LLM provider stream.
     Llm(Box<Result<Event, StreamError>>),
     /// A tick from the preparing indicator timer, carrying the elapsed time
@@ -155,7 +156,7 @@ fn event_keeps_waiting_indicator(event: &StreamingLoopEvent) -> bool {
             result.as_ref(),
             Ok(Event::KeepAlive | Event::Patch(_) | Event::Flush { .. })
         ),
-        StreamingLoopEvent::Interrupt | StreamingLoopEvent::PreparingTick(_) => false,
+        StreamingLoopEvent::Interrupt(_) | StreamingLoopEvent::PreparingTick(_) => false,
     }
 }
 
@@ -268,10 +269,11 @@ pub(super) async fn run_turn_loop(
     loop {
         // A Ctrl-C that landed between phases ends the turn gracefully:
         // commit any partial assistant content and complete.
-        if turn_interrupt_rx.try_recv().is_ok() {
+        if let Ok(notice) = turn_interrupt_rx.try_recv() {
             info!("Interrupt received between turn phases; completing the turn.");
             lock.as_mut()
                 .update_events(|stream| turn_coordinator.complete_early(stream));
+            notice.handled();
         }
 
         match turn_coordinator.current_phase() {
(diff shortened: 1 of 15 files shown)


crates/jp_cli/src/signals.rs
```diff
@@ -161,22 +175,53 @@ impl SignalRouter {
...
+    }
+}
+
+impl InterruptNotice {
+    /// The handler acted on this press.
+    ///
+    /// Clears the escalation count, so the next press starts a fresh ladder
+    /// instead of bypassing the handler stack.
+    pub fn handled(self) {
+        self.inner.reset_escalation();
```
JeanMertz: nit: A delayed notice can clear a newer unanswered press. For example, if a handler is slow or its notice is queued, the user can press Ctrl-C, wait past the cooldown, and press again; that second press starts a fresh ladder. If the first notice then reaches `handled()`, this unconditional reset erases the second press, so the next Ctrl-C is routed as another first press instead of shutting down. Carry a press generation in `InterruptNotice` and reset only when no later press has been recorded. A regression test can route at `t0`, route again after the cooldown, handle the first notice, then assert that a third press inside the second window escalates.

crates/jp_cli/src/signals.rs
```diff
@@ -161,22 +175,53 @@ impl SignalRouter {
...
     ///
-    /// The router notifies the next handler on the stack, or falls back to
-    /// graceful shutdown when no other handler exists.
-    pub fn decline(&self) {
+    /// Notifies the next handler down the stack, or requests a graceful
+    /// shutdown when this was the last one.
+    /// The press keeps its place on the escalation ladder: nothing has acted on
+    /// it yet.
+    pub fn decline(self) {
         self.inner.notify_next_or_shutdown();
```
JeanMertz: theoretical: The notice does not record which handler received it, so `notify_next_or_shutdown()` always routes relative to the stack's current top. If a newly registered handler appears before this notice is declined, the press can be sent back to the declining handler; similarly, if the next handler also declines while the original top remains registered, it is notified again rather than progressing down the stack. I could not find a current call path that creates either ordering, so this is not merge-blocking. Storing the receiving `HandlerId` in `InterruptNotice` and routing below that ID would make the documented contract hold independently of stack changes and repeated declines.

crates/jp_cli/src/cmd/query/turn_loop.rs
```diff
@@ -493,14 +512,25 @@ pub(super) async fn run_turn_loop(
...
                                                 // commits partial output as
                                                 // continuation context and breaks
                                                 // for a fresh request; a
                                                 // keep-polling Continue cannot
                                                 // occur here.
+                                                // A menu that could not run also
+                                                // breaks: the stream is dead, so
+                                                // there is nothing to resume.
                                                 StreamingInterruptResult::Continue
+                                                | StreamingInterruptResult::PromptFailed
```
JeanMertz: nit: With a transient provider error, Ctrl-C during retry backoff, and a prompt returning `NotTTY` or an I/O error, this immediately starts the next provider request. That shortens the configured backoff even though `PromptFailed` means no decision was made. Preserve or restart the remaining backoff before retrying, or carry enough timing state for this caller to do so.

PR #990, opened 2026-08-17
fix(config): Inherit tool style from `'*'` field by field

JeanMertz:
`[conversation.tools.'*'.style]` was silently ignored by any tool that set a style field of its own. `ToolConfig::style` is an `Option<DisplayStyleConfig>`, so the moment a tool declares one key the option is `Some` and every other field reads as deliberately set. Every neighbouring accessor fills from the defaults per field; only `style` fell back as a unit.

The practical effect was that a one-line default did nothing:

    [conversation.tools.'*'.style]
    hidden = true

Every tool that already set `style.inline_results` took its own style wholesale and dropped the default, with no error and no warning. Hiding read-only tool calls across a persona meant naming each tool by hand.

`PartialToolsConfig::fill_from` now fills each tool's style from the resolved `*` block. Doing that at partial-merge time rather than in `ToolConfigWithDefaults::style` is what makes it possible: the partial still records which keys the tool asked for, so the gaps are identifiable. Once the config is resolved that information is gone.


crates/jp_config/src/conversation/tool.rs
```diff
@@ -82,9 +82,28 @@ impl PartialConfigDelta for PartialToolsConfig {
...
+        // indistinguishable from the ones it did. Fill the gaps from the `*`
+        // block here, while the partial still records what the tool asked for,
+        // so a single `[conversation.tools.'*'.style]` key reaches every tool.
+        let tools = self
+            .tools
+            .into_iter()
+            .map(|(name, mut tool)| {
+                tool.style = tool
+                    .style
+                    .map(|style| style.fill_from(tool_defaults.style.clone()));
```
JeanMertz: blocking: A conversation with a tool that sets `style.inline_results`, followed by a query using `--cfg "conversation.tools.*.style.hidden=true"`, triggers a persistent precedence bug here. The resolved config's `to_partial()` reports this inherited `hidden = true` as a per-tool value, so `get_config_delta_from_cli` records both the `*` default and a tool-specific override. A later query using the same flag with `hidden=false` changes the default, but the generated per-tool `true` still wins, so the tool stays hidden with no warning. Keep the per-tool field presence separate through resolution, as `enable` does, and add a regression test that resolves, round-trips through `to_partial`/the conversation delta path, then changes the `*` default.

docs/ticket/0011-tool-style-does-not-inherit-field-by-field-from-the-defaults.md
```diff
@@ -0,0 +1,57 @@
...
+
+Found while hiding read-only tool calls for the RFD pipeline personas, which
+ended up enumerating twenty-two tools by name because the one-line version does
+not work.
+That list will drift as tools are added.
+
+## Fix
+
+`DisplayStyleConfig` already implements `FillDefaults`, and
+`PartialToolConfig::fill_from` already calls `style.fill_from(defaults.style)`
```
JeanMertz: nit: `PartialToolConfig` has no `FillDefaults` implementation, so this describes a method that does not exist. The existing fill at `tool.rs:325` is `PartialToolsDefaultsConfig` resolving the global style itself; the cross-key behavior added here lives in `PartialToolsConfig::fill_from`. Please update this completed ticket to describe the actual path.

crates/jp_config/src/conversation/tool.rs
JeanMertz: Confirmed, and it's worse than the `--cfg` sequence you describe. Fixed, though not the way you proposed.

The mechanism: `ToolConfig::to_partial` calls `partial_opt_config(self.style.as_ref(), defaults.style)` where `defaults` is `PartialToolConfig::default()`, whose `style` is `None`, so any resolved `Some(style)` round-trips through `DisplayStyleConfig::to_partial` where `partial_opt(x, None)` always yields `Some`. Every field comes back as an explicit per-tool key. `fill_from` then prefers that key regardless of which layer produced it, so the `*` block can't reach it.

The larger symptom: the fix didn't reach existing conversations at all. `ConversationStream::config` resolves the stream's base config to an `AppConfig`, and the conversation layer is that config's `to_partial()`. For every tool that already declares a style key, that layer carries `hidden`, `inline_results`, `results_file_link` and `parameters` frozen at conversation-creation time, so a `*` block added afterwards filled nothing for those tools. Your `--cfg` scenario is one instance of it.

On the prescribed fix: keeping per-field presence through resolution works for `enable`/`run`/`result`/`format` only because those are `Option<T>` in the resolved struct. `style` is `Option<DisplayStyleConfig>` with non-`Option` fields, and `style()` returns `&DisplayStyleConfig` to six call sites in `jp_cli`. Restructuring the resolved type is a large change for a defect whose mechanism is in `to_partial`, not in the fill.

`AppConfig::to_partial` already solves exactly this problem one level up: it subtracts the assistant from `conversation.inquiry.assistant`, with a comment noting that a recorded copy of an inherited value pins it against a later change to the thing it was inherited from. `ToolsConfig::to_partial` now does the same, subtracting the resolved `*` style from each tool's style and collapsing a fully-stripped result to `None`:

```rust
tool.style = tool
    .style
    .map(|style| defaults.style.delta(style))
    .filter(|style| !style.is_empty());
```

Cost, stated plainly: a tool that deliberately sets a field to the value the `*` block already holds is indistinguishable from one inheriting it, and follows a later `*`-only change. That is the same trade the inquiry code accepts, and it's recorded in the ticket's "Watch for" section.

Three tests, each seen red against the unstripped `to_partial` first:

- `resolved_tool_style_omits_fields_taken_from_the_global_defaults` (was: `hidden: Some(true), results_file_link: Some(Full), parameters: Some(Json)`)
- `a_global_style_change_reaches_a_tool_through_a_resolved_config` (the existing-conversation path; was: `assertion failed: tool.style().hidden`)
- `a_global_style_change_is_not_recorded_as_a_per_tool_override` (your delta path; was: `my_tool` present with `style: Some(PartialDisplayStyleConfig { hidden: Some(true), .. })`)

docs/ticket/0011-tool-style-does-not-inherit-field-by-field-from-the-defaults.md
JeanMertz: Correct on the facts: `grep` finds `FillDefaults` impls for `PartialToolsConfig`, `PartialToolsDefaultsConfig` and `PartialDisplayStyleConfig`, and none for `PartialToolConfig`. `tool.rs:325` is the `*` block filling from a lower layer's `*` block, a different relationship from the one that paragraph claimed.

That paragraph had already been rewritten in my working tree when this landed (along with the ticket's rename to `042d6hq-…`), so what's in the PR now says what you're asking for: `PartialToolsDefaultsConfig::fill_from` resolves the `*` block's own style, there is no `PartialToolConfig::fill_from`, and nothing filled a tool's style from `*`.

I've extended it further while addressing the other comment. The "Fix" section no longer offers the two options as an open choice (the ticket is `Done`, it should say what shipped): the fill happens in `PartialToolsConfig::fill_from`, and a new paragraph covers the second half, the `ToolsConfig::to_partial` subtraction that keeps a filled-in field from being republished as a key of the tool's own. "Watch for" now also records the equality caveat that subtraction carries. Dropped the bare `tool.rs:325` reference, since line numbers in a completed ticket drift on the next edit to the file.

docs/ticket/042d6hq-tool-style-does-not-inherit-field-by-field-from-the-defaults.md
```diff
@@ -0,0 +1,89 @@
...
+- **Status**: Done
+- **Kind**: Bug
+- **Authors**: jp
+- **Date**: 2026-08-11
+
+`[conversation.tools.'*'.style]` is silently ignored for any tool that sets a
+single style field of its own.
+
+`ToolConfigWithDefaults::style` resolves the whole struct at once
+(`crates/jp_config/src/conversation/tool.rs:1215`):
```
JeanMertz: nit: This points to `tool.rs:1215`, but `ToolConfigWithDefaults::style` starts at line 1256 in this PR. Refer to the symbol without a line number so edits to the file do not leave this completed ticket with a stale location.

PR #992, opened 2026-08-17
chore(tools, ticket): Preview the ticket a call will write

JeanMertz:
Approving a `ticket_create` or `ticket_comment` call meant reading its arguments
as a function signature and imagining the file. Both now answer the
format-arguments action by rendering the document in the shape it takes on disk,
quoted as a blockquote so the marker runs down the whole preview and the reader
can see where the ticket ends and the conversation resumes.

`style.parameters` points at the tool's own command so JP shells out to render
the call, and `format = "unattended"` lets that run before the approval prompt
rather than after it. That mode carries a contract: the formatter must be
side-effect-free. A create preview keeps it by rendering the document and
nothing else — no file, and no id drawn. Since RFD 102 the id lives only in the
filename, so the previewed document is complete without one, and the result line
names the id the ticket ends up with.

`render::comment` renders one comment block standalone, which appending never
needed. A comment preview shows it under the heading of the ticket it lands on,
and says so when that ticket does not exist rather than previewing a call that
is going to fail.


.config/jp/tools/src/ticket.rs
```diff
@@ -185,6 +255,36 @@ fn relative(root: &Utf8Path, path: &Utf8Path) -> String {
         .replace(MAIN_SEPARATOR, "/")
 }
 
+/// The title of a ticket on disk, if it is there and readable.
+fn title_of(root: &Utf8Path, id: TicketId) -> Option<String> {
+    let path = store::locate_ticket(&dir(root), id).ok()?;
+    let document = fs::read_to_string(path).ok()?;
+
+    parse::document(&document).ok().map(|ticket| ticket.title)
```
JeanMertz: blocking: A hand-edited ticket with a missing `Authors` metadata line triggers a false warning here. `parse::document` returns `Err`, so the preview says `No T-...; this call will fail`, but `store::append_comment` intentionally uses the header-tolerant `parse::comment_count` and will append successfully. The user can reject a valid write based on plausible-looking wrong approval information. Extract the title without requiring a fully valid ticket, or make the preview preflight match `append_comment`, and cover a malformed-header ticket in the preview tests.

.config/jp/tools/src/ticket.rs
```diff
@@ -35,21 +40,39 @@ pub fn run(ctx: Context, t: Tool) -> ToolResult {
...
         "comment" => {
             let id = match id_arg(&t.req("id")?) {
                 Ok(id) => id,
                 Err(message) => return error(message),
             };
-            comment(root, id, t.opt("re")?, &t.req::<String>("body")?)
+            let re = t.opt("re")?;
+            let body = t.req::<String>("body")?;
+
+            if ctx.action.is_format_arguments() {
```
JeanMertz: nit: A model can request `re: 3` on a ticket with one comment. This branch renders that reply as valid and asks for approval, then execution rejects it as `NoSuchComment`; a whitespace-only `body` follows the same path. The failure is contained and visible, but the preview should run the same body and reply-target checks as `comment` so impossible calls fail before the prompt. A `FormatArguments` regression test for each check would pin that parity.

.config/jp/tools/src/ticket.rs
```diff
@@ -35,21 +40,39 @@ pub fn run(ctx: Context, t: Tool) -> ToolResult {
...
-                kind,
-                &t.req::<String>("title")?,
-                t.opt::<String>("implements")?.as_deref(),
-                t.opt("body")?,
-            )
+            let title = t.req::<String>("title")?;
+            let implements = t.opt::<String>("implements")?;
+            let body = t.opt::<String>("body")?;
+
+            if ctx.action.is_format_arguments() {
```
JeanMertz: nit: A plausible `ticket_create` call with `title: "   "` gets a successful empty-heading preview here, asks the user for approval, and only then fails the check in `create`. Move the non-empty title check before the action split, or share a pure validator between preview and execution, so the preview cannot present a call that execution rejects. Please add the `FormatArguments` counterpart to `create_rejects_an_empty_title`.

JeanMertz: Taken, moved above the split.

I initially argued against this on the grounds that an error from a custom formatter suppresses the tool-call display entirely, so the user would see nothing instead of a broken preview. That reading was incomplete. `coordinator.rs:420` documents the intent: an `Err` from the formatter means the caller should "treat this as a tool failure and skip prompting", and `resolve_tool_call_decision` returns `ToolCallDecision::Failed` at line 514 *before* `prompt_permission`. Under `format = "unattended"` that is the desirable outcome — the assistant is told the call was not executed and may retry, and the user is never asked to approve a call that cannot land. `debug_jp_profile_heap` and `fs_create_file` already validate ahead of the action split for the same reason.

The check now lives in the `create` arm before `is_format_arguments()`, and the duplicate in `create` is gone. `preview_rejects_the_arguments_execution_would_reject` covers it through `Action::FormatArguments` alongside the `body` case.

.config/jp/tools/src/ticket.rs
JeanMertz: Both taken, in the two places they belong.

The `body` check moved above the action split next to the `title` one, so it fails for either action.

The reply target can't be validated from the arguments alone, so it moved into `preview_comment`, which now reads the ticket anyway. It uses `parse::comment_count` on the same document, which is exactly what `store::append_comment` counts with, so the preview and the write cannot disagree about what `#3` means:

```rust
let count = parse::comment_count(&document);
if let Some(position) = re
    && (position == 0 || position > count)
{
    return error(format!("No comment #{position} on {id}."));
}
```

The preview can also be more specific than the write path: `append_comment` collapses both failures into "No {id}, or no comment #{n} on it." because it can't distinguish them cheaply, while the preview checks them separately and names the one that applies.

`comment_preview_rejects_a_missing_reply_target` pins it. Writing it also caught that `comment_preview_renders_the_block_under_the_ticket_it_lands_on` was previewing a reply to `#1` on a ticket with no comments, which the new check correctly rejects — that test now creates the comment it replies to.

.config/jp/tools/src/ticket.rs
JeanMertz: Confirmed and fixed, and the fix took both halves you offered.

The mismatch was exactly as described: `metadata()` (`parse.rs:232-235`) returns `MissingField("Authors")`, so `parse::document` fails, `title_of` yields `None`, and the preview claimed the call would fail — while `append_comment` (`store.rs:265-267`) only calls `locate` and `parse::comment_count`, whose doc comment says it is "tolerant of a malformed header, so it can run on a file that is about to be appended to rather than validated". The write succeeds. `list_names_the_files_it_could_not_read` already writes a mangled ticket on purpose, so a file in that state is not hypothetical.

Two changes:

- `parse::title(source) -> Option<String>` exposes the existing `Doc::title` as a tolerant reader, with a doc comment pointing at `comment_count` for the same rationale. Re-deriving "first `# ` line" in the tools crate would have duplicated logic the format's crate owns. `reads_the_title_without_validating_the_header` asserts the same document that `document()` rejects still yields a title.
- The `⚠ … this call will fail` branch is gone entirely. Existence is decided by `store::locate_ticket`, which is what `append_comment` locates with, and a missing ticket now returns an error from the preview rather than rendering a warning — so the assistant gets corrected before the prompt instead of the user being asked to approve a doomed call. A ticket that exists but can't be titled renders a bare `# {id}` heading and no warning, which is honest: the comment will land.

`comment_preview_tolerates_a_malformed_header` covers a ticket with the `Authors` line removed, and asserts the follow-up write really does succeed — otherwise the test would only pin that the preview is optimistic, not that its optimism is justified.

One case I left deliberately: an unreadable file (permissions) makes the preview optimistic where the append would fail on I/O. Matching that exactly means duplicating the read semantics for a case no plausible workflow reaches.

.config/jp/tools/src/ticket.rs
```diff
@@ -123,6 +171,50 @@ fn comment(root: &Utf8Path, id: TicketId, re: Option<usize>, body: &str) -> Tool
...
+/// conditions the append itself rejects, so a call that cannot land is answered
+/// before it is put to the user.
+fn preview_comment(
+    root: &Utf8Path,
+    id: TicketId,
+    re: Option<usize>,
+    body: &str,
+    date: &str,
+) -> ToolResult {
+    let Ok(path) = store::locate_ticket(&dir(root), id) else {
```
JeanMertz: nit: A workspace where two stale branches have landed files with the same ticket id hits this branch. RFD 102 explicitly keeps `store::Error::Duplicate` for that state, including both paths, but the preview discards it and tells the assistant the ticket is missing. The failure is contained and visible, but the wrong diagnostic points at the wrong recovery. Match `NoSuchTicket` here and propagate other `locate_ticket` errors unchanged; a duplicate-id preview test would pin the distinction.

JeanMertz: Taken. `locate` (`store.rs:403-418`) does return `Error::Duplicate { id, paths }`, and its doc comment is explicit that this is "an error rather than a choice: a write by id would otherwise land on whichever one the directory happened to yield first" — so flattening it into `No {id}.` sent the assistant hunting for a different id when the repair is renaming one of two files.

On likelihood alone I'd have left this: a tail collision inside one five-second bucket, in two checkouts, both landed, and `ticket_show` / `ticket_list` already report it correctly since both propagate from `store::list`. What decided it is that the deviation was mine and local to this PR. `close` (`ticket.rs:222-223`) and `comment` (`ticket.rs:166-170`) both name the variant they handle and propagate the rest; `preview_comment` was the only place swallowing the enum whole, so this restores the file's own convention rather than adding a check for a rare input.

```rust
let path = match store::locate_ticket(&dir(root), id) {
    Ok(path) => path,
    Err(store::Error::NoSuchTicket(_)) => return error(format!("No {id}.")),
    Err(other) => return Err(other.into()),
};
```

Propagating rather than stringifying matches `close`, and the text still reaches the assistant: `main.rs` wraps it with the `Duplicate` display in `trace`, and `CommandResult::format_error` (`jp_llm/src/tool.rs:293-299`) appends the trace to the formatter-failure report.

`comment_preview_reports_a_duplicated_id` pins the distinction, asserting on the message prefix the way `store_tests::list_rejects_a_duplicated_id` does, since the paths carry a temp dir. It failed against the old code with `called Result::unwrap_err() on an Ok value: Error { message: "No T-02wt0kx." }`, which is the wrong diagnostic stated plainly.

PR #993, opened 2026-08-18
chore(tools): Add `root` option to the cargo tools

JeanMertz:
The cargo tools ran `cargo` with the workspace root as their working
directory, so they could only ever build the workspace JP itself lives
in. A cargo workspace nested elsewhere in the tree was unreachable: code
written there could be edited but never compiled or tested, which is not
a workable development loop.

Each cargo tool now reads a per-tool `root` option and uses it as the
working directory for `cargo`, and for `comfort` in `check` and
`format`. Resolution goes through `fs::utils::resolve_workspace_path`,
the same resolver the fs tools use, so absolute paths and `..` escapes
are refused and an approved `external` mount stays the only sanctioned
way out of the workspace. This matters more than for a read: `cargo`
runs build scripts and proc macros, so an unvetted directory is
arbitrary code execution. An empty value or `.` returns to the default,
which lets a config layer unload a previously-set root without naming
the default path, needed until negative config deltas land.

Set it per tool in the tool config:

    [conversation.tools.cargo_check.options]
    root = "crates/nested-project"

Failures from a redirected root name the directory cargo ran in.
Without it, building the host workspace while redirected fails with
`package ID specification 'tools' did not match any packages`, which
gives no hint that the redirect is the cause.

The option is read and interpreted entirely within the cargo tools.
Nothing about `Context` changes, so the framework makes no assumptions
about what a tool does with a root, and the free-form options map stays
tool-private as documented.

The five tools now take a `root: &Utf8Path` rather than a `&Context`.
They only ever used `ctx.root`, and passing the path directly avoids
fabricating a `Context` whose `root` would disagree with the workspace
its `access` policy was compiled against.

### DIFF
--- a/.config/jp/tools/src/cargo.rs
+++ b/.config/jp/tools/src/cargo.rs
@@ -1,9 +1,10 @@
 use camino::{Utf8Path, Utf8PathBuf};
-use jp_tool::{AccessPolicy, Outcome};
+use jp_tool::{AccessPolicy, Capability, Outcome};
+use serde_json::Value;
 
 use crate::{
     Context, Tool,
-    fs::utils::resolve_workspace_path,
+    fs::utils::{authorize, resolve_workspace_path},
     util::{ToolResult, error, unknown_tool},
 };
 
@@ -36,13 +37,33 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
     // Which cargo workspace to operate in. Defaults to the root the tool was
     // invoked with; set `options.root` in the tool config to point the cargo
     // tooling at a different cargo workspace.
-    let configured: Option<String> = t.option_or("root", None);
-    let root = match cargo_root(&ctx.root, configured.as_deref(), ctx.access.as_ref()) {
+    //
+    // A present `root` must be a string. `option_or` cannot be used here: it
+    // reports a malformed value as an absent one, which would silently run
+    // `cargo_format` or `cargo_update` against the host workspace — writing to
+    // it, and reporting success — after the user asked for another one.
+    let configured = match t.options.get("root") {
+        None | Some(Value::Null) => None,
+        Some(Value::String(value)) => Some(value.clone()),
+        Some(other) => {
+            return error(format!(
+                "The `root` tool option must be a string, got `{other}`."
+            ));
+        }
+    };
+
+    let subcommand = t.name.trim_start_matches("cargo_");
+    let root = match cargo_root(
+        &ctx.root,
+        configured.as_deref(),
+        ctx.access.as_ref(),
+        required_capabilities(subcommand),
+    ) {
         Ok(root) => root,
         Err(message) => return error(message),
     };
 
-    let outcome = match t.name.trim_start_matches("cargo_") {
+    let outcome = match subcommand {
         "check" => cargo_check(&root, t.opt("package")?, checksum_freshness).await,
         "expand" => {
             cargo_expand(&root, t.req("item")?, t.opt("package")?, checksum_freshness).await
@@ -67,6 +88,24 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
     }
 
     note_root(outcome, &root)
+}
+
+/// Capabilities a cargo subcommand needs on the directory it runs in.
+///
+/// `fmt` and `update` never compile: the first rewrites sources, the second the
+/// lockfile.
+/// The rest build, so they also write artifacts and run build scripts and proc
+/// macros.
+fn required_capabilities(subcommand: &str) -> &'static [Capability] {
+    match subcommand {
+        "format" | "update" => &[Capability::Read, Capability::Update],
+        _ => &[
+            Capability::Read,
+            Capability::Create,
+            Capability::Update,
+            Capability::Execute,
+        ],
+    }
 }
 
 /// Name the directory cargo ran in, for failures from a redirected root.
@@ -111,10 +150,16 @@ fn note_root(outcome: ToolResult, root: &Utf8Path) -> ToolResult {
 /// The target is required to be a directory, so a typo (or naming the manifest
 /// instead of the directory holding it) surfaces here rather than as a
 /// confusing cargo failure somewhere else.
+///
+/// A configured target must also grant `capabilities` outright.
+/// `external` only permits a path to resolve outside the workspace; it is not a
+/// capability grant, so reaching a mount says nothing about what may be done
+/// once there.
 fn cargo_root(
     default: &Utf8Path,
     configured: Option<&str>,
     access: Option<&AccessPolicy>,
+    capabilities: &[Capability],
 ) -> Result<Utf8PathBuf, String> {
     // An empty value or `.` resolves to the invocation root, so a config layer
     // can unload a previously-set root without naming the default.
@@ -131,6 +176,13 @@ fn cargo_root(
             "The `root` option `{configured}` resolved to `{}`, which is not a directory.",
             resolved.absolute
         ));
+    }
+
+    // Only a configured root is authorized. The invocation root is where these
+    // tools have always run, so demanding grants for it here would revoke
+    // access this option never handed out.
+    for capability in capabilities {
+        authorize(access, *capability, &resolved.relative)?;
     }
 
     Ok(resolved.absolute)
--- a/.config/jp/tools/src/cargo.rs
+++ b/.config/jp/tools/src/cargo.rs
@@ -1,7 +1,9 @@
 use camino::{Utf8Path, Utf8PathBuf};
+use jp_tool::{AccessPolicy, Outcome};
 
 use crate::{
     Context, Tool,
+    fs::utils::resolve_workspace_path,
     util::{ToolResult, error, unknown_tool},
 };
 
@@ -35,12 +37,12 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
     // invoked with; set `options.root` in the tool config to point the cargo
     // tooling at a different cargo workspace.
     let configured: Option<String> = t.option_or("root", None);
-    let root = match cargo_root(&ctx.root, configured.as_deref()) {
+    let root = match cargo_root(&ctx.root, configured.as_deref(), ctx.access.as_ref()) {
         Ok(root) => root,
         Err(message) => return error(message),
     };
 
-    match t.name.trim_start_matches("cargo_") {
+    let outcome = match t.name.trim_start_matches("cargo_") {
         "check" => cargo_check(&root, t.opt("package")?, checksum_freshness).await,
         "expand" => {
             cargo_expand(&root, t.req("item")?, t.opt("package")?, checksum_freshness).await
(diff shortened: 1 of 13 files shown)


.config/jp/tools/src/cargo.rs
```diff
@@ -29,21 +33,109 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
     // config.
     let checksum_freshness = t.option_or("checksum_freshness", false);
 
-    match t.name.trim_start_matches("cargo_") {
-        "check" => cargo_check(&ctx, t.opt("package")?, checksum_freshness).await,
-        "expand" => cargo_expand(&ctx, t.req("item")?, t.opt("package")?, checksum_freshness).await,
+    // Which cargo workspace to operate in. Defaults to the root the tool was
+    // invoked with; set `options.root` in the tool config to point the cargo
+    // tooling at a different cargo workspace.
+    let configured: Option<String> = t.option_or("root", None);
```
JeanMertz: blocking: A plausible config mistake such as `root = 1` reaches this line. `option_or` treats a deserialization failure exactly like an absent option, so `cargo_update` or `cargo_format` runs against `ctx.root` and can persist changes to the host workspace while reporting success. Parse a present `root` fallibly and return a configuration error; reserve `None` for an option that is actually absent.

.config/jp/tools/src/cargo.rs
```diff
@@ -29,21 +33,109 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
...
+) -> Result<Utf8PathBuf, String> {
+    // An empty value or `.` resolves to the invocation root, so a config layer
+    // can unload a previously-set root without naming the default.
+    let configured = configured.filter(|value| !value.is_empty() && *value != ".");
+
+    let Some(configured) = configured else {
+        return Ok(default.to_owned());
+    };
+
+    let resolved = resolve_workspace_path(default, configured, access)?;
```
JeanMertz: blocking: A plausible setup is an external mount `fork` granted only `read = true`, with `cargo_test.options.root = "fork"`. This resolver accepts the mount based on `external` and `approved_target` alone, then Cargo writes build artifacts and runs build scripts, proc macros, and tests even though create, update, and execute are denied. That turns `external` into a capability grant, contrary to `FsRuleConfig`'s contract. Before spawning Cargo, authorize the redirected root for the operations that command can perform, or require an explicit broad grant for running a subprocess workspace.

.config/jp/tools/src/cargo.rs
JeanMertz: Accepted, fixed.

Confirmed the mechanism: `Tool::option_or` (`lib.rs:80-88`) is documented as returning the default "if the key is missing **or the value can't be deserialized to `T`**", and discards the error with `.ok()`. With `T = Option<String>` a non-string yields `None`, indistinguishable from absent, so `cargo_root` returns `default`.

One thing that makes it worse than described: `note_root` only annotates when `root != ctx.root`, so a silent fallback also suppresses the one signal that would reveal the redirect didn't happen. The safety net is disabled by the same bug.

`run()` now reads the option directly (matching `fs.rs:34` and `git.rs:40`) and rejects a present non-string:

```rust
let configured = match t.options.get("root") {
    None | Some(Value::Null) => None,
    Some(Value::String(value)) => Some(value.clone()),
    Some(other) => {
        return error(format!(
            "The `root` tool option must be a string, got `{other}`."
        ));
    }
};
```

Residual gap worth naming: this does not catch a *misspelled* key (`roots = "..."`), which produces the identical silent fallback and is more plausible than a wrong type. Unknown options are silently forwarded by design (`jp_config/src/conversation/tool.rs:448-452`), so it can't be closed inside the tool.

.config/jp/tools/src/cargo.rs
JeanMertz: Accepted the finding, amended the fix.

The contract violation is real and verifiable three ways:

1. `FsRuleConfig::external` says it outright — "This is not a capability grant — it only permits the path to resolve externally" (`jp_config/src/conversation/tool/access.rs:105-112`).
2. `resolve_workspace_path` never authorizes; it validates input and bounds the external escape.
3. Every fs tool pairs resolution with an explicit `authorize` call (`create_file.rs:53`, `delete_file.rs:23`, `modify_file.rs:115`, `move_file.rs:59` and `:104`, `read_file.rs:18`). `cargo_root` was the only consumer of that resolver skipping the second half of the pattern.

Two amendments to the proposed fix:

**Only a configured root is authorized.** Applying it to the default `ctx.root` too would revoke access this option never handed out: a tool carrying a narrow pre-existing `access.fs` policy would suddenly fail to run cargo at the workspace root. That's a behaviour change beyond this PR's reach. Covered by `the_invocation_root_is_not_authorized`.

**Per-subcommand capability sets, not one blanket set.** `fmt` and `update` never compile — the first rewrites sources, the second the lockfile. Demanding `Execute` from them would reject a legitimate read-plus-update mount for no safety gain:

```rust
fn required_capabilities(subcommand: &str) -> &'static [Capability] {
    match subcommand {
        "format" | "update" => &[Capability::Read, Capability::Update],
        _ => &[Capability::Read, Capability::Create, Capability::Update, Capability::Execute],
    }
}
```

`cargo_root` then authorizes each capability against `resolved.relative` after the directory check.

The test that actually pins the contract is `a_configured_root_needs_the_capabilities_its_subcommand_uses`: a mount granting `read` + `update` permits `cargo_format` and denies `cargo_check`.

.config/jp/tools/src/cargo.rs
```diff
@@ -29,21 +34,160 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
...
+
+/// Capabilities a cargo subcommand needs on the directory it runs in.
+///
+/// `fmt` and `update` never compile: the first rewrites sources, the second the
+/// lockfile.
+/// The rest build, so they also write artifacts and run build scripts and proc
+/// macros.
+fn required_capabilities(subcommand: &str) -> &'static [Capability] {
+    match subcommand {
+        "format" | "update" => &[Capability::Read, Capability::Update],
```
JeanMertz: blocking: A plausible redirected workspace has no `Cargo.lock` yet, and its rule grants `read` + `update` but denies `create`. Cargo explicitly creates `Cargo.lock` when `cargo update` runs without one, so this mapping authorizes a command that performs the denied operation and leaves a new file behind. Give `update` its own arm that also requires `Capability::Create`; `format` can keep the narrower pair.

.config/jp/tools/src/cargo.rs
```diff
@@ -29,21 +34,160 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
...
+        _ => &[
+            Capability::Read,
+            Capability::Create,
+            Capability::Update,
+            Capability::Execute,
+        ],
```
JeanMertz: blocking: A plausible `cargo_test` target contains a test or build script that deletes generated state while the root rule grants these four capabilities but explicitly denies `delete`. These subprocesses are not sandboxed, so the deletion succeeds despite the policy. Since this branch intentionally models everything workspace code can do, it also needs `Capability::Delete`, or the tool needs a distinct explicit grant for unrestricted subprocess filesystem effects.

.config/jp/tools/src/cargo.rs
```diff
@@ -29,21 +34,160 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
...
-        "expand" => cargo_expand(&ctx, t.req("item")?, t.opt("package")?, checksum_freshness).await,
+    // Which cargo workspace to operate in. Defaults to the root the tool was
+    // invoked with; set `options.root` in the tool config to point the cargo
+    // tooling at a different cargo workspace.
+    //
+    // A present `root` must be a string. `option_or` cannot be used here: it
+    // reports a malformed value as an absent one, which would silently run
+    // `cargo_format` or `cargo_update` against the host workspace — writing to
+    // it, and reporting success — after the user asked for another one.
+    let configured = match t.options.get("root") {
```
JeanMertz: nit: The `root = 1` case caused a hidden wrong-workspace write, but the fix has no regression test through `run`. Add a `#[tokio::test]` that passes a non-string `options.root` and asserts an error outcome. It returns before spawning Cargo, so the test needs no process fixture, and it prevents this path from being folded back into `option_or` later.

.config/jp/tools/src/cargo.rs
```diff
@@ -29,21 +34,160 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
...
+    // can unload a previously-set root without naming the default.
+    let configured = configured.filter(|value| !value.is_empty() && *value != ".");
+
+    let Some(configured) = configured else {
+        return Ok(default.to_owned());
+    };
+
+    let resolved = resolve_workspace_path(default, configured, access)?;
+
+    if !resolved.absolute.is_dir() {
```
JeanMertz: blocking: A plausible typo such as `root = "crates/new-project"` can name an existing directory before that project has a `Cargo.toml`. Cargo searches parent directories for a manifest, so every command then falls through to JP's workspace; `cargo_format` or `cargo_update` can modify it and return success without the root note. Validate that the selected directory contains the intended `Cargo.toml`, or invoke Cargo with an explicit manifest and reject a workspace root that resolves elsewhere.

.config/jp/tools/src/cargo.rs
JeanMertz: Accepted — and it's under-labelled as a nit by this codebase's own standard.

There's direct precedent I should have followed. `fs_tests.rs:26-45` tests a malformed `suppress` option by driving `fs::run`, and its comment makes exactly this argument:

> `Tool::option_or` degrades exactly that way, so reading this option through it satisfies every other test while breaking the guarantee; **only driving `run` itself catches that**.

So the convention for option-parsing degradation was already established, and my unit tests covered `cargo_root` and `required_capabilities` while leaving the parse — the thing that was actually broken — untested. A refactor back to `option_or` would have kept the suite green.

Added `a_non_string_root_option_fails_the_invocation`, mirroring the `fs_tests.rs` shape. It confirms the early return at `cargo.rs:45-53` precedes both `cargo_root` and dispatch, so no process fixture is needed. One difference from the `fs` case: `error()` in the cargo tools returns `Ok(Outcome::Error { .. })` rather than `Err`, so the test destructures the outcome instead of using `expect_err`.

.config/jp/tools/src/cargo.rs
JeanMertz: Accepted, fixed.

Correct: `cargo update` resolves and writes `Cargo.lock` when the target has none, so declaring only `Read` + `Update` let the gate pass a command that then creates a file the policy denied.

The tool's own code already anticipated the missing-lockfile state — `update.rs` reads both `before` and `after` with `fs::read_to_string(&lockfile).ok()`, so a `None → Some` transition is an expected outcome. I wrote the capability set from an assumption about what `update` does rather than from what the code plainly showed.

`update` now has its own arm:

```rust
// Rewrites the lockfile, and creates it when the target has none.
"update" => &[Capability::Read, Capability::Create, Capability::Update],
```

`format` keeps the narrower pair, as you suggested — rustfmt and comfort rewrite existing files and create nothing. Covered by `updating_needs_create_for_a_missing_lockfile` and `formatting_needs_no_more_than_read_and_update`, which replace the combined assertion that previously locked both to `[Read, Update]`.

On severity: I'd call this contained and visible rather than blocking — one file, inside the named directory, and `cargo_update` reports the lockfile diff. But it was a false claim in a security-relevant declaration, which is the same defect class as the `external` finding, so it wanted fixing regardless of label.

.config/jp/tools/src/cargo.rs
JeanMertz: Accepted, fixed — and it doesn't even need adversarial test code to bite.

Cargo itself removes stale artifacts under `target/` during a build, so the building branch performs deletions as ordinary operation. On top of that, `cargo test` compiles and runs build scripts, proc macros and test binaries: arbitrary code carrying the process's own filesystem access. `AccessPolicy` is cooperative (enforced by tools calling `authorize`), not an OS sandbox, so nothing constrains any of it.

`Delete` added to the building arm. I took your first option rather than the second: a distinct grant for "unrestricted subprocess filesystem effects" is a new capability concept with its own config surface, and that belongs with the sandbox work in [RFD 075], not here.

Worth stating explicitly, since the fix could otherwise read as more than it is: even `Read + Create + Update + Delete + Execute` on the target directory **understates** reality, because an unsandboxed subprocess can touch paths outside it entirely. These sets gate whether cargo is spawned at all; they cannot bound what it does once running. I've written that into the doc comment on `required_capabilities` so the next reader doesn't mistake the enumeration for a confinement.

Covered by `building_subcommands_declare_every_capability`, which asserts all five capabilities across `check`, `test` and `expand`.

[RFD 075]: https://github.com/dcdpr/jp/blob/main/docs/rfd/075-tool-sandbox-and-access-policy.md

.config/jp/tools/src/cargo.rs
JeanMertz: Accepted the finding, amended the fix. This was the most serious of the batch.

Verified every step: `cargo_root` only checked `is_dir()`, cargo's manifest discovery walks ancestors, and `note_root` annotates only `Outcome::Error`/`Err` — successes fall through untouched via `ok => ok`. So a fall-through produced a silent, successful mutation of the host workspace.

The input is easier to hit than your example, too: `root = "crates"` as a typo for `crates/exodus/exodus`. That directory exists, has no manifest, and `cargo_format` would then rustfmt the entire host workspace and report success.

Amended on the second option: `--manifest-path` does **not** close this. It pins the package, but cargo still resolves the *workspace* from that manifest's membership, so a member manifest still resolves to the outer workspace. It's also not uniformly supported across the tools this would touch (`comfort` discovers from cwd).

Went with your first option — require a manifest in the resolved directory:

```rust
if !resolved.absolute.join("Cargo.toml").is_file() {
    return Err(format!(
        "The `root` option `{configured}` resolved to `{}`, which has no `Cargo.toml`. Cargo \
         would search parent directories and operate on the enclosing workspace instead.",
        resolved.absolute
    ));
}
```

Covered by `a_directory_without_a_manifest_is_rejected`. Two existing tests built bare directories and now go through a `package_dir` helper that writes a manifest.

**Residual, deliberately not fixed here.** This closes "no manifest anywhere" but not "manifest present, but its workspace root lies outside the target". Concretely, a directory whose `Cargo.toml` has no `[workspace]` and no workspace root above it inside its own tree will still resolve up into the enclosing workspace. Closing that means verifying cargo's chosen root (`cargo locate-project --workspace`) and rejecting a mismatch, which turns `cargo_root` from a pure function into one needing a `ProcessRunner` for testability. Happy to do it as a follow-up if you want it airtight — I left it out because the input is a deliberate long path rather than the easy typo this fix catches.

PR #994, opened 2026-08-18
feat(config, conversation, cli): Compact only oversized items

JeanMertz:
A compaction rule reaches every item in its turn range. That is the
right default for old turns, but it is the wrong tool for the case that
actually breaks a conversation: one `fs_read_file` that returned a 10 MB
log and now dominates the context window. Cutting it today means either
compacting a whole range of turns you wanted to keep, or editing
`events.json` by hand.

The reasoning and tool-call policies now accept an `over` size
threshold, so a rule reaches only the items large enough to matter:

    jp conversation compact --tools=sres --over 1mb
    jp query -k 'r,over=16kb+t=sres,over=1mb:..-3'

    [[conversation.compaction.rules]]
    tool_calls = { policy = "strip-responses", over = "1MB" }

Both surfaces compose with the existing range flags, so a threshold
narrows what a rule touches rather than replacing how it is scoped. This
also enables a tiered rule set: drop anything genuinely huge wherever it
sits, and strip everything older than the last few turns regardless of
size.

The threshold is stored in the compaction event and evaluated during
projection rather than resolved into a fixed set of matched items at
creation. Size comparison is cheap and deterministic, so it is a lazy
policy in RFD 064's terms, and projection stays a pure function of the
stored stream.

Sizes are judged per half for `Strip`, so a call with a short request
and a huge response loses only the response. `Omit` removes whole pairs,
so it is judged on the two halves combined and can never leave an
orphaned request behind. Measurement uses decoded content (tool
arguments are base64 at rest) against the raw stream, so re-running a
rule reaches the same items. `summary` takes no threshold: it replaces
its whole range rather than selecting from it.

Because a threshold reaches an unpredictable subset of its range, the
timeline names what it caught instead of leaving the range line to imply
everything was compacted:

    Would have compacted turns 2..14 (13 total, tool responses over 1MB).
      turn 4  fs_read_file (response)  10.3 MB
      turn 9  cargo_test (response)  2.1 MB

A rule without a threshold does not itemize, and a threshold that
matched nothing says so.

Nothing changes for a rule that sets no threshold, on disk or in
behavior: a policy without one serializes as the bare string it always
was, and the built-in defaults carry no threshold.

BREAKING CHANGE: `reasoning` in `jp conversation show -F json`

The `reasoning` field of a compaction entry was a boolean; it is now the
serialized policy (`null`, `"strip"`, or `{"policy": "strip", "over":
"16KB"}`), matching how `tool_calls` already reported itself. Without
this a consumer could read `tool_calls.over` but had no way to see a
reasoning threshold.

Scripts testing truthiness need updating: `.reasoning == true` becomes
`.reasoning != null`.


crates/jp_conversation/src/stream/projection.rs
```diff
@@ -426,17 +560,105 @@ fn strip_tool_response(event: &mut ConversationEvent, tool_names: &HashMap<Strin
...
+            continue;
+        };
+
+        if let Some(req) = conv_event.as_tool_call_request() {
+            let call = calls.entry(req.id.clone()).or_default();
+            call.name.clone_from(&req.name);
+            call.request_size = arguments_size(&req.arguments);
+        } else if let Some(resp) = conv_event.as_tool_call_response() {
+            calls.entry(resp.id.clone()).or_default().response_size =
+                byte_count(resp.content().len());
```
JeanMertz: blocking: A Google/Gemini conversation can reuse the same synthetic call ID across cycles, which `TurnMut` explicitly permits. If the first `tc1` response is 4 MB and a later `tc1` response is 2 bytes, this map entry ends with the later size, so `--tools=sres --over 1MB` keeps the 4 MB response and reports no match. That oversized content remains in subsequent projected queries. Pair requests and responses by turn plus occurrence (or by event position), rather than by ID alone, and add a regression test with reused IDs of different sizes.

crates/jp_config/src/types/policy_spec.rs
```diff
@@ -0,0 +1,201 @@
...
+        // is not knowable from here, so try the whole value first and fall back
+        // to the promoted string. Trying the map first matters: a tagged unit
+        // variant (`{"policy": "omit"}`) is indistinguishable from a promotion
+        // by shape alone, and only the map reading is correct for it.
+        let policy = match serde_json::from_value::<P>(value.clone()) {
+            Ok(policy) => policy,
+            Err(error) => {
+                let promoted = value
+                    .as_object()
+                    .and_then(|map| map.get("policy"))
```
JeanMertz: blocking: A plausible typo such as `tool_calls = { policy = "strip-responses", oer = "1MB" }` reaches this fallback, extracts `policy`, and silently ignores the leftover `oer` key. JP then accepts an unthresholded rule and strips every response in the range, rather than only oversized responses. The surrounding config types reject unknown fields, so this silently broadens compaction where users expect validation. After removing `over`, reject leftover keys in the promoted-string form and cover the typo case with a TOML test.

crates/jp_config/src/types/policy_spec.rs
```diff
@@ -0,0 +1,201 @@
...
+        let mut object = match serde_json::to_value(&self.policy).map_err(S::Error::custom)? {
+            Value::Object(map) => map,
+            bare => Map::from_iter([("policy".to_owned(), bare)]),
+        };
+
+        object.insert(
+            "over".to_owned(),
+            serde_json::to_value(over).map_err(S::Error::custom)?,
+        );
+
```
JeanMertz: blocking: A user can create a tool-response threshold and later open the workspace with an older JP binary. `ToolCallPolicy::Strip` already uses this tagged map shape, and its old serde representation ignores an unknown sibling field, so the older reader accepts `over` but applies `Strip` to every call in the range. The broadened projection then affects every query run by that binary without an error. Threshold-bearing stored policies need a shape or version that old readers reject rather than silently reinterpret; a compatibility test with a local copy of the pre-change tool policy would pin this down.

crates/jp_conversation/src/stream/projection.rs
```diff
@@ -237,6 +211,158 @@ pub(super) fn apply(events: &mut Vec<InternalEvent>) -> Vec<TurnOrigin> {
...
+
+        match &spec.policy {
+            ToolCallPolicy::Omit => {
+                // Removing a pair is not a per-half choice, so the threshold is
+                // judged on the two halves combined. Both halves read the same
+                // total, so a pair is never half-removed.
+                if (event.is_tool_call_request() || event.is_tool_call_response())
+                    && spec.covers(sizes.pair)
+                {
+                    return None;
```
JeanMertz: question: Is a thresholded overlay intended to cancel an earlier policy for items below its threshold? For example, suppose an existing compaction strips all tool responses in an old turn, then a later overlay says `omit` over 1 MB for that turn. `resolve_policies` selects the later spec for the whole turn before this check, so a 10 KB response fails `covers` and passes through raw instead of falling back to the earlier strip. That means adding a targeted rule can uncompact existing items. If `over` qualifies whether the policy applies to each item, resolution needs to retain earlier candidates and choose per item; if this override behavior is intentional, please document it in RFD 064 and pin it with a stacking test.

crates/jp_config/src/types/policy_spec.rs
```diff
@@ -0,0 +1,201 @@
...
+        // Either the bare policy or a table carrying it plus options. The table
+        // shape depends on `P`, so it is described only as an object.
+        schema.union(schematic::schema::UnionType {
+            variants_types: vec![
+                Box::new(schema.infer::<P>()),
+                Box::new(schema.infer::<JsonValue>()),
+            ],
+            ..Default::default()
+        })
+    }
```
JeanMertz: nit: `JsonValue` has `SchemaType::Unknown`, so this union accepts every JSON value and removes the enum validation and completion the bare policy schema provided. For example, a schema consumer accepts `reasoning = 42`, although config loading rejects it. Please describe the table as an object with required `policy` and optional `over` fields, while retaining `P` as the bare alternative.

crates/jp_config/src/conversation/compaction.rs
```diff
@@ -191,10 +194,32 @@ pub struct CompactionRuleConfig {
...
+    /// ```toml
+    /// tool_calls = { policy = "strip-responses", over = "1MB" }
+    /// ```
+    ///
+    /// With `over` set, only the large parts of a call are compacted.
+    /// Each half is judged on its own size, so `"strip"` on a call with a short
+    /// request and a huge response drops only the response.
+    /// `"omit"` removes whole pairs, so it is judged on the request and
+    /// response combined.
+    pub tool_calls: Option<PolicySpec<ToolCallsMode>>,
```
JeanMertz: nit: This text is emitted into generated `config.toml` files and the JSON Schema, but “a mode string” does not tell that reader which values are valid. List `strip`, `strip-requests`, `strip-responses`, and `omit`, and state that omitting `tool_calls` leaves tool calls unchanged for this rule.

crates/jp_cli/src/cmd/conversation/compact.rs
```diff
@@ -471,14 +509,46 @@ struct TimelineSegment {
...
+    let narrowed = compaction
+        .reasoning
+        .as_ref()
+        .is_some_and(|spec| spec.over.is_some())
+        || compaction
+            .tool_calls
+            .as_ref()
+            .is_some_and(|spec| spec.over.is_some());
+
+    narrowed.then(|| events.affected_items(compaction))
```
JeanMertz: nit: `-k 's+r,over=1KB'` is accepted, but a summary takes precedence over the mechanical policy. This still calls `affected_items`, so the timeline lists reasoning blocks as threshold-selected even though the threshold was ignored and the whole range was replaced by the summary. Return `None` here when `compaction.summary.is_some()` (or reject the combination), and add a timeline test for a summary plus a thresholded mechanical policy.

crates/jp_conversation/src/stream/projection.rs
JeanMertz: Confirmed and fixed in d0deed5d.

`TurnMut::build`'s own docs made this easy to verify — it says providers like Google reuse synthetic IDs across streaming cycles and validates responses by count for exactly that reason. Keying sizes by ID contradicted that.

Sizes are now keyed by stream position. A response pairs with the oldest request in its turn carrying the same ID that has no response yet, mirroring the count-based rule `TurnMut::build` already applies. Turn scoping matters for the orphaned-request case: without it, a later turn's response could capture an earlier turn's unmatched request.

Three regression tests, and I confirmed all three go red against the old ID-keyed lookup before fixing it. The `--tools=sres --over 1mb` case reported `["xxx…4096 chars", "ok"]` where it should report `["[compacted] fs_read_file: success", "ok"]`, and `affected_items` returned 0 items instead of 1 — exactly the outcome you described.

One bonus: the tool name on a stripped response had the same ID-keyed weakness, so a reused ID spanning two different tools would have mislabelled the status line. That's fixed by the same change.

crates/jp_config/src/types/policy_spec.rs
JeanMertz: Confirmed and fixed in d0deed5d.

Your supporting point turned out to be stronger than stated: `schematic_macros/src/common/macros.rs:161` emits `deny_unknown_fields` unless a type opts out, so every other `Partial*Config` in this tree rejects junk keys. `PolicySpec`'s hand-written `Deserialize` was the one place that didn't.

The promoted-string branch now rejects any leftover key after `over` is removed, naming it. Scoped to that branch only: the tagged-map branch legitimately hands remaining keys to `P`, and there's a test with a `#[serde(tag = "policy")]` stand-in carrying `request`/`response` siblings to make sure the rejection doesn't catch it.

Red-checked before fixing. The failure output is a good illustration of the hazard:

```
called `Result::unwrap_err()` on an `Ok` value: PartialCompactionRuleConfig {
    tool_calls: Some(PolicySpec { policy: StripResponses, over: None }), ... }
```

Config accepted, threshold gone, every response in range stripped. This was the most plausible input of the whole review — a typo, not a downgrade or a provider quirk.

crates/jp_config/src/types/policy_spec.rs
JeanMertz: The mechanism is right and I'm not disputing it: `ToolCallPolicy` has no `deny_unknown_fields`, so an older binary reads the threshold-bearing map as plain `Strip` and over-compacts the range.

I'm declining the remedy, though, because I checked what "reject" actually does here. `stream.rs:1812-1816` handles the `compaction` tag with a hard `map_err(Error::custom)`, and the `Unknown` passthrough below it only catches *unrecognized type tags* — `compaction` is recognized. So a shape an old reader rejects doesn't degrade, it fails the entire conversation load.

That trades "the projected view is more aggressively compacted until you upgrade" for "this conversation won't open until you upgrade", on a feature whose premise (RFD 064) is a non-destructive overlay where the raw events are always intact and `--reset` recovers. I think the graceful version is the better failure mode here, and it's consistent with the stream's existing forward-compat stance of degrading rather than failing.

Stating the accepted behavior plainly so it's on the record rather than a discovery: after a downgrade, a thresholded rule compacts everything in its range instead of only the large items. Nothing on disk is wrong, nothing is lost, upgrading restores correct behavior.

Happy to be overruled — the counter-argument is that silent over-compaction is invisible where a load failure is loud. If we want the signal, it belongs on the stream envelope as a version marker rather than on this one policy field, and that's its own change with its own migration story.

crates/jp_conversation/src/stream/projection.rs
JeanMertz: Your trace is correct: `resolve_policies` picks the latest spec per content type for the whole turn, so a 10 KB response under a later `omit over 1MB` fails `covers` and passes through raw rather than falling back to the earlier strip.

To answer the question: yes, intentional, but it isn't new. Latest-wins already un-compacts when a later overlay is less aggressive — `later_compaction_overrides_earlier_for_same_type` in `projection_tests.rs` carries the comment "Later: strip tool calls instead (less aggressive)". Thresholds make the granularity per-item instead of per-turn; the resolution rule is unchanged.

I'm declining the first half of the suggestion (retaining earlier candidates and choosing per item). It adds a second ordering rule interacting with the timestamp rule, and keeping resolution auditable was a deliberate call during design. Specificity precedence becomes *necessary* only once an exemption/`keep` policy exists, which is a named non-goal here — the two are coupled and should land together if they land.

Taking the second half, which you offered directly: d0deed5d adds `a_thresholded_overlay_replaces_an_earlier_policy_for_its_whole_turn`, asserting the below-threshold response stays raw while the oversized pair is omitted. So the behavior is pinned rather than incidental.

Left open deliberately: this PR doesn't touch RFD 064 at all, and the stacking section there describes turn-level resolution without mentioning thresholds. Worth deciding whether the RFD update rides along here or goes separately — I lean separately, since it's the whole feature's prose and not just this interaction.

crates/jp_cli/src/cmd/conversation/compact.rs
JeanMertz: Confirmed and fixed in d0deed5d.

It affects both paths, not just the preview: `build_compaction_for_range` layers the summary on top of `build_mechanical_compaction`, so the stored event carries the thresholded policy too and `segments_for_compactions` itemized it the same way.

Guarded in both places, since each is meaningful on its own terms:

- `threshold_items` returns `None` for a summary, so the timeline stays quiet.
- `projection::affected_items` returns empty for a summary, because the honest answer to "what does this compaction select" is "nothing, it replaces the range" regardless of who's asking.

Test asserts a single line for a summary rule carrying a thresholded `tool_calls`.

Agreed on the `nit:` label — display-only and the label already said `summary`. Worth fixing anyway: the itemization exists precisely because the range line can't be trusted alone, so it contradicting itself defeats the point.

crates/jp_config/src/conversation/compaction.rs
JeanMertz: Fixed in d0deed5d. Direct hit on this crate's own doc-comment rule — a reader in a TOML file has no intra-doc link to follow, so the values have to be spelled out.

`tool_calls` now lists all four with a gloss each:

```
- `"strip"`: replace request arguments *and* response content.
- `"strip-requests"`: replace request arguments, keep responses.
- `"strip-responses"`: replace response content, keep arguments.
- `"omit"`: remove the request and response entirely.
```

plus "If unset, tool calls in the range are left alone."

Gave `reasoning` the same treatment while I was in there — it named `"strip"` but didn't say it's the only mode or what omitting the key does.

crates/jp_config/src/types/policy_spec.rs
JeanMertz: Confirmed and fixed in d0deed5d. `json_value.rs:316` does set `SchemaType::Unknown`, so `P | Unknown` collapsed to "anything" and the union was doing no work at all.

Replaced the `JsonValue` arm with an explicit `StructType`: `policy` inferred from `P` and marked required, `over` inferred from `ByteSize` and optional by omission. The bare `P` stays as the other arm. So a consumer validates `policy` and rejects an unknown key instead of waving everything through.

One thing your comment made me notice that I'm *not* fixing here: `ToolCallsMode::build_schema` is `schema.string_default()`, so the `tool_calls` arm was only ever "string" — the enum values aren't in the schema and never were. `reasoning` gets them via `ConfigEnum`. Real completion for `tool_calls` means giving it an `EnumType` schema, which touches a type this PR otherwise leaves alone. Flagging rather than folding in; say the word if you'd rather it rode along.

PR #996, opened 2026-08-18
chore(rfd): Prune terminal RFDs from board

JeanMertz:
The priority board contains only open RFDs, so completed and retired work cannot retain a planned position. Status workflows remove affected entries, and the docs build rejects stale board ids.

`just rfd-board-prune` repairs boards after manual status edits.


docs/.vitepress/loaders/rfd-shared.mjs
```diff
@@ -200,6 +198,32 @@ export function checkPriority(entries, priority) {
...
+// no place on it. `graph` supplies the statuses; ids it doesn't know (drafts,
+// when the graph is published-only) are left to `checkPriority`.
+//
+// This is an error rather than a warning because the board file is what a human
+// reads and reorders: nothing renders a stale id, so nothing else would ever
+// point it out.
+export function checkTerminalOnBoard(graph, priority) {
+    const placed = [...priority.order, ...(priority.backlog ?? [])]
+    const terminal = [...new Set(placed)]
+        .filter(num => TERMINAL_STATUSES.has(graph.get(num)?.status))
```
JeanMertz: blocking: A contributor can plausibly mark a draft `Abandoned`; RFD 001 explicitly permits abandoning drafts. The caller passes the published-only graph here, so a terminal `DNN` entry is skipped, while `checkPriority` only confirms that the draft exists. The docs build therefore succeeds with that stale board entry, and `just rfd-board-prune` cannot repair it because it scans only `docs/rfd/NNN-*.md`. Please validate against the combined published and draft graph and make the prune recipe scan draft files too. A regression case with an abandoned draft in `.priority.json` would cover both paths.

justfile
```diff
@@ -1791,6 +1800,73 @@ _rfd-priority-rewrite OLD NEW:
...
+    ' "$priority_file" > "${priority_file}.tmp" && mv "${priority_file}.tmp" "$priority_file"
+
+# Drop terminal RFDs from the priority board.
+#
+# The board ranks work that is still open, so an Implemented, Superseded, or
+# Abandoned RFD has no place on it and the docs build rejects one that lingers.
+# `rfd-promote`, `rfd-abandon`, and `rfd-supersede` prune the id as they change
+# the status; this recipe is the repair path for a status edited by hand.
+[group('rfd')]
+rfd-board-prune:
```
JeanMertz: nit: This public `rfd` command is missing from RFD 001's Tooling table, which is the command reference for this workflow. A contributor repairing a manually edited status will not find the recipe there until the build reports it. Please add `just rfd-board-prune` to that table.

docs/.vitepress/loaders/rfd-shared.mjs
JeanMertz: Both mechanical claims check out, and I've fixed both.

- `rfds.data.js` now merges a draft graph in for this check only, following the mirror-image precedent already in `rfd-drafts.data.js` (which reads published files purely to supply statuses to `checkRequiresOnImplemented`). `checkTerminalOnBoard`'s doc comment now states that `graph` must span every id space the board can hold.
- `rfd-board-prune` iterates `docs/rfd/drafts/D[0-9][0-9]-*.md` as well. `num=${basename_f%%-*}` already yields `D24`, and `_rfd-priority-remove` accepted draft ids from the start.

Two corrections to the framing, neither of which changes the outcome:

`just rfd-abandon D12` did not work. The recipe resolved `docs/rfd/${num}-*.md` with `num` from `printf "%03d"`; that glob cannot reach `docs/rfd/drafts/`. So the input was a hand-edited `Status` line, not the recipe — RFD 001:165-167 described a path the tooling refused. Rather than file that as a follow-up I fixed it here: `rfd-abandon` now resolves through `_rfd-resolve` like `rfd-promote` and `rfd-renumber`, so it accepts `D12` and prunes the board on the way out.

The display half was already covered. `mergePriority` keys off `entry.status`, and draft entries carry a real status, so an abandoned draft was already denied a position in `rfd-list` and the web board. The gap was narrower than "the docs build therefore succeeds" implies: the file lied, and nothing reported it. That's still worth closing — the check's own justification ("nothing renders a stale id, so nothing else would ever point it out") is exactly as true for `DNN`, and shipping a gate over half the id space the board deliberately mixes is an incomplete invariant.

On the regression case: there's no JS test harness in this repo. `docs/package.json` lists `sortablejs`, `typescript`, `vitepress`, `vue`, and no justfile recipe runs JS tests. Adding one is a separate decision, not a line item on this PR. I verified by hand instead: no draft currently carries a non-`Draft` status, so the widened check doesn't newly fail the build.

justfile
JeanMertz: Not taking this one — the premise is contradicted by the two lines immediately above that table (RFD 001:478-479):

> All RFD commands are in the `rfd` group.
> Run `just --list --group rfd` to see them.

The document names `just --list --group rfd` as the command reference and the table as a selection from it. It is one: the justfile has 15 `[group('rfd')]` recipes and the table lists 9. Absent are `rfd-review`, `rfd-triage`, `rfd-implement`, `rfd-summaries`, and `rfd-manage` — that last being the recipe for editing the very board under discussion. Adding a repair command while the board's own editor stays missing would make the table more misleading, not less.

The discoverability concern also resolves itself. "Will not find the recipe there until the build reports it" — the build reporting it is the moment you need it, and the error message names the command verbatim. The docs workflow runs on every `pull_request`, so a hand-edited status is caught before merge with the fix in hand. There's no earlier moment where someone is searching for "how do I prune the board".

If we want that table to be exhaustive, that's a worthwhile change on its own covering all six omissions, not a row appended for the newest recipe.

PR #998, opened 2026-08-18
fix(config, llm)!: Preserve MCP tool schemas as declared

JeanMertz:
An `enum` on an array parameter produced a schema OpenAI rejects,
failing every query that enabled the tool. The enum belonged under
`items`, but nothing said so until the provider refused the request.

An MCP server's `inputSchema` was translated into JP's own parameter
tree, which cannot hold a `$ref`, and each provider rebuilt a schema
from it. A tool's parameters are now the document its source
declared, with configuration written into it as an overlay. Reading
follows same-document references, so a referenced enum constrains
values like an inline one, and adapting a schema to what an API
accepts is left to the provider that needs it.

That closed two silent failures: Ollama received no parameter schema
at all, and referenced types were dropped for every provider.
Schemas are validated before the request, reporting the config key
at fault instead of an HTTP 400, and a tool JP cannot describe is
skipped rather than failing the query.

BREAKING CHANGE: MCP overrides may no longer contradict the server.

Move an `enum` of individual values on an array to `items.enum`; an
enum of complete arrays keeps its JSON Schema meaning. A `type`
override must match the server's type set or be removed. Arrays must
declare `items`, local parameters must declare `type`, and a
`default` outside a narrowed `enum` is rejected. On MCP parameters
`enum = []` clears an inherited enum. Overrides that only restated a
referenced type can be deleted.


crates/jp_llm/src/tool.rs
```diff
@@ -1317,153 +1331,254 @@ async fn resolve_mcp_tool(
...
+            }),
+        };
+    }
+
+    let types = schema
+        .get("anyOf")
+        .and_then(Value::as_array)
+        .into_iter()
+        .flatten()
+        .filter_map(|variant| variant.get("type").and_then(Value::as_str))
```
JeanMertz: blocking: An MCP server can plausibly expose an optional object as `anyOf: [{"$ref": "#/$defs/Options"}, {"type": "null"}]`. This `filter_map` silently drops the `$ref` branch, resolves the parameter as null-only, and then passes validation, so every provider receives a schema that no longer describes the server's non-null input. Please preserve or resolve the branches, or reject any `anyOf` variant that cannot be represented rather than filtering it out. A regression test with `$ref` plus `null` would cover the hidden narrowing.

crates/jp_llm/src/tool/schema.rs
```diff
@@ -0,0 +1,298 @@
...
+            return Err(ToolError::InvalidSchema {
+                path: format!("{path}.default"),
+                message: format!(
+                    "default value {default} has type {}, but the schema requires {}",
+                    value_type(default),
+                    format_types(&config.kind),
+                ),
+            });
+        }
+        validate_nested_schema_value(&format!("{path}.default"), default, config, "default value")?;
```
JeanMertz: blocking: A local parameter with `type = "string"`, `enum = ["json"]`, and `default = "yaml"` passes this validation because the default is checked against the type and nested shape, but never against `config.enumeration`. If the model omits the argument, `apply_parameter_defaults` inserts `"yaml"` and the local command receives a value its schema forbids. Please validate the current node's enum before recursing, ideally through one value-against-schema helper that also applies to nested properties, and add this case as a regression test.

crates/jp_llm/src/tool.rs
```diff
@@ -1317,153 +1331,254 @@ async fn resolve_mcp_tool(
...
+fn merge_mcp_schema_node(
+    path: &str,
+    source: &Value,
+    override_config: Option<&ToolParameterConfig>,
+    required_in_mcp: bool,
+) -> Result<ToolParameterSchema, ToolError> {
+    let source_kind = parse_schema_kind(path, source)?;
+    let override_kind = override_config.and_then(|config| config.kind.as_ref());
+    let kind = match (source_kind, override_kind) {
+        (Some(source), Some(config)) if &source != config => {
```
JeanMertz: nit: A user can plausibly repeat an MCP union as `type = ["null", "string"]` while the server emits `["string", "null"]`. Direct enum equality rejects that override even though JSON Schema type-array order has no meaning, which conflicts with the documented promise that matching inherited types are accepted. Compare normalized type sets here, and treat `One("string")` as equivalent to `Many(["string"])`; duplicate detection can remain in schema validation.

crates/jp_llm/src/tool.rs
```diff
@@ -1181,19 +1170,39 @@ async fn resolve_tool(
...
+    let definition = match config.source() {
         ToolSource::Local { .. } | ToolSource::Builtin { .. } => {
             let docs = ToolDocs::from_config(config);
-            Ok(ToolDefinition {
+            let parameters = config
+                .parameters()
+                .iter()
+                .map(|(parameter_name, parameter)| {
+                    resolve_config_parameter(
+                        &format!("tools.{name}.parameters.{parameter_name}"),
```
JeanMertz: nit: For a user with an invalid local schema, this path starts at `tools.<name>`, but the public config key is `conversation.tools.<name>`. The enum error also presents this path in a prescriptive “use ...items.enum” hint, so it cannot be copied into the TOML file or a root-level `--cfg` assignment. Please use the full config path, or keep the diagnostic path abstract and phrase the fix as setting `items.enum` under the affected parameter.

JeanMertz: Accepted, fixed.

Confirmed against the crate's own existing diagnostic: `reject_access_on_non_local_tools` already emits `conversation.tools.{name}: ...` (`jp_config/src/conversation/tool.rs:175`), so the truncated prefix was inconsistent as well as uncopyable.

All three production sites now use the full config key (`tool.rs:1181`, `:1200`, `:1255`), so errors read:

```text
Invalid schema at `conversation.tools.bear_note_create.parameters.tags.enum`:
enum value "projects/jp" has type string, but the schema requires array; use
`conversation.tools.bear_note_create.parameters.tags.items.enum` to constrain
array elements
```

That makes the hint work verbatim as both a TOML table header and a `--cfg` assignment, so I kept the prescriptive phrasing rather than making the path abstract. Test expectations and the error sample in `docs/features/tools.md` updated, plus a line in the docs stating the path is a full configuration key.

crates/jp_llm/src/tool.rs
JeanMertz: Accepted, fixed.

Worth recording how bad the second case was: `OneOrManyTypes` derives `PartialEq`, so server `{"type":"string"}` against config `type = ["string"]` produced

```text
Invalid schema at `parameters.name.type`: MCP declares string, but the configuration declares string
```

because `format_types` renders `One("string")` and `Many(["string"])` identically. The reordered-union case produced `MCP declares string or null, but the configuration declares null or string`. Both are reproduced by the new red tests.

`merge_mcp_schema_node` now compares normalized type sets via `types_match` (sorted, deduplicated), keeping the source's declaration as the resolved `kind` so provider output is unchanged.

One correction to "duplicate detection can remain in schema validation": it would not have. Validation runs on the resolved schema, which takes the server's type list, so a duplicate or misspelling in the override would have been silently dropped once set comparison accepted it. `merge_mcp_schema_node` therefore also runs `validate_types` on the override's own `kind` before comparing.

New tests: `override_may_reorder_an_inherited_union`, `override_may_restate_a_scalar_type_as_a_single_element_array`, `duplicate_type_in_an_override_is_rejected`. `user_cannot_change_mcp_parameter_type` still passes. `docs/features/tools.md` now states that matching is by type set, order-insensitive.

crates/jp_llm/src/tool.rs
JeanMertz: The narrowing is real and I reproduced it: `{"anyOf":[{"$ref":"#/$defs/Options"},{"type":"null"}]}` resolves to `{"type":["null"]}`, passes validation, and reaches the provider. schemars emits exactly that shape for `Option<Struct>`, so it is a plausible input.

Two things change the shape of the fix, though.

First, it is not introduced here. The pre-PR code ran the identical `filter_map` over `anyOf` variants in its fallback branch and produced the same `Many(["null"])`. This diff preserves the behavior.

Second, rejecting is a net loss. `resolve_tool` errors propagate through `tool_definitions` (`tool.rs:1160`) to `jp_cli/src/cmd/query.rs:604` with `?`, so one MCP tool with an unrepresentable `anyOf` would fail *every* query that enables it, rather than degrading one tool. Trading a silently narrow tool for a dead session is not obviously the better failure mode.

So I took the "hidden" half now and left the fidelity half for a follow-up:

- `parse_schema_kind` logs a `warn!` naming the path and the resolved type set whenever a variant is dropped, so the narrowing is visible in the trace log instead of silent.
- Added `optional_reference_variant_narrows_to_null`, pinning the `$ref` plus `null` result so any future change to this path shows up as a test update. It is a characterization test: the behavior predates this PR, so it passes today. The warning itself is not asserted, since the crate has no log-capture harness.

Deferred, and worth filing: resolve `$ref` against the tool's root `inputSchema` `$defs` so the non-null branch survives. That needs the root schema threaded through `merge_mcp_schema_node` and a decision on the failure mode when a reference cannot be resolved. That same follow-up should revisit whether a per-tool schema error skips the tool with a warning instead of aborting the query, which would also soften the items-position rejection this PR introduces (`unresolved_mcp_item_reference_is_rejected`). I do not have issue-creation access from here, so this needs filing by hand.

crates/jp_llm/src/tool/schema.rs
JeanMertz: Accepted, fixed. Traced the whole chain: `apply_parameter_defaults` inserts the default verbatim (`tool.rs:1043`) and `validate_tool_arguments` only checks presence and unknown keys (`tool.rs:1074-1089`), so nothing downstream catches it.

The trigger is more plausible than a hand-written contradiction, which is what makes it worth the fix: narrowing an inherited MCP enum while inheriting the server's `default`. For a server parameter `{"type":"string","enum":["all","open","closed"],"default":"all"}` plus `enum = ["open","closed"]`, JP was building `{"type":"string","enum":["open","closed"],"default":"all"}` and shipping it, which is exactly the provider-rejected shape this PR exists to prevent.

Took the single-helper suggestion. `validate_value_against_schema` now applies type, then the node's `enum`, then recurses into array elements and object properties, and it backs both `default` and the nested contents of enum values. That closed a second gap in passing: the old `validate_nested_schema_value` enforced `items.enumeration` but never `property.enumeration`, so an object default violating a property enum also slipped through.

New red-first tests: `default_outside_a_narrowed_enum_is_rejected` and `default_outside_a_property_enum_is_rejected`. Existing `default_values_must_match_nested_item_schema` and `complete_array_enum_values_must_match_item_schema` still pass unchanged, so paths and messages are preserved; the only message change is `is not allowed by the item enum` becoming `is not allowed by the enum`, since the path already carries the index. `docs/features/tools.md` lists the case under validation and under compatibility notes.

crates/jp_llm/src/tool.rs
JeanMertz: Update: this is fixed in the PR rather than deferred. Correcting my earlier reply.

Two things changed my mind on scope. First, the escalation cost I used to argue against rejecting was avoidable: `tool_definitions` now drops a tool whose schema will not resolve and logs the reason, matching the existing "backing server is not running" skip a few lines above, so a bad tool no longer kills the query. A tool named by `--tool` still errors, since silently omitting an explicitly requested tool is worse than failing. Second, once the query no longer dies, resolving the reference properly is the obvious move.

What landed:

- `inline_schema_ref` follows same-document pointers (`#/$defs/...`) against the tool's root `inputSchema`, which `resolve_mcp_tool` already had in hand. Sibling keys override the referenced definition, per JSON Schema 2020-12.
- `flatten_nullable_union` folds `anyOf: [T, {"type":"null"}]` into a single node. Collecting only the variants' type names, which is all the old code did, would still have dropped `T`'s `properties`, `items` and `enum`, so `Option<Struct>` would have gone from null-only to a typeless object. It now resolves to `{"type":["object","null"]}` carrying the referenced properties.
- `MAX_SCHEMA_DEPTH = 8` bounds expansion. Inlining makes a self-referential type expand forever, so this is new exposure that the guard closes; `$ref` cycles are also tracked per node.
- Remote pointers are not followed, by decision. They leave the node without a type, which configuration can still fill in.

Concrete win beyond the reported bug: `bookworm`'s `crate_search_items.kinds` is `Vec<EntryType>`, whose 12 allowed values live in `$defs.EntryType`. `.jp/mcp/tools/rust/crate_search_items.toml` currently compensates with a hand-written `items.type` plus the value list spelled out in English in the summary. That enum now flows through automatically, with no config change, and the override can be dropped once someone verifies it against a running server.

Tests: `optional_reference_variant_resolves_to_the_referenced_shape` replaces the characterization test I added earlier, and inverting that assertion is the red/green pair for this change. Plus `referenced_item_type_resolves_from_defs`, `sibling_keys_override_the_referenced_definition`, `self_referential_reference_stops_at_the_depth_cap`, `remote_reference_is_not_resolved`, `tool_with_an_unresolvable_schema_is_skipped`, and `forced_tool_with_an_unresolvable_schema_still_errors`.

One caveat worth naming: the skip warning goes to the trace log, which the terminal does not show below `-v` and which is deleted on a successful run. A tool can now disappear from a query without the user noticing. `docs/features/tools.md` says so explicitly, but if that turns out to bite, surfacing skipped tools on the status line is the follow-up.

PR #1002, opened 2026-08-19
feat(xct2cli): Vendor a reader for Instruments trace bundles

JeanMertz:
Profiling the macOS app needs symbolicated stacks, which `xctrace export`
does not produce. This crate reconstructs them; `README.md` covers how,
and what a recorded bundle discloses.

Vendored because we need a different crate from the one upstream
publishes — not because upstream is unmaintained:

- **Swift demangling**, which upstream has none of. Without it every
  symbol in our own app reads `$s2JP17Conversation…`, which is the whole
  reason for having this.
- **Environment redaction.** A bundle embeds the recorded process's full
  environment, including every key the launching shell exported.
- **quick-xml 0.41**, against upstream's 0.39.
- **Four subsystems removed** — binary, terminal renderer, disassembly
  annotator, hardware counters — dropping `capstone`,
  `annotate-snippets` and `owo-colors`.

MIT OR Apache-2.0, from `9ebb2e0`, both licence files kept alongside.
Source stays close to upstream so it can be re-synced: style lints it
does not satisfy are allowed by name rather than fixed.


crates/contrib/xct2cli/src/xctrace.rs
```diff
@@ -0,0 +1,127 @@
...
+        &self,
+        template: &str,
+        output_trace: &Utf8Path,
+        target: &Path,
+        target_args: &[OsString],
+        extra_env: &[(String, String)],
+    ) -> Result<()> {
+        let mut args: Vec<OsString> = vec![
+            "record".into(),
+            "--template".into(),
```
JeanMertz: blocking: On Xcode 26, `record_launch("Time Profiler", ...)` takes the `--template` path that this PR's README says creates a bundle which later fails export with `Document Missing Template Error`. This method returns success after persisting that unusable bundle, so the caller only finds out during a later read. Use `--instrument` for this recording path, or make template versus instrument an explicit input, and add an argument-construction regression test.

crates/contrib/xct2cli/src/xctrace.rs
```diff
@@ -0,0 +1,127 @@
...
+        }
+        self.run_args(sub, &args)
+    }
+
+    fn run_args(&self, sub: &'static str, args: &[OsString]) -> Result<Vec<u8>> {
+        let mut cmd = Command::new(&self.binary);
+        cmd.args(args);
+        cmd.stdout(Stdio::piped());
+        cmd.stderr(Stdio::piped());
+        tracing::debug!(?args, binary = ?self.binary, "spawning xctrace");
```
JeanMertz: blocking: A caller that records with `extra_env = [("ANTHROPIC_API_KEY", "sk-...")]` puts the full `KEY=VALUE` in `args`, and this debug event serializes it verbatim. Enabling debug tracing silently copies the credential into logs, which defeats the redaction guarantee added by this PR. Log a sanitized argument list, or omit values following `--env`, and cover that path with a capture test.

crates/contrib/xct2cli/src/trace/toc.rs
```diff
@@ -0,0 +1,336 @@
...
+    loop {
+        buf.clear();
+        match reader.read_event_into(&mut buf)? {
+            Event::Empty(s) | Event::Start(s) if local_name(&s)? == "process" => {
+                let attrs = collect_attrs(&s)?;
+                let name = attrs.get("name").cloned().unwrap_or_default();
+                let pid = attrs.get("pid").and_then(|v| v.parse().ok()).unwrap_or(-1);
+                let path = attrs.get("path").cloned();
+                out.push(TocProcess { name, pid, path });
+                if matches!(reader.read_event_into(&mut Vec::new())?, Event::End(_)) {}
```
JeanMertz: blocking: A valid compact TOC such as `<processes><process .../><process .../></processes>` loses the second process here because the `Event::Empty` arm unconditionally consumes one extra event. With one process it consumes `</processes>` and eventually reports EOF; with adjacent processes it can return a plausible but incomplete process list. Split `Event::Empty` from `Event::Start`, and only consume or skip the matching end tag for the latter. A compact version of `sample-toc.xml` would pin this.

crates/contrib/xct2cli/src/xctrace.rs
```diff
@@ -0,0 +1,127 @@
...
+        target_args: &[OsString],
+        extra_env: &[(String, String)],
+    ) -> Result<()> {
+        let mut args: Vec<OsString> = vec![
+            "record".into(),
+            "--template".into(),
+            template.into(),
+            "--output".into(),
+            output_trace.as_os_str().into(),
+            "--no-prompt".into(),
```
JeanMertz: nit: A first recording with a fresh CPU Counters or custom template can require one-time interactive setup. Hardcoding `--no-prompt` makes that ordinary input fail, and this API gives the caller no way to allow the prompt. Make prompt suppression an explicit option and inherit stdio when prompting is allowed, as otherwise the prompt would be captured invisibly.

crates/contrib/xct2cli/src/analysis/callgraph.rs
```diff
@@ -0,0 +1,260 @@
...
+            return Ok(None);
+        }
+        let slide = match &self.slide {
+            SlideMode::Manual(s) => *s,
+            SlideMode::Auto => self
+                .binary
+                .as_deref()
+                .and_then(|bin| BinaryInfo::open(bin).ok())
+                .and_then(|info| {
+                    let loads = self.bundle.image_loads().unwrap_or_default();
```
JeanMertz: nit: An attach-mode trace has no recorded dyld loads, as the README points out, and an `image_loads` export can also fail. Both paths silently become `Slide::ZERO` here, so a caller that supplied a binary and dSYM gets a report of raw addresses with no diagnostic. This is visible in the report, but hard to diagnose. Preserve `BinaryInfo::open` errors and warn or return a typed error when automatic slide recovery produces no slide, consistent with `HotspotsBuilder`.

crates/contrib/xct2cli/src/analysis/hotspots.rs
```diff
@@ -0,0 +1,401 @@
...
+                            Some(s) => {
+                                tracing::info!(
+                                    %s,
+                                    "auto-detected slide from kdebug DBG_DYLD events"
+                                );
+                            }
+                            None => {
+                                tracing::warn!(
+                                    "could not auto-detect slide; symbols will be wrong. Use \
+                                     `xct2cli slide` to inspect candidates and pass --slide \
```
JeanMertz: nit: A trace with no dyld load events reaches this warning, but this vendored crate deliberately removes the `xct2cli` binary and its `slide` command. The suggested recovery is therefore unavailable to callers. Point them to `BinaryInfo::enumerate_slides` plus `SlideMode::Manual`, or return enough structured information for the caller to present its own recovery instructions.

crates/contrib/xct2cli/src/redact.rs
```diff
@@ -0,0 +1,87 @@
...
+/// Self-closing `<environment/>` elements are left as they are; they hold
+/// nothing.
+/// Elements whose name merely starts with `environment`, such as
+/// `<environment-info>`, are not touched.
+///
+/// Input that opens a block and never closes it is truncated at that point
+/// rather than passed through, so malformed XML cannot become a disclosure.
+#[must_use]
+pub fn strip_environment(xml: Vec<u8>) -> Vec<u8> {
+    let mut out = Vec::with_capacity(xml.len());
```
JeanMertz: nit: A long Time Profiler bundle can produce a very large kdebug XPath export, and that XML has no `<environment>` element. `Command::output` already holds the full export, then this allocation and the final copy temporarily double its memory footprint before parsing. That can make an otherwise readable trace fail from memory pressure. Return the original `Vec` when `find_element(&xml)` is `None`, or redact in place/while streaming.

PR #1003, opened 2026-08-19
chore(tools): Add a `cargo_install_tools` tool

JeanMertz:
Changing anything under \`.config/jp/tools\` had no effect until somebody
ran \`just install-tools\` in a terminal: the binary serving a tool call is
the one installed earlier, so an edit made during a conversation stayed
dormant and the next call still ran the old code. Nothing in the tool
surface could close that loop, which made editing the tools the one kind
of work that could not be finished from inside a conversation.

\`cargo_install_tools\` rebuilds and reinstalls the binary. It shells out
to \`just install-tools\` rather than calling \`cargo install\` directly, so
the install flags stay defined in one place and cannot drift from a
hand-run install.

\`cargo install\` renames the new binary over the old one while the current
process keeps running from the file it already opened, so the rebuilt
code answers the *next* call rather than this one. The success message
says so outright, because the alternative is a caller concluding their
change is live and then reading a stale tool's output as evidence about
new code.

The tool definition ships disabled and the \`rust-development\` skill
turns it on, so it reaches any persona extending that skill.

### DIFF
--- a/.config/jp/tools/src/cargo/install_tools.rs
+++ b/.config/jp/tools/src/cargo/install_tools.rs
@@ -1,0 +1,55 @@
+use camino::Utf8Path;
+
+use super::MAX_DIAGNOSTIC_BYTES;
+use crate::util::{
+    ToolResult, error,
+    runner::{DuctProcessRunner, ProcessOutput, ProcessRunner},
+    truncate,
+};
+
+pub(crate) async fn cargo_install_tools(root: &Utf8Path) -> ToolResult {
+    cargo_install_tools_impl(root, &DuctProcessRunner)
+}
+
+fn cargo_install_tools_impl<R: ProcessRunner>(root: &Utf8Path, runner: &R) -> ToolResult {
+    // Going through `just` keeps the install flags defined in one place, so this
+    // and a hand-run `just install-tools` cannot drift apart.
+    let ProcessOutput {
+        stdout,
+        stderr,
+        status,
+    } = runner.run("just", &["install-tools"], root)?;
+
+    if !status.is_success() {
+        let diagnostics = strip(&stderr);
+        return error(format!(
+            "Rebuilding the tools binary failed:\n\n```\n{}\n```",
+            if diagnostics.is_empty() {
+                strip(&stdout)
+            } else {
+                diagnostics
+            }
+        ));
+    }
+
+    // `cargo install` renames the new binary over the old one, so the process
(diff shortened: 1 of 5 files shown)


.config/jp/tools/src/cargo/install_tools.rs
```diff
@@ -0,0 +1,55 @@
...
+pub(crate) async fn cargo_install_tools(root: &Utf8Path) -> ToolResult {
+    cargo_install_tools_impl(root, &DuctProcessRunner)
+}
+
+fn cargo_install_tools_impl<R: ProcessRunner>(root: &Utf8Path, runner: &R) -> ToolResult {
+    // Going through `just` keeps the install flags defined in one place, so this
+    // and a hand-run `just install-tools` cannot drift apart.
+    let ProcessOutput {
+        stdout,
+        stderr,
```
JeanMertz: blocking: A JP developer using sibling worktrees with a shared `CARGO_TARGET_DIR` can get a successful install of a stale artifact from the other worktree. This repo already enables `CARGO_UNSTABLE_CHECKSUM_FRESHNESS` for `cargo_check`, `cargo_test`, and `cargo_expand` because ordinary mtime freshness causes exactly that failure, but this invocation runs the install recipe without it and then reports that the edited code is active. The stale install is hidden behind a success result. Please make `just install-tools` opt into checksum freshness so the hand-run and tool paths stay aligned, and test that the environment reaches `just`.

.config/jp/tools/src/cargo/install_tools.rs
```diff
@@ -0,0 +1,55 @@
...
+        return error(format!(
+            "Rebuilding the tools binary failed:\n\n```\n{}\n```",
+            if diagnostics.is_empty() {
+                strip(&stdout)
+            } else {
+                diagnostics
+            }
+        ));
+    }
+
```
JeanMertz: blocking: With `CARGO_INSTALL_ROOT=/tmp/jp-install` while `PATH` still resolves `~/.cargo/bin/jp-tools`, `cargo install` exits successfully after writing `/tmp/jp-install/bin/jp-tools`, but the next `just serve-tools` call still starts the stale binary from `PATH`. Cargo emits a successful-install warning when its destination is absent from `PATH`, but this branch discards all successful stderr and unconditionally claims the new code will answer the next call. Please verify that the installed executable is the one `serve-tools` resolves, or make the recipe install explicitly over the executable serving this call before returning this success message.

PR #1004, opened 2026-08-19
refactor(workspace, cli): Move workspace opening into `jp_workspace`

JeanMertz:
`jp_cli::load_workspace` assembled a workspace from disk by hand: find
the root, load the ID, build the filesystem backend, wire the user-local
silo, then store the ID back. That is workspace knowledge living in the
CLI, and a second consumer has to repeat it. Repeating it wrong is
silent: wiring the filesystem backend without user-local storage still
compiles and still returns conversations, just fewer of them.

`Workspace::open` owns that sequence now, with `DEFAULT_STORAGE_DIR`
moved alongside it. A new `Error::WorkspaceNotFound` carries the
directory that was searched, so the CLI keeps printing its `jp init`
hint while other callers receive a typed error rather than a formatted
string.

The two in-memory constructors become `in_memory` and
`in_memory_with_id`, so each constructor says which side of the disk it
sits on. Two accessors come with them: `fs_storage` returns the
filesystem backend an opened workspace holds, and `sessions` returns the
session backend, which the `--no-persist` path wraps in
`ReadOnlySessionBackend` instead of reaching for the filesystem backend
that happened to be serving that role.

Behavior is unchanged. A test covers the wiring that is otherwise
invisible when wrong: a conversation created with `--local` lives only
in the user-local silo, and a workspace opened from disk must list it.

### DIFF
--- a/crates/jp_workspace/src/lib.rs
+++ b/crates/jp_workspace/src/lib.rs
@@ -28,8 +28,8 @@ use jp_config::AppConfig;
 use jp_conversation::{Conversation, ConversationId, ConversationStream};
 use jp_storage::{
     backend::{
-        ConversationFilter, ConversationIndexEntry, InMemoryStorageBackend, LoadBackend,
-        LockBackend, NullPersistBackend, PersistBackend, Projection, SessionBackend,
+        ConversationFilter, ConversationIndexEntry, FsStorageBackend, InMemoryStorageBackend,
+        LoadBackend, LockBackend, NullPersistBackend, PersistBackend, Projection, SessionBackend,
         StoragePresence,
     },
     lock::LockInfo,
@@ -42,6 +42,10 @@ use tracing::{debug, trace, warn};
 use crate::session::Session;
 
 const APPLICATION: &str = "jp";
+
+/// The directory a workspace stores its data in, relative to the workspace
+/// root.
+pub const DEFAULT_STORAGE_DIR: &str = ".jp";
 
 #[derive(Debug)]
 pub struct Workspace {
@@ -62,6 +66,9 @@ pub struct Workspace {
 
     /// Backend for session-to-conversation mapping storage.
     sessions: Arc<dyn SessionBackend>,
+
+    /// The filesystem backend, for workspaces opened from disk.
+    fs: Option<Arc<FsStorageBackend>>,
 
     /// The in-memory state of the workspace.
     state: State,
@@ -87,23 +94,25 @@ impl Workspace {
         }
     }
 
-    /// Creates a new workspace with the given root directory.
+    /// Creates a workspace with the given root directory, backed by memory.
     ///
-    /// The workspace starts with in-memory backends (no filesystem
-    /// persistence).
-    /// Call [`with_backend`] to wire in a storage backend.
+    /// Nothing is read from or written to disk.
+    /// Call [`with_backend`] to wire in a storage backend, or [`open`] to open
+    /// a workspace that already exists on disk.
     ///
+    /// [`open`]: Self::open
     /// [`with_backend`]: Self::with_backend
-    pub fn new(root: impl Into<Utf8PathBuf>) -> Self {
-        Self::new_with_id(root, id::Id::new())
+    pub fn in_memory(root: impl Into<Utf8PathBuf>) -> Self {
+        Self::in_memory_with_id(root, id::Id::new())
     }
 
-    /// Creates a new workspace with the given root directory and ID.
+    /// Creates a workspace with the given root directory and ID, backed by
+    /// memory.
     ///
     /// All four backend slots are wired to a single shared
     /// [`InMemoryStorageBackend`], so data written through one trait is visible
     /// through the others.
-    pub fn new_with_id(root: impl Into<Utf8PathBuf>, id: id::Id) -> Self {
+    pub fn in_memory_with_id(root: impl Into<Utf8PathBuf>, id: id::Id) -> Self {
         let root = root.into();
         trace!(root = %root, id = %id, "Initializing Workspace.");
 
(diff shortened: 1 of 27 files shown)


crates/jp_workspace/src/lib.rs
```diff
@@ -115,10 +124,71 @@ impl Workspace {
...
+        trace!(root = %root, "Found workspace root.");
+
+        let storage = root.join(storage_dir);
+        trace!(storage = %storage, "Initializing workspace storage.");
+
+        let id = Id::load(&storage)
+            .transpose()
+            .ok()
+            .flatten()
+            .unwrap_or_default();
```
JeanMertz: blocking: A plausible input is a `.jp/.id` file that was truncated or contains merge-conflict text. This converts the parse error into a fresh ID, `with_user_storage` selects a new silo for that ID, and line 187 persists the replacement. Any `--local` conversations in the original silo then disappear from listings without a diagnostic. Please treat only a genuinely missing ID as the new-workspace case and propagate malformed or unreadable IDs before touching user storage. A regression test with an invalid `.id` should verify that `open` returns `Error::Id` and leaves the file unchanged.

crates/jp_workspace/src/lib.rs
```diff
@@ -115,10 +124,71 @@ impl Workspace {
...
+            slug,
+            id.to_string(),
+        )?);
+
+        let mut workspace = Self::in_memory_with_id(root, id).with_backend(fs.clone());
+        workspace.fs = Some(fs);
+
+        workspace.id().store(&storage)?;
+
+        Ok(workspace)
```
JeanMertz: blocking: A workspace containing existing conversations still returns an empty `conversations()` iterator immediately after the documented `Workspace::open` call, because `state` remains the default until a caller separately remembers `load_conversation_index()`. That is a hidden data-visibility failure for the second consumer this API is meant to support. Please make the opened value usable before returning, or encode the unloaded state so conversation APIs cannot silently return empty data. If sanitization must precede indexing, a single preparation operation can own both steps.

crates/jp_workspace/src/lib.rs
```diff
@@ -153,6 +223,31 @@ impl Workspace {
...
+    #[must_use]
+    pub fn fs_storage(&self) -> Option<&Arc<FsStorageBackend>> {
+        self.fs.as_ref()
+    }
+
+    /// The backend session-to-conversation mappings are read from and written
+    /// to.
+    ///
+    /// Set by [`with_sessions`] or [`with_backend`]; an in-memory workspace
+    /// starts with one that discards writes.
```
JeanMertz: nit: This describes the opposite behavior. `in_memory_with_id` installs an `InMemoryStorageBackend`, whose `save_session` stores mappings in its map; it does not discard writes. Please say that the in-memory workspace keeps session mappings for the lifetime of the backend.

crates/jp_workspace/src/lib.rs
```diff
@@ -115,10 +124,71 @@ impl Workspace {
...
+        Self::open_with_storage_dir(dir, DEFAULT_STORAGE_DIR)
+    }
+
+    /// Open the workspace containing `dir`, looking for a store named
+    /// `storage_dir`.
+    ///
+    /// Behaves exactly like [`open`], which uses [`DEFAULT_STORAGE_DIR`].
+    ///
+    /// [`open`]: Self::open
+    pub fn open_with_storage_dir(dir: &Utf8Path, storage_dir: &str) -> Result<Self> {
```
JeanMertz: theoretical: I cannot name a CLI input that needs a custom workspace store name. The production path added here calls `open`, while the direct parameterized call in this diff exists to keep the missing-workspace test independent of parent directories. Making this public adds a second workspace-opening contract and allows callers to diverge from `DEFAULT_STORAGE_DIR`. Keep the helper private, or test `find_root` directly, until a real consumer needs custom store names.

JeanMertz: Agreed, and done — dropped the `pub`.

Confirmed there are exactly two callers: `open` at `lib.rs:149` and the missing-workspace test at `lib_tests.rs:1023`. Nothing in `jp_cli`, `jp_ffi` or the plugins reaches for it, so the `pub` was buying only the second workspace-opening contract you describe.

I took the first of your two options rather than the second. Testing `find_root` directly would drop coverage of `Error::WorkspaceNotFound`, which is new in this PR and is what the CLI matches on to print its `jp init` hint. The parameterized helper still earns its keep for the test — `.jp-no-such-store` is what keeps the assertion independent of whether the temp directory happens to sit beneath a real workspace, which `open` with its hardcoded `.jp` can't do. Making it private keeps that and removes the public surface: `lib_tests.rs` is included as `mod tests` at `lib.rs:961`, a child of the crate root, so it still sees the function.

crates/jp_workspace/src/lib.rs
JeanMertz: The mechanism is exactly as you describe, and I traced each link to be sure:

- `Id::load` returns `Option<Result<Id>>` (`id.rs:41`) — `None` for an unreadable file, `Some(Err(_))` when the last line fails `from_str`, which rejects anything that isn't five characters of `[0-9a-z]` (`id.rs:96-106`). Conflict markers fail both checks.
- `.transpose().ok().flatten()` discards the `Err` arm and `.unwrap_or_default()` mints a new random ID.
- `with_user_storage(..., id.to_string())` keys the silo by that ID, so a fresh ID selects a different silo.
- `id().store(&storage)` at line 187 then overwrites `.jp/.id` with the replacement.

Spreads and hides, and the input is plausible — `.jp/.id` is tracked in this repository, so a conflicted merge in it is ordinary.

I'd still rather not fix it in this PR. That five-line chain is not new: it's removed verbatim from `jp_cli::load_workspace` and pasted into `jp_workspace` in this same diff, and "behaviour is unchanged" is the claim a 25-file mechanical rename is reviewed against. Adding a behaviour change plus a new failure-path test costs that, and the fix deserves to be the subject of its own change rather than a rider on a relocation.

Tracked separately. The fix is to stop flattening:

```rust
let id = match Id::load(&storage) {
    None => Id::new(),
    Some(result) => result?,
};
```

`Error::Id` already exists (`id.rs:97`), so `open` needs no new variant and the CLI's `From<jp_workspace::Error>` already covers it — along with the regression test you describe, asserting `Error::Id` and that the file is left alone.

What I have changed here is the doc comment, which said "a store with no readable ID file is assigned a fresh ID" and read as though it covered only the missing-file case. It now says missing, unreadable, or malformed, so the hazard is at least written down while the fix is pending.

crates/jp_workspace/src/lib.rs
JeanMertz: The observation is right — `conversations()` is empty until `load_conversation_index()` runs — but the primary fix would introduce a bug, so I've taken the third option you offered rather than the first.

Loading the index inside `open` would break a documented ordering. `sanitize.rs:15` says "Call this before [`load_conversation_index`] to guarantee the backing store is consistent", and `jp_cli` follows it: the trashed-conversation sweep runs at `lib.rs:480-493`, then `load_conversation_index()` at line 500. Indexing inside `open` would index a store that hasn't been repaired yet.

Two corrections to the premise, offered in the spirit of getting the record straight rather than deflecting. This isn't new — the pre-change API had the identical requirement, and every construction site in the tree already pairs itself with `load_conversation_index()`. And it hasn't landed on the second consumer: `jp_ffi::jp_workspace_open` does call it, and says so in its doc.

So for this PR, `open`'s doc comment now states the contract — that the index is not populated, that `load_conversation_index` is the caller's next step, and that `sanitize` comes before it.

Your closing suggestion is the right shape for the real fix, and I've tracked it separately: a single preparation operation owning sanitize-then-index, or the typestate encoding, so `conversations()` cannot be reached on an unindexed workspace. Both touch every construction site in the workspace crate and the CLI, which is why they aren't riding along here.

crates/jp_workspace/src/lib.rs
JeanMertz: Correct, and it's my error — fixed.

`in_memory_with_id` wires all four slots to one `InMemoryStorageBackend` (`lib.rs:120-127`), whose `save_session` inserts into `sessions: Arc<Mutex<HashMap<String, Value>>>` (`memory.rs:216-222`). Writes are kept, not dropped.

The wording almost certainly leaked from `ReadOnlySessionBackend`, which is the type that does discard writes — `null.rs:108-110` returns `Ok(())` without touching the inner backend — and which appears a few lines away in this same diff, in the `--no-persist` path.

Reworded to say the in-memory workspace keeps its mappings for the lifetime of the backend.

crates/jp_workspace/src/lib.rs
JeanMertz: Filed as `T-055mxpv`, carrying the mechanism, the fix above, and the regression test you asked for.

crates/jp_workspace/src/lib.rs
JeanMertz: Filed as `T-055n0nw`, with both shapes written up — a preparation operation owning sanitize-then-index, and the typestate encoding — and why the sanitize ordering rules out doing it inside `open`.

PR #1005, opened 2026-08-19
feat(conversation): Expose events in a form readers outside Rust use

JeanMertz:
A reader outside this workspace — the FFI boundary, a plugin, a web
view — needs two things the crate did not offer. It needs timestamps in
a format its platform can parse, and it needs to know what an entry in
the stream is without decoding the storage encoding itself.

`rfc3339` formats a timestamp the one way every platform's date parser
accepts. Storage keeps timestamps in `time`'s human-readable spelling
(`2024-09-01 10:00:00.0`), which nothing outside Rust reads, and each
reader otherwise reinvents the conversion. Sub-second precision is kept
when the value has any. `rfc3339_str` does the same for a caller holding
raw JSON rather than a typed event, returning `None` for a value that
parses as neither spelling so the caller can leave it as it found it.

`EventKind::type_tag` returns the tag serde writes, which is what a
reader switches on. `as_str` returns the Rust name and is for messages
addressed to somebody reading this code; the two were the same string by
coincidence and nothing held them that way. Two tests do now: one checks
every variant's tag against what serde actually serializes, the other
against `TYPE_TAGS`, the list the deserializer uses to decide whether an
entry is a known event. A tag missing from that list makes its variant
unreachable, and the stream keeps every such event as raw JSON instead.

`StreamEntry` and `ConversationStream::iter_entries` give a borrowed view
of every entry in order, including the config deltas, compaction overlays
and unrecognized entries that `iter` and `iter_events_by_turn` skip.
Those are part of what happened, and a reader presenting the stream to
somebody wants them. Nothing is copied or re-encoded on the way out.

### DIFF
--- a/crates/jp_conversation/src/event_tests.rs
+++ b/crates/jp_conversation/src/event_tests.rs
@@ -5,8 +5,29 @@ use super::{
     InquiryResponse, InquirySource, ToolCallRequest, ToolCallResponse, TurnStart,
 };
 
-/// One value of every variant, so a new variant fails to compile here rather
-/// than going unnoticed.
+/// Fails to compile when a variant is added to [`EventKind`].
+///
+/// That is the reminder to extend [`every_kind`] below and
+/// [`EventKind::TYPE_TAGS`].
+/// Neither is checked by anything else, and the two tests here agree with each
+/// other when a variant is missing from both: the tags they compare are derived
+/// from `every_kind`, so a variant absent there is absent from both sides.
+/// A variant missing from `TYPE_TAGS` is unreachable on load, and the stream
+/// keeps every one of its events as raw JSON instead.
+#[expect(dead_code)]
+fn every_variant_is_listed(kind: &EventKind) {
+    match kind {
+        EventKind::TurnStart(_)
+        | EventKind::ChatRequest(_)
+        | EventKind::ChatResponse(_)
+        | EventKind::ToolCallRequest(_)
+        | EventKind::ToolCallResponse(_)
+        | EventKind::InquiryRequest(_)
+        | EventKind::InquiryResponse(_) => {}
+    }
+}
+
+/// One value of every variant.
 fn every_kind() -> Vec<EventKind> {
     vec![
         TurnStart.into(),
--- a/crates/jp_conversation/src/event_tests.rs
+++ b/crates/jp_conversation/src/event_tests.rs
@@ -1,0 +1,55 @@
+use serde_json::{Map, Value};
+
+use super::{
+    ChatRequest, ChatResponse, EventKind, InquiryId, InquiryQuestion, InquiryRequest,
+    InquiryResponse, InquirySource, ToolCallRequest, ToolCallResponse, TurnStart,
+};
+
+/// One value of every variant, so a new variant fails to compile here rather
+/// than going unnoticed.
+fn every_kind() -> Vec<EventKind> {
+    vec![
+        TurnStart.into(),
+        ChatRequest::from("hi").into(),
+        ChatResponse::message("hello").into(),
+        ToolCallRequest::new("call-1".to_owned(), "read_file".to_owned(), Map::new()).into(),
+        ToolCallResponse {
+            id: "call-1".to_owned(),
+            result: Ok("contents".to_owned()),
+        }
+        .into(),
+        InquiryRequest::new(
+            InquiryId::new("q1"),
+            InquirySource::User,
+            InquiryQuestion::text("Which file?".to_owned()),
+        )
+        .into(),
+        InquiryResponse::new(InquiryId::new("q1"), Value::Null).into(),
+    ]
+}
+
+/// The tag a variant serializes as is what every reader outside Rust switches
+/// on, so it has to be the tag serde actually writes rather than a name kept
+/// alongside it by hand.
+#[test]
+fn every_variants_tag_is_the_one_serde_writes() {
+    for kind in every_kind() {
+        let serialized = serde_json::to_value(&kind).expect("serializes");
+        let written = serialized
+            .get("type")
+            .and_then(Value::as_str)
+            .expect("carries a type tag");
+
+        assert_eq!(kind.type_tag(), written, "for {}", kind.as_str());
+    }
+}
+
+/// The deserializer decides whether an entry is a known event by looking its
+/// tag up in this list, so a tag missing from it makes the variant unreachable:
+/// the stream would keep every one of those events as raw JSON instead.
+#[test]
+fn every_variants_tag_is_listed_as_recognized() {
+    let tags: Vec<&str> = every_kind().iter().map(EventKind::type_tag).collect();
+
+    assert_eq!(tags, EventKind::TYPE_TAGS);
+}
--- a/crates/jp_conversation/src/lib.rs
+++ b/crates/jp_conversation/src/lib.rs
@@ -43,8 +43,8 @@ pub use compaction::{
 pub use conversation::{Conversation, ConversationId};
 pub use error::Error;
 pub use event::{ConversationEvent, EventKind};
-pub use storage::{decode_event_value, rfc3339, rfc3339_str};
-pub use stream::{ConversationStream, IterTurns, StreamEntry, StreamError, Turn, TurnMut};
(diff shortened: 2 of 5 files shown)


crates/jp_conversation/src/lib.rs
```diff
@@ -43,8 +43,8 @@ pub use compaction::{
 pub use conversation::{Conversation, ConversationId};
 pub use error::Error;
 pub use event::{ConversationEvent, EventKind};
-pub use storage::decode_event_value;
-pub use stream::{ConversationStream, IterTurns, StreamError, Turn, TurnMut};
+pub use storage::{decode_event_value, rfc3339, rfc3339_str};
+pub use stream::{ConversationStream, IterTurns, StreamEntry, StreamError, Turn, TurnMut};
```
JeanMertz: question: Which concrete caller needs this full public surface? In stacked PR #1007 I can find only `rfc3339`; it does not call `rfc3339_str`, `EventKind::type_tag`, or `iter_entries`. Those APIs establish contracts over raw stream entries before a consumer has shaped their required semantics. If there is no current caller, please keep only `rfc3339` in this PR and introduce the others with the consumer that defines their behavior and tests.

crates/jp_conversation/src/event_tests.rs
```diff
@@ -0,0 +1,55 @@
+use serde_json::{Map, Value};
+
+use super::{
+    ChatRequest, ChatResponse, EventKind, InquiryId, InquiryQuestion, InquiryRequest,
+    InquiryResponse, InquirySource, ToolCallRequest, ToolCallResponse, TurnStart,
+};
+
+/// One value of every variant, so a new variant fails to compile here rather
+/// than going unnoticed.
+fn every_kind() -> Vec<EventKind> {
```
JeanMertz: theoretical: I cannot name a current event that triggers this because every current variant is listed. The claimed compile-time exhaustiveness is not present, though. Adding `EventKind::Foo` and its required `type_tag` arm while forgetting both this vector and `TYPE_TAGS` leaves both tests green; a stored `Foo` event then reloads as `Unknown` in release builds. Generate the enum, tags, and test cases from one source, or otherwise make an omitted fixture case fail to compile.

crates/jp_conversation/src/stream.rs
```diff
@@ -153,6 +153,30 @@ impl InternalEvent {
...
+/// to somebody needs to see all of it.
+/// This is that view: what an entry is, without exposing the storage encoding
+/// [`InternalEvent`] carries.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum StreamEntry<'a> {
+    /// An event in the conversation.
+    Event(&'a ConversationEvent),
+
+    /// A change to the configuration every later entry is bound to.
+    ConfigDelta(&'a ConfigDelta),
```
JeanMertz: nit: A plausible input is an event written by a newer JP whose metadata contains an encoded string such as `{"note":"c2VjcmV0"}`. `InternalEvent::Unknown` is retained before `decode_event_value` runs, so this reference exposes storage JSON while `Event` exposes decoded domain values. A reader following “readable only as JSON” can present base64 as content. Please document `Unknown` as opaque, storage-encoded JSON that readers should not present directly, or provide a separate decoded view without changing the raw value retained for round-tripping.

crates/jp_conversation/src/lib.rs
JeanMertz: Verified, and acted on: `rfc3339_str`, `StreamEntry` and `iter_entries` are gone.

Your reading of #1007 holds up — I checked it rather than take it on trust. `jp_ffi/src/display.rs` imports `ConversationEvent, EventKind, event::ChatResponse, rfc3339, stream::ConversationStream` and iterates with `iter_events_by_turn`. It touches none of the three.

Two design documents also point away from that surface, which I should have checked before writing it:

- **RFD 099**, which specifies the app meant to consume this, lists as a non-goal: *"Tool calls, attachments, and reasoning display. Events other than user messages and assistant messages render as a minimal placeholder."* That is exactly the set `iter_entries` exists to yield.
- **RFD 097** Phase 2 decides against it outright: *"config deltas, compactions, and unknown entries gain no new programmatic surface (they remain ID-addressable in the persisted JSON)."*

Removed along with them: the `parse_dt` import in `storage.rs`, and its `pub(crate)` bump — which turned out to be unnecessary in the first place, since `parse_dt` sits at the crate root and `stream.rs:251` and `stream.rs:1743` were already calling it as `crate::parse_dt` while it was private. `stream.rs` is now byte-identical to `main`.

**One thing I'd keep against your "only `rfc3339`":** `EventKind::type_tag`. You're right that it has no production caller either, but its two tests pin an invariant with real consequences — that a variant's serde tag agrees with `TYPE_TAGS`, and a mismatch makes the variant unreachable on load, silently degrading stored events to `Unknown`. The tests are the value and `type_tag` is what lets them be written. That reads differently to me than an untested enum nothing constructs, but say the word if you disagree and it goes too.

crates/jp_conversation/src/event_tests.rs
JeanMertz: The doc comment was making a promise the code doesn't keep — `every_kind` is a `vec![]` literal, so nothing about it fails to compile. That claim is worse than no claim, so it's gone either way.

Mapping out what actually guards this, because it's more than nothing and less than the comment implied:

- `type_tag`'s match (`event.rs:428`) — compile error on a new variant.
- `encode_event`'s match (`storage.rs:94-102`) — compile error, and it carries an explicit `NOTE: don't forget to update decode_event_value`.
- `stream.rs:1854-1862` — a `debug_assert` that fires with *"event tag `{tag}` is missing from `EventKind::TYPE_TAGS`"* when an unknown-tagged event turns out to deserialize as a known one. Debug-only and requires loading such an event, which is why you scoped your claim to release builds.

So a contributor already hits two compile errors before reaching the test file. The genuinely unguarded thing is `TYPE_TAGS`, and the tests don't close it: `assert_eq!(tags, EventKind::TYPE_TAGS)` catches a variant added to one but not the other, and passes when it's missing from both — because the left-hand side is derived from `every_kind`, so an absent variant is absent from both sides of the comparison.

Rather than generate the enum, tags and fixtures from one source, I added an exhaustive `match` in the test file whose only job is to fail compilation:

```rust
#[expect(dead_code)]
fn every_variant_is_listed(kind: &EventKind) {
    match kind {
        EventKind::TurnStart(_)
        | EventKind::ChatRequest(_)
        // …
        | EventKind::InquiryResponse(_) => {}
    }
}
```

That closes the missing-from-both gap and puts the error in the file that needs editing, for about ten lines. Codegen felt disproportionate for a `theoretical:` item, but it's the right answer if `EventKind` grows much past seven variants.

crates/jp_conversation/src/stream.rs
JeanMertz: Correct in every particular, and it's resolved by deletion rather than documentation — `StreamEntry` is gone (see the thread on the `pub use` line).

Recording the mechanism anyway, because it's worth knowing if this surface is ever reintroduced. `stream.rs:1864` returns `Self::Unknown(value)` and `decode_event_value(&mut value)` doesn't run until line 1868, on the recognized path only. So `Unknown` held the raw storage JSON while `Event` held a fully decoded typed value — two variants of one enum at two different encoding levels, with nothing in the type or the docs saying so.

Your example is stronger than "plausible": `encode_event` base64-encodes `metadata` **unconditionally for every event kind** (`storage.rs:84-87`), before the per-variant match. So *any* event a newer JP writes with string metadata arrives encoded, not just one that happens to carry an encoded field. My own doc line — *"readable only as JSON: there is no typed form of an entry this build has never heard of"* — actively invited a reader to render it.

For the record on your two options: the first (document it as opaque) was the cheap fix, and the second (a separate decoded view) was the only one that could have been correct for a reader, because decoding in place is not available — `Unknown` is retained verbatim precisely so it round-trips losslessly (`stream.rs:51-56`) and `Serialize` passes the raw value straight through (`stream.rs:103`), so decoding it would write unencoded JSON back to disk.

I'd also revise the `nit:` label upward if this surface returns. Base64 rendered as message content is wrong output a user cannot distinguish from real content, which is the hidden kind rather than the visible kind.

PR #1006, opened 2026-08-19
feat(plugin, cli): Report `pinned_at` in conversation summaries

JeanMertz:
A plugin listing conversations received the title, the last activation
time and the event count, but nothing about pinning. A pinned
conversation was indistinguishable from any other, so a plugin could
neither mark it nor sort by it without opening every conversation to
find out.

`ConversationSummary` carries the `pinned_at` timestamp the conversation
metadata already holds. The field is skipped when serializing an
unpinned conversation and defaults to absent when reading, so a plugin
built against the older shape keeps working and one built against the
newer shape reads an older host's messages.

### DIFF
--- a/crates/jp_plugin/src/message.rs
+++ b/crates/jp_plugin/src/message.rs
@@ -166,6 +166,10 @@ pub struct ConversationSummary {
     /// When the conversation was last activated.
     pub last_activated_at: DateTime<Utc>,
 
+    /// When the conversation was pinned, absent if it is not pinned.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub pinned_at: Option<DateTime<Utc>>,
+
     /// Number of events in the conversation.
     pub events_count: usize,
 }
--- a/crates/jp_cli/src/cmd/plugin/dispatch.rs
+++ b/crates/jp_cli/src/cmd/plugin/dispatch.rs
@@ -284,6 +284,7 @@ fn handle_list_conversations(workspace: &Workspace, req_id: Option<String>) -> H
             id: id.as_deciseconds().to_string(),
             title: meta.title.clone(),
             last_activated_at: meta.last_activated_at,
+            pinned_at: meta.pinned_at,
             events_count: meta.events_count,
         })
         .collect();
--- a/crates/jp_plugin/src/lib_tests.rs
+++ b/crates/jp_plugin/src/lib_tests.rs
@@ -1,5 +1,3 @@
-use serde_json::json;
-
 use crate::message::*;
 
 #[test]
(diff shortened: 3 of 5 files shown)


crates/jp_plugin/src/message.rs
```diff
@@ -166,6 +166,10 @@ pub struct ConversationSummary {
     /// When the conversation was last activated.
     pub last_activated_at: DateTime<Utc>,
 
+    /// When the conversation was pinned, absent if it is not pinned.
```
JeanMertz: nit: Please pin this wire contract with serialization and deserialization tests. The plausible inputs are a pinned summary (`Some(timestamp)`), an unpinned summary, and a response from an older host with no `pinned_at` key. The existing test fixtures only set this to `None`; they do not verify that `Some` is emitted, `None` is omitted, or a missing key becomes `None`. Those are the compatibility guarantees the PR depends on.

JeanMertz: Taking this — and it's more than a nit, because `message_tests.rs` already applies exactly this pattern to every other optional field on the protocol types (`paths_info_omits_none_fields`, `host_init_paths_default_when_absent`, `plugin_exit_without_reason_deserializes`, `plugin_describe_without_command_deserializes`). `pinned_at` was the only one without it.

Confirmed all three gaps at this head: `lib_tests.rs`'s single test asserts only about `id`, `serve-web`'s `list_conversations_roundtrip` asserts only `id` and `title`, and `message_tests.rs` never mentioned `ConversationSummary`. Nothing pinned the shape.

Added three tests to `message_tests.rs`, with fixed timestamps so the expected JSON is a literal rather than something built at runtime:

- `conversation_summary_emits_pinned_at_when_set`
- `conversation_summary_omits_pinned_at_when_unset`
- `conversation_summary_defaults_pinned_at_when_absent`

I checked they actually fail rather than just pass, by stripping `#[serde(default, skip_serializing_if = ...)]` and re-running. That turned up something worth recording:

**`omits_pinned_at_when_unset` goes red** without `skip_serializing_if`, catching `"pinned_at":null` — so it guards the guarantee the commit message makes.

**`defaults_pinned_at_when_absent` passed anyway.** serde already treats `Option<T>` as optional on deserialize, so the `default` in that attribute is a no-op. The backwards-compatibility half of the claim holds, just not for the reason the attribute suggests. I've left `default` in place — it's harmless and states the intent — but it isn't what makes the behaviour work, and the test now pins the behaviour regardless of which attribute provides it.

Also tightened the pre-existing assertion in `lib_tests.rs` while I was next to it:

```rust
assert!(json.get("id").is_none() || json.get("id") == Some(&json!(null)));
```

That passes whether `id` is omitted *or* serialized as null — i.e. whatever the code does — despite the test being named `conversations_response_serializes_without_null_id`. Now `assert_eq!(json.get("id"), None, "an absent id must be omitted, not null")`, which can only pass while the key is genuinely absent.

PR #1007, opened 2026-08-19
feat(ffi): Read JP workspaces over a C ABI

JeanMertz:
A native reader cannot call `jp_workspace` without a boundary, and the
options for building one are all worse than a direct link. Shelling out
to `jp --format json` binds to shapes `jp_cli::output` derives from table
rows, which are not stable and would be frozen by Hyrum's Law the moment
anything depended on them, and pays a full workspace load per read.
Reading `.jp/` from the other language reimplements the storage format
twice over and makes the on-disk layout a public contract. A sidecar
process buys failure isolation that a read-only viewer does not need,
at the cost of supervision and a protocol.

`jp_ffi` compiles as a static library beside an rlib and exposes six
entry points: open a workspace, list its conversations, read one
conversation's turns, close the handle, free a returned string, and
collect the last error. The rlib is what lets those entry points be
unit-tested in-process rather than only through a linked app.

Four rules hold the boundary together. Every entry point catches panics,
because unwinding into the calling language is undefined behavior; a
caught panic becomes a null return and a message. Failures return null
and leave that message in a thread-local slot for `jp_last_error`. Only
owned data crosses: a read copies out and drops its lock guard before
returning, so no guard, reference or borrow escapes. And Rust frees what
Rust allocates, which is what `jp_string_free` is for.

Reads also measure their own phases and report them through an optional
out-parameter. The timings ride back on the call that produced them
rather than on a call of their own, so they cannot be attributed to the
wrong read when two overlap, and they are durations rather than
timestamps because the caller already has a clock and two that nearly
agree are worse than one. Measuring happens here; writing does not.

`display` decides what a reader shows. Which events carry prose, and
where turn boundaries fall, are judgements about the conversation model
rather than about any one reader: events before the first `TurnStart`
form an implicit leading turn, and a `TurnStart` opens a new one only
when the turn before it holds something. Neither is recoverable from the
event shape, so both belong on this side rather than being re-derived by
every reader. It reads the typed stream instead of a serialized copy,
which would base64-encode the fields storage encodes only to decode them
again, reparse every timestamp, and allocate a second copy of the
conversation to read four fields off it.

The crate depends on `jp_workspace`, `jp_conversation` and `jp_plugin`.
Not `jp_config`, and not `jp_cli`: a reader has no per-tool style to
apply, no reasoning display mode and no hidden-tool filtering, so
keeping `jp_config` out keeps the layered load pipeline out with it.

`just build-ffi` builds the library, stages it with a generated header
under `apps/macos/.build/<profile>`, and is what Xcode invokes from a
build phase so there is one build entry point rather than two competing
ones. It asks cargo where it writes rather than assuming `./target`,
because the target directory is redirectable and sibling worktrees here
share one outside the checkout.


crates/jp_ffi/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+    guard("jp_workspace_open", || {
+        // SAFETY: `path` is NUL-terminated per this function's contract. The
+        // borrow does not escape: it is consumed by `Workspace::open` below,
+        // which copies the path, well before this call returns to the caller
+        // that owns the string.
+        let path = unsafe { borrow_str(path, "path") }?;
+        let mut workspace = Workspace::open(Utf8Path::new(path)).map_err(|e| e.to_string())?;
+
+        workspace.load_conversation_index();
+
```
JeanMertz: blocking: A user who opens this handle, reads a conversation, and then runs `jp query` in another terminal will keep getting the metadata/event stream cached in `Workspace`'s `OnceLock`s; conversations created after this one-time index load never appear either. The JSON still looks valid, so the reader silently shows a stale workspace. #968 fixed this lifetime issue for plugin reads by rerunning `load_conversation_index()` before both list and event reads, which clears both caches. Please give this boundary equivalent refresh semantics, for example by synchronizing mutable access to `Workspace` and reloading before each read, and add a regression test that modifies the store between two reads through one handle.

crates/jp_ffi/src/display.rs
```diff
@@ -0,0 +1,127 @@
...
+    // `iter_events_by_turn` rather than `iter_turns`: the latter resolves and
+    // clones the accumulated config for every event and materializes the whole
+    // stream up front, and none of that is read here.
+    for (index, event) in stream.iter_events_by_turn() {
+        let Some(event) = project_event(event) else {
+            continue;
+        };
+
+        match turns.last_mut() {
+            Some(turn) if turn.index == index => turn.events.push(event),
```
JeanMertz: nit: A provider response split into adjacent message items is a normal stored shape; the CLI renderer explicitly treats consecutive same-kind responses as one Markdown region. This pushes a separate `AssistantMessage` for each item, so a reader repeats the speaker header and parses each fragment as independent Markdown. A fence or word split at the item boundary will render incorrectly. The caller cannot repair this reliably because dropping non-message events makes `message, message` and `message, tool call, message` project to the same shape. Please coalesce only consecutive raw `ChatResponse::Message` events, reset that coalescing on any intervening event, and cover a split Markdown construct in a test.

justfile
```diff
@@ -127,6 +128,59 @@ stage-and-commit: _install-jp
...
+    if [ "{{PROFILE}}" = "debug" ]; then
+        cargo build {{quiet_flag}} --package jp_ffi
+    else
+        cargo build {{quiet_flag}} --package jp_ffi --profile "{{PROFILE}}"
+    fi
+
+    # Ask cargo where it writes rather than assuming `./target`. The target
+    # directory is redirectable, and sibling git worktrees here share one that
+    # sits outside the checkout entirely.
+    target_dir=$(cargo metadata --format-version=1 --no-deps | jq -r '.target_directory')
```
JeanMertz: nit: With a configured Cargo build target, for example `[build] target = "aarch64-apple-darwin"` or `CARGO_BUILD_TARGET`, Cargo writes this artifact under `target/<triple>/<profile>/`, while `cargo metadata` still reports only the outer target directory. `just build-ffi` then reports that Cargo produced no library even though the build succeeded. This is also the path needed for architecture-specific Xcode builds. Consider consuming the `compiler-artifact.filenames` entry from `cargo build --message-format=json` instead of reconstructing the artifact path.

crates/jp_ffi/src/lib.rs
JeanMertz: Fixed, and you were right on a point I initially pushed back on.

My first read was that this couldn't be what #968 did, because `handle_list_conversations` and `handle_read_events` both take `&Workspace` while `load_conversation_index` needs `&mut self`. I was looking one level too low in the call stack — #968 puts `refresh_conversations(&mut Workspace)` in `handle_request`, which already holds the mutable borrow, and leaves the leaf handlers alone. I should have opened the PR instead of inferring it from this tree, where it's absent only because it's unmerged.

The same room exists here: the handle arrives as `*mut WorkspaceRef` and `borrow_workspace` was merely narrowing it to `&`.

Both readers now re-read the index before answering. Two things worth flagging in the implementation:

**The synchronisation you asked for is load-bearing, not incidental.** Taking `&mut` through a raw pointer makes two overlapping calls on one handle instant UB, and overlapping calls are the expected usage — the app's convention is that FFI reads run off the main thread. So `WorkspaceRef` holds a `Mutex<Workspace>`. `parking_lot`'s, specifically: `guard()` catches panics, and a panic unwinding through a `std` guard would poison the mutex and turn one caught panic into a permanently dead handle.

**The refresh runs before `acquire_conversation`, not after.** D36 notes that replacing the index under a live `ConversationLock` detaches its `Arc` from the cache. Ordering it first means there's never a held conversation at the moment the index is replaced.

Timings gained an `index.refresh` span in both entry points, and `a_failed_read_still_writes_the_timings_slot` changed premise — a read that fails at the ID lookup now reports the one span it completed rather than an empty array.

Two regression tests, both of which I confirmed fail with the refresh removed:

- `a_read_sees_a_conversation_created_since_the_handle_was_opened`
- `a_read_sees_turns_appended_since_the_conversation_was_last_read` — the interesting one, since the stream cache is per conversation and separate from the index, so a reader could have the right list and a stale transcript for the one it's showing.

RFD 099's "Live updates" non-goal contradicted this, so it's been narrowed rather than deleted: the boundary is now fresh per call, but the app still has no trigger of its own. Filed T-056wsst for the app re-requesting on window activation; watching the store stays D36's problem.

crates/jp_ffi/src/display.rs
JeanMertz: Fixed. `project_turns` now joins adjacent assistant messages.

Checked the premise before changing anything: `render/turn_view.rs:148` states the rule ("Consecutive text-bearing responses of the same kind form one markdown region; the separation between them comes from separators in the content, not from the response boundary"), and `chat_tests.rs:1076` pins it — two `Message` events rendering as `"First paragraph\n\nSecond paragraph\n\n"`. So the shape is real and the terminal already handles it.

Your unrepairable-downstream point is the one that settled it. There was a reasonable argument that joining is the caller's business, but the payload wasn't leaving the caller a decision to make: everything without prose is dropped, so `message, message` and `message, tool call, message` arrived identical. That isn't "the caller decides", it's no policy at all.

The run is tracked over the stream rather than over the projection, so an event that shows nothing still ends it — a tool call between two messages means the provider closed the first and opened the second. Reasoning ends it too, matching the same-kind rule. Fragments join without a separator, since segmentation travels in the content, and the region keeps the timestamp it opened at.

Three tests, the first with a code fence split across the boundary as you asked:

- `joins_a_message_split_across_adjacent_items`
- `keeps_messages_apart_when_a_tool_call_separates_them`
- `keeps_messages_apart_when_reasoning_separates_them`

I ran the first against the unjoined code to see it red, and the failure is exactly the reported bug: `"Here:\n\n```rust\nfn main() {"` and `"}\n```\n"` as two separate messages, one with an unclosed fence and one with a stray closer.

One thing this gives up, for the record: the app can no longer see where one response item ended, which would matter for a debugging view or for showing an interrupted response. RFD 099's non-goals exclude both today, and adding a boundary marker to the payload later is a compatible change — whereas shipping fragments and having readers depend on joining them would not be.

justfile
JeanMertz: Taken, and implemented the way you suggested — the recipe now reads the artifact path out of `compiler-artifact` rather than reconstructing it:

```sh
lib=$(cargo build {{quiet_flag}} --package jp_ffi --profile "$build_profile" \
        --message-format=json-render-diagnostics |
    jq -r 'select(.reason == "compiler-artifact" and .target.name == "jp_ffi")
           | .filenames[] | select(endswith(".a"))' |
    tail -n 1)
```

`json-render-diagnostics` rather than plain `json`, so compiler errors still reach the terminal instead of disappearing into the pipe.

This is more worth doing than the failure mode alone suggests, for two reasons beyond the one you named. `.gitignore` already tracks `/.cargo/config.toml` as a developer-local file, so `[build] target = …` in a checkout is something this repo explicitly anticipates rather than an exotic setup. And asking cargo also retires the *other* hand-coded assumption about its layout — the `dev`-builds-into-`debug` special case, which now only picks the `--profile` flag instead of also having to predict the output directory.

Worth noting what I couldn't check: I have no way to run `just build-ffi` from here, so the jq filter and the `--quiet` interaction are unverified. A `just build-ffi` and a `just build-ffi release` before this merges would be worth it.

PR #1008, opened 2026-08-19
feat(macos): Add a native app for browsing conversations

JeanMertz:
Conversations are readable from the terminal and, through the plugin
system, from a browser. Neither suits browsing: scrolling back through
weeks of turns, skimming several conversations side by side, or reading
on the machine where the work happened. This is a real Mac app for that,
with native windows, tabs, menus and state restoration rather than a web
view in a window.

It is a viewer. It opens a workspace and reads conversations, never
writes them, and has no compose field. A window shows the conversation
list beside the transcript, split by a divider whose position is
restored across launches, and opening any directory inside a workspace
opens that workspace, matching how `jp` itself walks up to find `.jp`.

The scene is a plain `WindowGroup`. Keying it by workspace path made a
window's identity its workspace, so ⌘N on a workspace already on screen
brought that window forward instead of opening one, and ⌘T had nothing
to duplicate. Each window decides which workspace it shows and keeps
that choice in its own scene storage.

Timestamps arrive as text and stay that way. The Rust side emits a
fractional-seconds part whenever the stored value has one, and
`JSONDecoder`'s `.iso8601` strategy rejects those, so a `Date` on the
Swift struct would decode the whole-second case and fail on every real
workspace. `ConversationDate` parses at the point of display instead.

`ConversationSummary` and the event payloads are hand-maintained mirrors
of Rust types with no compiler checking that they agree, so each one has
a test pinning the exact JSON it decodes. The timings payload is pinned
on both sides — `WorkspaceReaderTests` and `jp_ffi`'s `timing_tests`
hold the same literal — and if one is edited alone the other is what
says so.

`project.yml` is the source of truth for the Xcode project, which is
generated by `just gen-app` rather than committed, so targets, build
settings and the Rust build phase stay reviewable. Swift 6 language mode
with complete concurrency checking is on and warnings are errors,
holding the app to the bar the Rust side is held to. Debug builds carry
`dwarf-with-dsym` and disable Xcode's debug dylib split, because a
profiler pointed at the bundle's executable otherwise reads a launcher
stub whose UUID matches nothing in the trace.

`just run-app` launches it in the foreground with output attached, and
`just open-app` goes through LaunchServices for the AppKit behavior that
only a normally registered launch produces.


apps/macos/Sources/WorkspaceReader.swift
```diff
@@ -0,0 +1,243 @@
...
+/// the UI stalls behind a slow read.
+struct WorkspaceReader: ~Copyable {
+    private let handle: OpaquePointer
+
+    /// Open the workspace containing `path`, which may be the workspace root or
+    /// any directory inside it.
+    ///
+    /// Opening writes to disk: it creates the user-local conversation store if
+    /// missing and moves corrupt conversations aside, as `jp` does on startup.
+    init(path: String) throws(WorkspaceError) {
```
JeanMertz: blocking: A plausible input is an older workspace whose `.jp/.id` is absent or unreadable, or a workspace that has not been migrated to user-local storage. `jp_workspace_open` calls `Workspace::open`, which persists the workspace ID, creates the user-local silo and symlink, and can import workspace conversations during migration. Merely viewing that workspace therefore mutates the workspace and user data, despite the viewer's stated "never writes" contract; the workspace mutation stays hidden until the user checks the filesystem or `git status`. This needs a read-only workspace/FFI open path. Calling `disable_persistence` after opening would be too late because these writes happen inside `Workspace::open`.

apps/macos/QA.md
```diff
@@ -0,0 +1,344 @@
+# QA checklist
+
+The behavior in [`AFFORDANCES.md`](AFFORDANCES.md) that has to be checked against
+a running app.
+
+Each item says who checks it:
+
+- **`JPUITests/<test>`** — a committed test in `apps/macos/UITests`. CI runs the
```
JeanMertz: question: Were the UI-test files omitted from this commit? The PR head has no `apps/macos/UITests` directory, `JPUITests` target, `just test-app-ui` recipe, or `swift_test_ui` tool, although this checklist marks many behaviors as covered by them. This also makes `ClipboardPolicyTests.uiTestSources()` throw when `just test-app` reaches it because `contentsOfDirectory` is called on the missing directory. As committed, neither the unit suite nor the documented UI suite can run as described.

apps/macos/Sources/WorkspaceWindow.swift
```diff
@@ -0,0 +1,527 @@
...
+    /// and leaves a harness unable to point an instance anywhere.
+    /// `nonisolated` because it reads none of the view's state. A `View` is
+    /// main-actor isolated and its statics inherit that, which this does not need.
+    nonisolated static func chooseWorkspace(
+        stored: String?,
+        mostRecent: URL?,
+        environment: [String: String]
+    ) -> String? {
+        if let named = environment["JP_WORKSPACE"], !named.isEmpty {
+            return named
```
JeanMertz: nit: Launching with `just run-app A` leaves `JP_WORKSPACE=A` in the process environment. If the user then chooses workspace B with File > Open Workspace or Open Recent, `show` stores B, this task reruns, and this branch immediately chooses A and stores it again. The explicit open appears to do nothing. Treat the environment value as an initial launch choice, or otherwise let an explicit `show(_:)` take precedence after the window has opened.

apps/macos/Sources/WorkspaceWindow.swift
```diff
@@ -0,0 +1,527 @@
...
+    /// When the conversations on screen were read.
+    ///
+    /// What the rows date themselves against. Fixed at the moment of the read
+    /// rather than taken fresh per render, because a render happens on every frame
+    /// of a divider drag and a clock reading that changes each time would make the
+    /// list unequal to itself and undo the skipping that keeps the drag smooth.
+    ///
+    /// The cost is that "21 minutes ago" is 21 minutes after the workspace was
+    /// opened, not after now.
+    @State private var listingReadAt = Date()
```
JeanMertz: nit: A user who leaves the reader open keeps the same relative labels indefinitely. A row that says `21 minutes ago` at load still says that hours later, and it remains on the same-day branch after midnight because `now` never advances. The wrong value is visible and limited to the list, but relative time should either refresh on a coarse timer, such as once per minute, or use a non-relative format that does not require refreshing.

apps/macos/Sources/ConversationEvent.swift
```diff
@@ -0,0 +1,91 @@
...
+        case type
+        case timestamp
+        case author
+        case text
+    }
+
+    init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let timestamp = try container.decode(String.self, forKey: .timestamp)
+
```
JeanMertz: theoretical: The forward-compatibility path only skips an unknown presentation if that payload also has a valid `timestamp`, because this decode happens before the `type` switch. An added presentation with a different shape would throw `keyNotFound` here and fail the whole conversation rather than reaching `UnknownPresentation`. Decode `type` first, then decode `timestamp` inside the two known branches. I cannot name a current shipped input because the FFI library is statically bundled with the matching app; this affects the stated compatibility guarantee rather than today's payload.

apps/macos/QA.md
```diff
@@ -0,0 +1,344 @@
...
+points the app at a private pasteboard of its own and reads that back. The
+variable is compiled out of a release build, and an unset one means the system
+pasteboard, so the shipped behaviour is the only behaviour a user can get.
+
+`ClipboardPolicyTests` scans `apps/macos/UITests` and fails on any spelling of
+the system pasteboard, so this holds without anyone remembering it.
+
+### Multiple selection is not implemented
+
+`AFFORDANCES.md` says the context menu acts on the whole selection, so three
```
JeanMertz: question: Which behavior is intended for this release? A user can plausibly shift-click three rows expecting the affordance contract's multi-selection behavior, but `List` binds a single `String?`, so only one URI is copied or opened. Leaving the contract explicitly unsettled also leaves no testable acceptance criterion. Please either bind a `Set<ConversationSummary.ID>` and cover the multi-row actions, or change `AFFORDANCES.md` to specify single selection.

apps/macos/AFFORDANCES.md
```diff
@@ -0,0 +1,276 @@
...
+- **The window holds its two panes itself**, rather than in a
+  `NavigationSplitView`. `NSSplitView` draws a translucent divider over whatever
+  is behind it and offers no way to change its colour or width, which left the
+  line between the panes two pixels of two different greys that shifted with the
+  content underneath. The divider is now the app's own: two points, one colour,
+  and draggable through a wider invisible strip around it.
+- **Restored per window**: the workspace, the selected conversation, the
+  transcript's scroll position, the sidebar's width, and whether the sidebar is
+  showing.
+- **One window per workspace**, keyed by the workspace path. Opening the same
```
JeanMertz: nit: This contradicts the unkeyed `WindowGroup` implemented in `JPApp` and the PR's rationale for allowing Command-N to create another window on the same workspace. It also does not hold when window B uses Open Recent to choose the workspace already shown in window A; B switches to that path rather than bringing A forward. Update this contract and the matching QA item to describe the duplicate-window behavior the app intentionally implements.

apps/macos/AFFORDANCES.md
```diff
@@ -0,0 +1,276 @@
+# Affordance map
+
+What the JP reader responds to, and what each response is meant to do. This is
+the contract phase 4 of [RFD 099] holds the app to; the QA checklist beside it
```
JeanMertz: question: Was RFD 099 omitted from this commit? The linked `docs/rfd/099-native-macos-app-for-browsing-conversations.md` does not exist at either the PR base or head, so the design contract this document relies on is unavailable and the link is broken.

apps/macos/AFFORDANCES.md
```diff
@@ -0,0 +1,276 @@
...
+- **No sidebar toggle button**, because there is no title bar to put one in.
+  View ▸ Hide Sidebar (⌃⌘S) is how the sidebar is hidden and brought back.
+- **The window holds its two panes itself**, rather than in a
+  `NavigationSplitView`. `NSSplitView` draws a translucent divider over whatever
+  is behind it and offers no way to change its colour or width, which left the
+  line between the panes two pixels of two different greys that shifted with the
+  content underneath. The divider is now the app's own: two points, one colour,
+  and draggable through a wider invisible strip around it.
+- **Restored per window**: the workspace, the selected conversation, the
+  transcript's scroll position, the sidebar's width, and whether the sidebar is
```
JeanMertz: question: What restores the transcript scroll position across a process relaunch? `TranscriptTextView` creates a plain `NSScrollView`; its coordinator tracks only the built conversation, turn count, appearance, and a notification token. The scroll view has no restoration identifier, and this window stores no offset in scene storage. A user who quits halfway through a long transcript therefore appears to reopen at the top. If AppKit is restoring this through another mechanism, a test should pin it; otherwise this item belongs in the implementation or the not-implemented list.

apps/macos/Sources/WorkspaceReader.swift
```diff
@@ -0,0 +1,243 @@
...
+            Self.conversationsSpan, target: Self.traceTarget, spans: spans)
+        defer { timing.end() }
+
+        // Asked for only while something is listening. Unrecorded, the library
+        // allocates no timings string and nothing here has one to release.
+        var timings: UnsafeMutablePointer<CChar>?
+        let raw =
+            Trace.isRecording
+            ? jp_workspace_conversations(handle, &timings)
+            : jp_workspace_conversations(handle, nil)
```
JeanMertz: blocking: A plausible input is a workspace containing one conversation with malformed or partially written `metadata.json`. `Workspace::conversations()` drops that entry after `ensure_all_metadata_loaded` logs the load error, and `jp_workspace_conversations` still returns a successful JSON array. This layer therefore presents an apparently complete list with the conversation silently missing; a normal GUI launch gives the user no indication that the result was partial. The FFI response needs to carry per-entry load failures, or fail the list read with enough context for the sidebar to show that some conversations could not be loaded.

apps/macos/AFFORDANCES.md
JeanMertz: Correct, nothing restores it. `@SceneStorage` in `WorkspaceWindow` covers `workspacePath`, `selectedConversation`, `sidebarVisible` and `sidebarWidth`; `TranscriptTextView.makeNSView` builds a plain `NSScrollView` with no restoration identifier. The document was describing an intention.

Fixed by correcting the document rather than implementing it: `the transcript's scroll position` is out of the Restored-per-window bullet, and `## Not implemented` gains

> **The transcript's scroll position is not restored.** A relaunched window reopens at the top of the conversation it had selected.

RFD 099 phase 4 does list scroll position, so this stays on the roadmap — it just isn't in this PR, and `AFFORDANCES.md` should not have claimed otherwise. QA.md already had the item unchecked, so the two documents now agree.

apps/macos/AFFORDANCES.md
JeanMertz: RFD 099 is a separate PR, cut off `main` with no dependency on this one — a design document doesn't need to be gated behind its implementation. The link resolves once both land.

You have surfaced a real ordering constraint though, which I hadn't thought about: nothing enforces which merges first. If this PR goes in ahead of the RFD, `main` carries a broken relative link until the RFD lands. The RFD PR is a single file with nothing blocking it, so it should merge first. Noted and I'll sequence it that way.

apps/macos/QA.md
JeanMertz: Two halves, and the second one is a real break.

The UI test files are a separate PR stacked directly on this one, so the checklist is describing the pair. That part is expected.

`ClipboardPolicyTests` is not. You're right that it throws: `uiTestSources()` calls `contentsOfDirectory(at:)` on `apps/macos/Tests/../UITests`, which does not exist at this commit, inside a `throws` test with no `catch`. `just test-app` fails on this PR alone. That's a mistake in how I split the branch — the file lives in `Tests/` so I assigned it by directory, but its subject is `UITests/`, and its `#expect(sources.count >= 3)` guard is deliberately built to fail on an empty scan, so loosening it to tolerate a missing directory would defeat the thing it exists to do.

Removed from this PR; it moves to the UI tests PR at the same path, alongside the sources it scans. `just test-app` now passes here: 146 tests, 21 suites.

It also exposed a gap in how the split was verified — every branch got `cargo check --locked`, but the Swift suite only ran on the top of the stack, where this passes. Running it per-branch from now on.

apps/macos/AFFORDANCES.md
JeanMertz: Right on both counts. `JPApp` uses `WindowGroup(id: "workspace")` with no `for:` key, and this PR's own commit message explains why keying was abandoned — ⌘N brought the existing window forward instead of opening one, and ⌘T had nothing to duplicate. The bullet describes the design that was removed. Your second point holds too: `show(_:)` writes `workspacePath` on the focused window, so Open Recent moves that window rather than surfacing another.

Rewritten to describe what the app does:

> **Windows are not keyed by workspace.** ⌘N opens another window on the same workspace rather than bringing the existing one forward, and two windows may show one workspace. Each window holds its own choice in scene storage. Paths are still canonicalized, so the recents list and a window's stored path agree on spelling.

The canonicalization sentence is kept because that part is still true and still load-bearing for the recents list; it just isn't a window key.

apps/macos/QA.md
JeanMertz: Taking the second option: the affordance map is wrong, not the code.

`ConversationList` binds `@Binding var selection: ConversationSummary.ID?` and `WorkspaceWindow.selectedIDs` is `selection.map { [$0] } ?? []`, so one conversation, always. RFD 099 doesn't ask for multi-select, and adding it here would be a feature arriving without a design or a test.

`AFFORDANCES.md` now says:

> The list holds one conversation at a time, so the context menu acts on that one: Open in New Window opens a single window and Copy Link copies a single URI. Shift-clicking does not extend the selection.

and QA.md's `### Multiple selection is not implemented` section is deleted, since there is no longer a disagreement for it to record. You're right that leaving two documents in stated contradiction, with a third section explaining that they contradict, was the worst of the available options.

apps/macos/Sources/ConversationEvent.swift
JeanMertz: Confirmed and fixed — `timestamp` is decoded inside each branch now, so the `type` tag is read first and an unrecognized presentation reaches `UnknownPresentation` regardless of what else it carries.

Chasing it turned up something worse than the bug, which is why this is worth more than a reorder. The doc comment on `SkippableEvent` justified the whole mechanism as tolerating *"an app that predates"* a newer library — and as you note, `jp_ffi` is statically linked, so that skew cannot happen. The rationale was wrong even though the mechanism is worth keeping. What it actually protects against is an edit, not a version: a presentation added to `DisplayEvent` on the Rust side without the matching case added here, which is exactly the hand-maintained-mirror risk RFD 099 lists as a drawback. Comment rewritten to say that.

Added a test that would have caught the original bug — the existing skip test gave its unknown event a `timestamp`, which is precisely why it passed:

```swift
{"type":"tool_call","tool":"read_file","arguments":{}}
```

An added presentation is the thing least likely to share the shape of the two that exist, so the test uses one with no fields in common.

apps/macos/Sources/WorkspaceReader.swift
JeanMertz: The behaviour is exactly as you describe, but it's a decision rather than an oversight, so I'm leaving it.

RFD 099 scopes it out explicitly, under Non-Goals:

> **Read-only workspace opening.** Opening a workspace creates the user-local silo, repairs its symlink, and persists the workspace ID, exactly as `jp` does today. Making that side-effect-free is out of scope.

On the contract: I went looking for the "never writes" claim and can't find it in the form quoted. Neither `AFFORDANCES.md` nor `QA.md` says it, and the doc comment three lines above the flagged code says the opposite —

> Opening writes to disk: it creates the user-local conversation store if missing and moves corrupt conversations aside, as `jp` does on startup.

The commit message's "never writes them" refers to conversations, which does hold: nothing here writes conversation data.

A read-only open path is worth having, and I'd support it — but it belongs in `jp_workspace` where `jp` would get it too, not bolted onto the FFI. As you say, `disable_persistence` after the fact is too late; it needs `Workspace::open` to have a variant that doesn't migrate. Happy to file that if you want it tracked, though it reads more like an RFD amendment than a ticket, since RFD 099 named it as out of scope on purpose.

apps/macos/Sources/WorkspaceReader.swift
JeanMertz: Verified and filed as **T-057ee4y**.

The path is exactly as you describe: `ensure_all_metadata_loaded` logs a `warn!` and leaves the cell unset, then `conversations()` `filter_map`s on `cell.get()` and drops the entry, and the call still succeeds. A partially written or hand-edited `metadata.json` is a plausible input, especially since `jp conversation edit --events` encourages hand editing.

It's worse in the app than you flagged, in fact: `jp_ffi` installs no tracing subscriber, so the `warn!` has nowhere to go at all. In the CLI it at least reaches stderr.

Not fixing it here, for one reason: this is `Workspace::conversations()` as it exists on `main`, shared by `jp conversation ls` and `jp serve web`. Carrying per-entry failures only through `jp_ffi` would make the app and the CLI disagree about what a workspace contains, which is a worse state than both being wrong the same way. The ticket proposes surfacing them at the `jp_workspace` layer so all three consumers get it at once.

apps/macos/Sources/WorkspaceWindow.swift
JeanMertz: Accurate, and filed as **T-057ejs5** rather than fixed here.

The midnight case you name is the sharper half — `ConversationDate.label` picks its branch from `now`, so a conversation from yesterday keeps a minutes-and-hours label instead of switching to `2 Aug`.

Your suggested fix is the right one, and worth spelling out why the obvious version isn't: a fresh `Date()` per render makes `ConversationList` unequal to itself, and a render happens on every frame of a divider drag, so it would undo the skipping that keeps the drag smooth. A timer writing `listingReadAt` once a minute keeps that intact — the value changes once a minute rather than once a frame — and the labels are only accurate to the minute anyway. That's what the ticket proposes.

Left out of this PR because it's contained and visible: the reader sees a stale relative time and nothing else is affected.

apps/macos/Sources/WorkspaceWindow.swift
JeanMertz: Traced it and you're right: `show(B)` writes `workspacePath`, `.task(id:)` fires, `load()` asks `chooseWorkspace`, the environment wins, A is written back, and the window settles on A. Since `just run-app` always sets `JP_WORKSPACE`, this is the primary development workflow rather than an edge case. (Xcode's own scheme sets it to `""`, which is ignored, so running from Xcode was unaffected — which is probably why it survived.)

Fixed along the lines of your first suggestion: the environment is a launch instruction, not a standing override. `chooseWorkspace` takes a `honorEnvironment: Bool`, the window holds `hasResolvedWorkspace` and passes `!hasResolvedWorkspace`, and sets it after the first resolution.

The precedence stays as documented for the case it was written for — `JP_WORKSPACE` still outranks a stored path at launch, so `just run-app <workspace>` isn't a no-op after the first run and a harness can still point an instance anywhere. It just outranks it once.

Kept the parameter on the pure function rather than gating at the call site so it stays testable. Two cases added: with `honorEnvironment: false`, a stored path beats `JP_WORKSPACE`, and with nothing stored it falls through to the recents list.

apps/macos/AFFORDANCES.md
```diff
@@ -0,0 +1,307 @@
...
+  See `Sources/WindowButtons.swift`.
+- **No sidebar toggle button**, because there is no title bar to put one in.
+  View ▸ Hide Sidebar (⌃⌘S) is how the sidebar is hidden and brought back.
+- **The window holds its two panes itself**, rather than in a
+  `NavigationSplitView`.
+  `NSSplitView` draws a translucent divider over whatever is behind it and
+  offers no way to change its colour or width, which left the line between the
+  panes two pixels of two different greys that shifted with the content
+  underneath.
+  The divider is now the app's own: two points, one colour, and draggable
```
JeanMertz: nit: A normal workspace window draws this divider at `dividerWidth = 1`, and the matching QA item correctly calls it one point, or two pixels on a Retina display. Calling it "two points" here describes a line twice as wide as the implementation. Change this to "one point" or "two pixels" so the visual contract has one unit and value.

apps/macos/Sources/WorkspaceReader.swift
JeanMertz: triage: RFD 099 explicitly defines the viewer as not writing conversation data and lists side-effect-free workspace opening as a Non-Goal. That contract makes the silo, symlink, and ID writes intentional for this release, so this comment should not gate the PR. I withdraw the `blocking` classification.

apps/macos/Sources/WorkspaceReader.swift
JeanMertz: triage: Filing T-057ee4y records the root issue, but it does not change the merge verdict here. This PR adds the GUI path where the existing warning has no subscriber and the returned list still looks complete. A partially written `metadata.json` therefore remains a plausible, hidden omission in the new app. I still consider this `blocking` unless the app surfaces a partial read or the workspace-layer fix lands first.

apps/macos/AFFORDANCES.md
JeanMertz: nit: `AFFORDANCES.md` matches the implementation, but the old keyed-window invariant remains in `QA.md:300-301`, `RecentWorkspaces.swift:34`, the related URL/test comments, and accepted RFD 099 line 76. Since windows are intentionally unkeyed, these should describe canonical path spelling without claiming it controls window identity.

apps/macos/QA.md
JeanMertz: nit: Removing `ClipboardPolicyTests` fixes this branch's unit suite, but `QA.md:204-205` still says that test scans `apps/macos/UITests` in this PR. If the policy test belongs to the stacked UI-test PR, move this claim with it or mark it as forthcoming so the checklist describes what this head can run.

apps/macos/AFFORDANCES.md
```diff
@@ -0,0 +1,307 @@
...
+message: a text view lays out what its viewport needs and re-wraps
+incrementally, where a stack of views each measure themselves and a width change
+costs the sum of them.
+That is also why selection runs across messages rather than stopping at one, and
+why the scroll bar can state a real height instead of an estimate.
+
+**Only messages are shown.** A user message and an assistant message each render
+under the name of whoever said it.
+Tool calls, reasoning, inquiries, config changes and turn markers have no prose
+to show and never cross the FFI boundary — the library leaves them out rather
```
JeanMertz: nit: The implementation and this affordance map agree that non-message events are omitted, but accepted RFD 099 still says they "render as a minimal placeholder." A conversation containing tool calls therefore has two conflicting design contracts. Amend the RFD to say these events do not cross the FFI boundary, or change this projection to emit the promised placeholder.

apps/macos/Sources/ConversationEvent.swift
JeanMertz: theoretical: The tag-first fix handles an unknown payload, but an unknown-only turn decodes as `ConversationTurn(events: [])`. If every turn has only newly added presentations, `turns.isEmpty` is false and the app shows a loaded but blank transcript instead of the empty state. I still cannot name a shipped input because the static library and app match; for the edit-mismatch case this compatibility path targets, filter turns whose decoded `events` are empty or test emptiness by visible events.

apps/macos/AFFORDANCES.md
JeanMertz: Right, and there is a second site you did not have in view. `dividerWidth` is `1`, documented at `WorkspaceWindow.swift:437` as "One point, which is two pixels on a retina display and matches Bear" — and `WorkspaceWindow.swift:194` makes the same mistake as the affordance map, justifying the grab strip "because two points is not something a person can reliably hit".

Both now say one point. While in that line of `AFFORDANCES.md` I also dropped "now" from "The divider is now the app's own" — the reader has no earlier version to update, and the section already explains what `NSSplitView` did instead.

apps/macos/AFFORDANCES.md
JeanMertz: Confirmed at all four, and the QA item was the worst of them: `QA.md:300-301` asked a checker to verify that ⌘O and Open Recent bring an existing window forward, which nothing implements — `show(_:)` writes `workspacePath` on the focused window and there is no window-forwarding code anywhere.

What canonicalization is actually for survives the correction, so the comments say that instead of citing window identity:

- `RecentWorkspaces.note(_:)` — noting a URL as given records a second spelling of a workspace already in the list, so the menu grows two entries for one directory and neither matches the path a window is showing.
- `URL.canonicalized` — a list holding `/a/b/` compares unequal to a window showing `/a/b`, so the workspace is listed twice and choosing it reloads a window already on it.
- `QA.md` — the item now checks the real behavior: ⌘O and Open Recent change the frontmost window's workspace rather than opening or surfacing another.

RFD 099 line 76 is fixed here too rather than left for a later PR, since this PR is what makes it false. It now says a window shows one workspace but windows are not keyed by workspace, and a workspace may be open in several at once.

apps/macos/AFFORDANCES.md
JeanMertz: Verified: RFD 099:285-286 promises the placeholder, and `jp_ffi`'s `project_event` returns `None` for everything but `ChatRequest` and `ChatResponse::Message`, so the events never reach Swift and no placeholder is possible.

Amending the RFD rather than the projection. Dropping non-message events at the library boundary is the better of the two, and it is what the rest of that RFD argues for: which events have prose to show is a judgement about the conversation model, so it belongs on the side that owns the model rather than being re-derived by the terminal renderer, the web view and the app separately. The Non-Goal now says they do not cross the FFI boundary and are not shown at all, with that reasoning attached.

One consequence worth recording, because it settles a question elsewhere: a placeholder rendering is close to the only thing that would need `ConversationStream::iter_entries` and `StreamEntry`, which landed in #1005 with no caller. Amending the RFD this way means nothing is planned that uses them.

apps/macos/QA.md
JeanMertz: Correct — `ClipboardPolicyTests.swift` is deleted in this head and `apps/macos/UITests` does not exist here, so that paragraph asserted two things this branch cannot support.

Marked as forthcoming rather than removed: the surrounding prose is the policy, and the policy holds now (no test here touches the system pasteboard, and `JP_DEBUG_PASTEBOARD` is already wired). What is not true yet is that anything enforces it. The paragraph now ends: "It arrives with the UI tests it scans, so nothing enforces the rule until then."

apps/macos/Sources/ConversationEvent.swift
JeanMertz: Confirmed and taken, though it is theoretical exactly as you say — the Rust side never emits an empty turn (`DisplayTurn`: "a turn with nothing to show is absent rather than empty"), so the only way in is the Rust/Swift edit mismatch `SkippableEvent` exists for.

Taking it anyway because leaving it is a half-built compatibility path: the mechanism handles a mixed turn correctly and blank-screens an all-unknown one, which is the failure mode it was written to prevent.

The emptiness test moved out of the view into a factory on `TranscriptState`, mirroring `WorkspaceModel.state(for:)`:

```swift
case .success(let turns) where turns.allSatisfy(\.events.isEmpty):
```

That subsumes the old `turns.isEmpty` rather than adding a branch beside it — an empty array satisfies it vacuously. Moving it out is what makes it testable at all; `load()` is a view method and was not reachable from a test.

`TranscriptStateTests` covers five cases, and the two that matter are the ones the old code got wrong and right respectively: a conversation whose turns all render nothing is the empty state, and one where only some turns were emptied still shows the transcript.

apps/macos/Sources/WorkspaceReader.swift
JeanMertz: Agreed, and thanks for closing it out explicitly rather than leaving it hanging. Nothing changed here — the doc comment on `WorkspaceReader.init(path:)` already states that opening writes, so the behavior is documented where a caller meets it rather than only in the RFD.

apps/macos/Sources/WorkspaceReader.swift
JeanMertz: You are right on severity and I undersold it last time. Running the project's own test: a plausible input (a partially written or hand-edited `metadata.json`, and `jp conversation edit --events` actively encourages hand editing), damage contained to one listing, but **hidden** — the user cannot tell the list is short. Contained-and-hidden gates a merge. And your sharper point holds: the app is not merely equal to the CLI here, it is worse, because `crates/jp_ffi` installs no tracing subscriber at all, so the `warn!` has nowhere to go where the CLI's reaches stderr.

Where I still disagree is the shape, and there is now a scope fact that settles where it lands: **`jp_ffi` is not in this PR.** It merged separately as #1007 and sits in this branch's base. This PR is Swift, docs and the justfile. So neither half of the fix is reachable from here — the app cannot surface a partial read that the FFI never tells it about, and the FFI is not open for editing in this diff.

Carrying per-entry failures through the FFI response is also still the wrong shape, for the reason I gave before: it would make the app and `jp conversation ls` disagree about what a workspace contains, which is worse than both being wrong identically.

Split accordingly:

- **T-057ee4y** — the root cause, at `jp_workspace::conversations()`, where all three consumers get the fix at once.
- **T-05yf4mv** (new) — `jp_ffi` installs no subscriber, so library warnings are discarded. This is the part that makes the GUI worse than the CLI, it is one crate, and it is independent of the data-model question. Installing a stderr subscriber behind a `Once` in `jp_workspace_open` puts the app level with `jp`: `just run-app` attaches stderr, and a LaunchServices launch sends it to Console.app.

That leaves this PR with no lever to pull, so I am asking you to drop the block here and move it to T-05yf4mv, which can land without waiting on the workspace-layer redesign. If you would rather hold the app until one of the two ships, say so and I will treat it as a merge-order constraint rather than a change request — but I do not think a missing row in a read-only viewer is proportionate to blocking a branch that cannot fix it.

JeanMertz: Correcting my previous reply: asking you to move the block was the wrong way round, since I am the one who can move it. Done — it now lives on **T-05yf4mv**, recorded as comment #1 there.

That comment carries your severity assessment verbatim rather than my summary of it, because it is what justifies the ticket: plausible input, contained damage, but hidden, which is the combination that gates rather than the combination that suggests. It also records the point I had understated — that the app is worse than the CLI here rather than equal to it, since `jp conversation ls` at least reaches stderr.

Nothing about the analysis has changed, only where it is tracked:

- **T-05yf4mv** — `jp_ffi` installs no subscriber. The diagnostics half, one crate, independent of the data-model question.
- **T-057ee4y** — `jp_workspace::conversations()` drops entries silently. The root cause, where all three consumers get the fix at once.

Treating this thread as settled on that basis. If you would rather hold the app until one of the two ships, that is a merge-order call and I will not argue it — say so and I will leave the PR unmerged rather than reopen the question of where the fix belongs.

PR #1009, opened 2026-08-19
feat(drive): Add the `jpdrive` accessibility driver

JeanMertz:
Debugging the macOS app from a conversation needs a way to read what is
actually on screen and act on it. Screenshots answer "what does it look
like" but not "what is the button called, is it enabled, does the menu
item exist" — and none of that is reachable from the app's own test
bundle, which runs inside the process and cannot see menu enablement,
the pasteboard, or a relaunch.

`jpdrive` reads and drives another application through its accessibility
tree, and writes one JSON document to stdout either way: a result, or an
error with a non-zero exit status. It is what the `debug_app_*` tools
shell out to.

A standalone SwiftPM package rather than a target in the app's Xcode
project, so the binary lands at a predictable path and nothing has to
search derived data for it. The logic lives in a `DriveKit` library with
a one-line executable on top, because SwiftPM cannot cleanly test an
executable target and the tree traversal is where the bugs are. The
tests run against a fake accessibility tree, so they need no running app
and no accessibility grant.

The package mirrors the app's strictness: Swift 6 language mode,
`ExistentialAny`, and warnings as errors. SwiftPM 6.0 has no first-class
setting for the last of those, so it goes through `unsafeFlags`, which
is rejected only for a package consumed as a dependency — this one never
is.

Accessibility is gated by TCC, and a grant given to a terminal does not
obviously reach a tool that terminal started. `just drive-doctor`
reports whether the calling process may read another app's tree, so that
question is answered by running it under each host rather than guessed
at.

### DIFF
--- a/apps/macos/Tools/jpdrive/Sources/DriveKit/AXElement.swift
+++ b/apps/macos/Tools/jpdrive/Sources/DriveKit/AXElement.swift
@@ -110,12 +110,17 @@ struct AXElement {
         return value
     }
 
-    /// Read several attributes in one round-trip.
+    /// Read several attributes in one round-trip, or `nil` when the call fails.
     ///
     /// Results are positional and the same count as `names`. An attribute that
     /// could not be read arrives as CoreFoundation's null or as an `AXValue`
     /// boxing the error, both of which ``text(_:)`` reports rather than discards.
-    func values(_ names: [String]) -> [CFTypeRef?] {
+    ///
+    /// A `nil` return is the whole call failing, which a target that is busy or
+    /// exiting answers with. Reporting that as a row of absent attributes would
+    /// make an element nothing could be read from look exactly like one that
+    /// reports nothing.
+    func values(_ names: [String]) -> [CFTypeRef?]? {
         guard !names.isEmpty else { return [] }
 
         var raw: CFArray?
@@ -130,7 +135,7 @@ struct AXElement {
             let values = raw as? [CFTypeRef],
             values.count == names.count
         else {
-            return Array(repeating: nil, count: names.count)
+            return nil
         }
 
         return values
@@ -162,7 +167,13 @@ extension AXElement: Element {
     /// Children come back in the same batch as everything else: asking for them
     /// separately would add a hop per element, and every walk asks for them.
     func read(_ names: [String]) -> Reading<AXElement> {
-        let values = self.values(names + [kAXChildrenAttribute])
+        guard let values = self.values(names + [kAXChildrenAttribute]) else {
+            return Reading(
+                text: Array(repeating: nil, count: names.count),
+                children: [],
+                failed: true
+            )
+        }
 
         let children = (values.last.flatMap { $0 } as? [AXUIElement] ?? [])
             .map { AXElement(element: $0) }
@@ -171,6 +182,24 @@ extension AXElement: Element {
             text: values.dropLast().map(Self.optionalText),
             children: children
         )
+    }
+
+    /// Enumerate the element's attributes and read them all.
+    ///
+    /// Two round-trips: one to learn the names, one to read them. Only a dump
+    /// pays that, which is why it is not folded into ``read(_:)``.
+    func reportedAttributes(settable: Bool) -> [Attribute]? {
+        let names = self.names().sorted()
+        guard !names.isEmpty else { return [] }
+        guard let values = values(names) else { return nil }
+
+        return zip(names, values).map { name, value in
+            Attribute(
+                name: name,
+                value: value.map(Self.text) ?? "<null>",
+                settable: settable ? isSettable(name) : nil
+            )
+        }
     }
 
     /// Read a boolean attribute.
--- a/apps/macos/Tools/jpdrive/Sources/DriveKit/AXElement.swift
+++ b/apps/macos/Tools/jpdrive/Sources/DriveKit/AXElement.swift
@@ -1,0 +1,359 @@
+import ApplicationServices
+import Foundation
+
+/// An element of a running application's accessibility tree.
+///
+/// Every method here is one or more synchronous round-trips to the target's main
+/// thread. That cost dominates everything the driver does, so callers batch reads
+/// with ``values(_:)`` rather than reading attributes one at a time, and hold onto
+/// an element they will read again instead of walking to it twice.
+///
+/// A reference stays valid while the underlying element lives. Once it is gone,
+/// reads answer `invalid_ui_element` rather than crashing.
+struct AXElement {
+    let element: AXUIElement
+
+    /// The root element of the application owning `pid`.
+    ///
+    /// Succeeds whether or not the process exists; the first read is what fails.
+    static func application(pid: pid_t) -> AXElement {
+        return AXElement(element: AXUIElementCreateApplication(pid))
+    }
+
+    /// `AXRole`, or `nil` when the element does not report one.
+    var role: String? {
+        return read(kAXRoleAttribute).flatMap { $0 as? String }
+    }
+
+    /// `AXIdentifier`, or `nil` when the element carries none.
+    ///
+    /// SwiftUI's `.accessibilityIdentifier` surfaces here, but it also composites
+    /// with identifiers the framework generates itself, so a value like
+    /// `"workspace-AppWindow-1, SidebarNavigationSplitView"` is possible.
+    var identifier: String? {
+        return read(kAXIdentifierAttribute).flatMap { $0 as? String }
+    }
+
+    /// The element's human-readable label.
+    ///
+    /// Tries `AXAttributedDescription`, then `AXDescription`, then `AXTitle`.
+    /// SwiftUI populates the first of those for list rows and leaves the others
+    /// empty, while AppKit controls tend to do the reverse.
+    var label: String? {
+        if let attributed = read(Self.attributedDescription) as? NSAttributedString {
+            return attributed.string
+        }
+
+        for name in [kAXDescriptionAttribute, kAXTitleAttribute] {
+            guard let text = read(name).flatMap({ $0 as? String }), !text.isEmpty else {
+                continue
+            }
+            return text
+        }
+
+        return nil
+    }
+
+    /// `AXAttributedDescription`, which has no constant in the SDK headers.
+    static let attributedDescription = "AXAttributedDescription"
+
+    /// `AXActivationPoint`, which has no constant in the SDK headers.
+    ///
+    /// Where the element says a click on it belongs, in screen coordinates. Not
+    /// always the middle of its frame.
+    static let activationPoint = "AXActivationPoint"
+
+    /// The element's children, or an empty array when it has none.
+    ///
+    /// A round-trip of its own. A walk should take children from ``read(_:)``,
+    /// which fetches them alongside everything else it needs.
+    var children: [AXElement] {
+        guard let value = read(kAXChildrenAttribute), let raw = value as? [AXUIElement] else {
+            return []
+        }
+        return raw.map { AXElement(element: $0) }
+    }
+
+    /// Actions the element accepts, such as `AXPress`.
+    var actions: [String] {
+        var names: CFArray?
+        guard AXUIElementCopyActionNames(element, &names) == .success,
+            let names = names as? [String]
+        else {
+            return []
+        }
+        return names
+    }
+
+    /// Every attribute name the element advertises.
+    func names() -> [String] {
+        var names: CFArray?
+        guard AXUIElementCopyAttributeNames(element, &names) == .success,
+            let names = names as? [String]
+        else {
+            return []
+        }
+        return names
+    }
+
+    /// Read one attribute, or `nil` when the read fails or the value is absent.
+    ///
+    /// Use ``values(_:)`` when reading more than one: this costs a round-trip per
+    /// call, which is what makes a naive tree walk take seconds.
+    func read(_ name: String) -> CFTypeRef? {
+        var value: CFTypeRef?
+        guard AXUIElementCopyAttributeValue(element, name as CFString, &value) == .success
+        else {
+            return nil
+        }
+        guard let value, CFGetTypeID(value) != CFNullGetTypeID() else { return nil }
+        return value
+    }
+
+    /// Read several attributes in one round-trip.
+    ///
+    /// Results are positional and the same count as `names`. An attribute that
+    /// could not be read arrives as CoreFoundation's null or as an `AXValue`
+    /// boxing the error, both of which ``text(_:)`` reports rather than discards.
+    func values(_ names: [String]) -> [CFTypeRef?] {
+        guard !names.isEmpty else { return [] }
+
+        var raw: CFArray?
+        let status = AXUIElementCopyMultipleAttributeValues(
+            element,
+            names as CFArray,
+            AXCopyMultipleAttributeOptions(),
+            &raw
+        )
+
+        guard status == .success,
+            let values = raw as? [CFTypeRef],
+            values.count == names.count
+        else {
+            return Array(repeating: nil, count: names.count)
+        }
+
+        return values
+    }
+
+    /// Whether `name` can be written on this element.
+    ///
+    /// A failed query reports as not settable: the accessibility API answers this
+    /// for every attribute it advertises, so a failure means the element is gone
+    /// or the attribute is not really there.
+    func isSettable(_ name: String) -> Bool {
+        var settable = DarwinBoolean(false)
+        guard AXUIElementIsAttributeSettable(element, name as CFString, &settable) == .success
+        else {
+            return false
+        }
+        return settable.boolValue
+    }
+
+    /// Perform `action`, returning the API's own status.
+    func perform(_ action: String) -> AXError {
+        return AXUIElementPerformAction(element, action as CFString)
+    }
+}
+
+extension AXElement: Element {
+    /// Read the named attributes and the element's children in one round-trip.
+    ///
+    /// Children come back in the same batch as everything else: asking for them
+    /// separately would add a hop per element, and every walk asks for them.
+    func read(_ names: [String]) -> Reading<AXElement> {
+        let values = self.values(names + [kAXChildrenAttribute])
+
+        let children = (values.last.flatMap { $0 } as? [AXUIElement] ?? [])
+            .map { AXElement(element: $0) }
+
+        return Reading(
+            text: values.dropLast().map(Self.optionalText),
+            children: children
+        )
+    }
+
+    /// Read a boolean attribute.
+    ///
+    /// `CFBoolean` bridges to `NSNumber` rather than to `Bool`, so a direct cast
+    /// answers `nil` for a perfectly good `0` or `1`.
+    func flag(_ name: String) -> Bool? {
+        guard let value = read(name) as? NSNumber else { return nil }
+        return value.boolValue
+    }
+
+    /// Write a boolean attribute, answering the API's own status.
+    func setFlag(_ name: String, _ value: Bool) -> AXError {
+        return AXUIElementSetAttributeValue(
+            element,
+            name as CFString,
+            value ? kCFBooleanTrue : kCFBooleanFalse
+        )
+    }
+
+    /// The point held in an attribute, in screen coordinates.
+    func point(_ name: String) -> CGPoint? {
+        guard let value = read(name), CFGetTypeID(value) == AXValueGetTypeID() else {
+            return nil
+        }
+
+        let boxed = unsafeDowncast(value, to: AXValue.self)
+        guard AXValueGetType(boxed) == .cgPoint else { return nil }
+
+        var point = CGPoint.zero
+        guard AXValueGetValue(boxed, .cgPoint, &point) else { return nil }
+
+        return point
+    }
+
+    /// The size held in an attribute, in points.
+    func size(_ name: String) -> CGSize? {
+        guard let value = read(name), CFGetTypeID(value) == AXValueGetTypeID() else {
+            return nil
+        }
+
+        let boxed = unsafeDowncast(value, to: AXValue.self)
+        guard AXValueGetType(boxed) == .cgSize else { return nil }
+
+        var size = CGSize.zero
+        guard AXValueGetValue(boxed, .cgSize, &size) else { return nil }
+
+        return size
+    }
+
+    /// Write a size attribute, answering the API's own status.
+    ///
+    /// The value has to be boxed in an `AXValue`: the API takes `CFTypeRef` and a
+    /// bare `CGSize` is not one, so passing it any other way fails the write with
+    /// no indication of why.
+    func setSize(_ name: String, _ value: CGSize) -> AXError {
+        var size = value
+        guard let boxed = AXValueCreate(.cgSize, &size) else {
+            return .failure
+        }
+
+        return AXUIElementSetAttributeValue(element, name as CFString, boxed)
+    }
+
+    /// Write a string attribute, answering the API's own status.
+    func setText(_ name: String, _ value: String) -> AXError {
+        return AXUIElementSetAttributeValue(element, name as CFString, value as CFString)
+    }
+
+    /// The elements held in an attribute.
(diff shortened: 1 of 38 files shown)


apps/macos/Tools/jpdrive/Sources/DriveKit/AXElement.swift
```diff
@@ -0,0 +1,359 @@
...
+            names as CFArray,
+            AXCopyMultipleAttributeOptions(),
+            &raw
+        )
+
+        guard status == .success,
+            let values = raw as? [CFTypeRef],
+            values.count == names.count
+        else {
+            return Array(repeating: nil, count: names.count)
```
JeanMertz: blocking: A target that exits or is temporarily unresponsive during a walk can make this batch return `invalid_ui_element` or `cannot_complete`. Converting that failure into all-`nil` values makes `tree` produce a valid-looking empty node, or a filtered read report `identifier_not_found`, instead of reporting that the read failed. That is a plausible hidden wrong result while debugging a busy app. Please carry the batch status through `Reading` and turn whole-call failures into `DriveError`; only absent or unsupported individual attributes should become `nil`. The single-value read helpers have the same error-erasure risk.

apps/macos/Tools/jpdrive/Sources/DriveKit/Act.swift
```diff
@@ -0,0 +1,1218 @@
...
+                hint:
+                    "an element with no place to be clicked is usually one that wants `select` "
+                    + "or `press` instead"
+            )
+        }
+
+        // Raised first, because the click lands on whatever is at that coordinate
+        // rather than on the element that named it. A window behind another one
+        // would otherwise have its click swallowed by the window in front.
+        raiseWindow(in: path)
```
JeanMertz: blocking: A normal `jpdrive` call starts with the terminal frontmost and JP in the background. `AXRaise` only changes ordering within JP, as the drag implementation notes, so this click can land in the terminal or another application at the same screen coordinate. That spreads the action outside the requested target. Bring the application frontmost and wait for confirmation before raising and posting, and refuse the click if activation or raising fails.

apps/macos/Tools/jpdrive/Sources/DriveKit/Act.swift
```diff
@@ -0,0 +1,1218 @@
...
+        // ordering between applications follows activation, so raising a
+        // background app's window leaves it under the active app's windows: the
+        // gesture lands on whatever is on top at those coordinates, which is
+        // whatever the person at the keyboard is using. Measured, not assumed — a
+        // drag posted without this was received by the frontmost terminal.
+        //
+        // The cost is that a gesture takes focus. Nothing here can give it back:
+        // this process handles one step and exits, so the restore belongs to
+        // whatever drives the whole list.
+        activate(root, within: activation)
```
JeanMertz: blocking: If macOS refuses activation, or the app does not become frontmost within the timeout, `activate` silently returns and this code still posts a global drag. A modal frontmost app is a plausible trigger, and the gesture then goes to that app rather than JP. Make activation a required, throwing precondition for real pointer gestures. The fake-tree tests can use an injected activator or a root already marked frontmost instead of making production activation best-effort.

apps/macos/Tools/jpdrive/Sources/DriveKit/Act.swift
```diff
@@ -0,0 +1,1218 @@
...
+                kind: .notClickable,
+                message: "\(target.identifier) reports no frame to drag across",
+                hint: "an element with no position or size cannot be aimed at"
+            )
+        }
+
+        let steps = max(target.steps ?? defaultDragSteps, 1)
+        let pause = target.pauseMs.map { Duration.milliseconds($0) } ?? defaultDragPause
+
+        let start = point(target.from, in: origin, size)
```
JeanMertz: blocking: The documented fraction is not validated. A plausible model output such as `dx: 100` (treating it as a percentage) resolves far outside the element's frame, after which the global drag can press or release over the desktop or another window. Validate every `dx` and `dy` as finite and within `0...1` before activating the app or posting any event.

apps/macos/Tools/jpdrive/Sources/DriveKit/Pixels.swift
```diff
@@ -0,0 +1,252 @@
...
+                kind: .notFound,
+                message:
+                    "\(options.axis.rawValue) \(options.at) is outside the image, which is "
+                    + "\(bitmap.width)x\(bitmap.height) pixels",
+                hint: "a row is indexed down from the top and a column across from the left"
+            )
+        }
+
+        let from = max(options.from ?? 0, 0)
+        let to = min(options.to ?? extent - 1, extent - 1)
```
JeanMertz: blocking: `pixels --image <100px-wide.png> --scan row --at 0 --to 150` silently scans only through pixel 99 and returns success. The output is plausible data for a different range, and the caller is not told that its bound moved. This also conflicts with the stated policy of rejecting out-of-image coordinates. Reject `from` or `to` outside `0..<extent` rather than clamping them.

apps/macos/Tools/jpdrive/Sources/DriveKit/Pixels.swift
```diff
@@ -0,0 +1,252 @@
...
+        guard
+            let target,
+            let context = CGContext(
+                data: nil,
+                width: width,
+                height: height,
+                bitsPerComponent: 8,
+                bytesPerRow: width * 4,
+                space: target,
+                bitmapInfo: CGImageAlphaInfo.premultipliedLast.rawValue
```
JeanMertz: blocking: Window captures plausibly contain semitransparent pixels around rounded corners and antialiased edges. This bitmap stores premultiplied RGBA, so a source pixel `#FF000080` is read from `bytes` as roughly `#80000080`; `Pixel.hex` then reports that premultiplied value as an ordinary color. The result is a plausible-looking but wrong color. Unpremultiply RGB before formatting (handling alpha zero), and add a PNG round-trip test with partial alpha rather than testing `Pixel` only after construction.

apps/macos/Tools/jpdrive/Sources/DriveKit/Act.swift
```diff
@@ -0,0 +1,1218 @@
...
+        case type
+        case menu
+        case waitFor = "wait_for"
+        case resize
+        case drag
+    }
+
+    init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
```
JeanMertz: nit: The advertised single-key shape is not enforced here. Direct input such as `{"select": {...}, "press": {...}}` is accepted and silently runs `select` because it is checked first. The downstream harness validates this, so the usual damage is contained and the result names the verb that ran, but `jpdrive act` itself still accepts an ambiguous command. Check that exactly one known key is present before decoding it, and reject extra keys with `bad_usage`.

apps/macos/Tools/jpdrive/Sources/DriveKit/Act.swift
```diff
@@ -0,0 +1,1218 @@
...
+            let size = element.size(kAXSizeAttribute)
+        else {
+            throw DriveError(
+                kind: .notClickable,
+                message: "\(target.identifier) reports no frame to drag across",
+                hint: "an element with no position or size cannot be aimed at"
+            )
+        }
+
+        let steps = max(target.steps ?? defaultDragSteps, 1)
```
JeanMertz: blocking: `steps` comes from model-produced JSON and has no upper bound. A value such as `1_000_000_000` makes the next `map` allocate a billion `CGPoint`s before any action runs, which can exhaust the user's memory and terminate without the promised JSON error document. Reject counts outside a small documented range before constructing the route; clamping only values below one does not bound the dangerous side.

apps/macos/Tools/jpdrive/Sources/DriveKit/Dump.swift
```diff
@@ -0,0 +1,119 @@
...
+
+        let attributes = zip(names, element.values(names)).map { name, value in
+            DumpAttribute(
+                name: name,
+                value: value.map(AXElement.text) ?? "<null>",
+                settable: options.settable ? element.isSettable(name) : nil
+            )
+        }
+
+        let all = depth < options.maxDepth ? element.children : []
```
JeanMertz: blocking: Any ordinary `dump --depth N` reaches this branch with real children at depth `N`, but replacing them with `[]` before computing `elidedChildren` makes the output claim no children were omitted. The truncated node is indistinguishable from a leaf, which is the hidden result this field exists to prevent. Keep `available = element.children`, derive `walked` from it only when below the depth cap, and compute elision from `available.count` as `Tree.node` does. A depth-limit regression test would catch this.

apps/macos/Tools/jpdrive/Sources/DriveKit/WindowIDs.swift
```diff
@@ -0,0 +1,133 @@
...
+        let all = capturable(from: everywhere, pid: pid)
+        let shown = Set(here.map(\.id))
+
+        // The preflight variant, never the requesting one: raising the system's
+        // permission dialog from a background tool leaves a prompt nobody is
+        // watching, in front of the app being measured.
+        return WindowIDReport(
+            screenRecording: CGPreflightScreenCaptureAccess(),
+            windows: here,
+            otherSpaces: all.filter { !shown.contains($0.id) }
```
JeanMertz: blocking: A minimized or hidden JP window is absent from `.optionOnScreenOnly` but present in the unrestricted list, so this reports it under `other_spaces` even when it belongs to the active Space. That gives the caller the opposite diagnosis for a common app state. The window-server set difference only proves “not currently on screen”; report it under that neutral name, or add another source of state before claiming it is on another Space.

apps/macos/Tools/jpdrive/Sources/DriveKit/Act.swift
```diff
@@ -0,0 +1,1218 @@
...
+                    hint: "`under` names a container that must already exist"
+                )
+            }
+            scope = container
+        } else {
+            scope = root
+        }
+
+        let timeout = target.timeoutMs.map { Duration.milliseconds($0) } ?? defaultTimeout
+        let interval = target.intervalMs.map { Duration.milliseconds($0) } ?? defaultInterval
```
JeanMertz: nit: A plausible `wait_for` payload with `interval_ms: 0` (or a negative value) turns this into a tight loop of synchronous accessibility walks for the full timeout. Against a small scope that can hammer the target's main thread, although the slowdown is visible and ends with this command. Reject non-positive intervals and negative timeouts while decoding the step.

apps/macos/Tools/jpdrive/Sources/DriveKit/Act.swift
JeanMertz: Agreed, and this was the strongest of the batch — fixed.

`click` now calls `front(_:within:)` before `raiseWindow`, the same activate-and-wait `menu` already used and `drag` was doing by hand. The normal invocation is from a terminal, so the terminal is frontmost and JP is not; the raise only reordered within JP and the click went to the terminal while the result named the element it aimed at. Reached by every `click` step, escapes the target application, and reports success — the worst combination available.

`ClickTests.activatesTheApplication` pins the activation, and `refusesToClickWithoutActivating` pins that a refused activation stops the step with nothing posted.

apps/macos/Tools/jpdrive/Sources/DriveKit/Act.swift
JeanMertz: Agreed — fixed, together with the `click` comment below, since they were the same bug.

`activate` is gone; `drag` now calls the throwing `front(_:within:)`. Best-effort was the wrong call for exactly the reason the comment three lines above it records: a drag posted without activation was *measured* landing in the frontmost terminal.

On the test justification, you were right to push: `DragTests.dragsWithoutActivating` claimed "a tree that is not a running application has nothing to activate", but `FakeElement.setFlag` writes the flag and `flag` reads it back — `activatesTheApplication` asserts precisely that. It only failed to activate because the test set `writeStatus = .cannotComplete` deliberately. The fake models activation fine, so the stated reason for best-effort did not hold.

That test is now `refusesToDragWithoutActivating` and asserts the opposite: `writeFailed`, and `poster.drags.isEmpty`.

apps/macos/Tools/jpdrive/Sources/DriveKit/Act.swift
JeanMertz: Agreed — fixed. This is the third face of the same bug as the two comments above: a gesture posted into global screen space without establishing where it will land.

`drag` now validates both endpoints before it activates or posts anything. Each of `dx`/`dy` must be finite and within `0...1`, else `bad_usage` naming the offending value. Checking `to` as well as `from` matters: a drag that pressed on the element and released over another application would be worse than one that never started.

The percentage misreading resolves a long way out, which is what makes it dangerous rather than merely wrong — `dx: 100` on an 800pt window aims 80,000 points to the right of it. `DragTests.refusesAnOffsetOutsideTheFrame` covers `100`, `-0.5`, `1.5`, NaN and infinity; `refusesAnOutOfFrameDestination` covers the far endpoint.

apps/macos/Tools/jpdrive/Sources/DriveKit/Act.swift
JeanMertz: Bounded, but clamped rather than rejected — and I'd push back on which harm is the reachable one.

A billion is a construction. The plausible version is `steps: 100000` for "a slow drag", and that is 100k posted events at 8ms each: a thirteen-minute hang, not an allocation failure. Both are contained and visible, so on the severity framework alone this is a suggestion. What makes it worth fixing is the contract: the driver promises one JSON document to stdout either way, and a process killed for memory writes nothing.

Rejecting would contradict the code immediately above it, which clamps the count *up* from zero with a comment explaining why — "a caller computing the count from a distance cannot produce a path with nothing in it". The same reasoning applies in the other direction, so it is now `min(max(steps ?? default, 1), maxDragSteps)` with the cap at 1000 and documented on the constant.

`DragTests.clampsAnEnormousStepCount` pins it.

apps/macos/Tools/jpdrive/Sources/DriveKit/Dump.swift
JeanMertz: Confirmed and fixed. `Tree.node` has the same shape written correctly, which made this a direct comparison: it keeps `available = reading.children` and computes elision from that, while `Dump.node` overwrote with `[]` at the cap, so `all.count > walked.count` was false and the field came back `nil`.

Reached by any `dump --depth N` on a tree deeper than N, and the result is precisely the reading `DumpNode.elidedChildren`'s own doc says it exists to prevent. `Dump.node` now mirrors `Tree.node`.

On the regression test — you're right that one would have caught it, and I'm not adding one here. `Dump.node` takes a concrete `AXElement` and calls `names()` and `values(_:)`, neither of which is on the `Element` protocol, so it cannot be driven by `FakeElement` the way every other traversal in the package is. That is why there is no `DumpTests.swift` at all.

I wrote one, then deleted it: it reimplemented the elision arithmetic inside the test file rather than calling the production code, and a pure `elision(available:walked:)` test would not have caught this bug anyway, since the defect was in *what got passed* to that expression. Coverage that cannot fail for the right reason is worse than none.

Filed as T-05phnfp — bring `Dump` behind the `Element` protocol with a method answering rendered name/value pairs, then add the coverage `Tree` already has. This fix ships verified by reading, which I'd rather say than dress up.

apps/macos/Tools/jpdrive/Sources/DriveKit/Pixels.swift
JeanMertz: Agreed — the arithmetic checks out and it's fixed. `Bitmap.pixel(x:y:)` now divides the colour components back out by alpha before constructing the `Pixel`, guarding the alpha-zero case, which carries no colour to recover.

One note for anyone reading this later: switching the context to a non-premultiplied format is not an option. `CGBitmapContext` accepts only `premultipliedLast`, `premultipliedFirst`, `noneSkipLast` and `noneSkipFirst` for 8-bit RGBA — straight `.last` is not a supported layout — so unpremultiplying on read is the only route.

You were also right about the test. `spellsAlphaOnlyWhenItMatters` constructs a `Pixel` directly and never touches the decode path where the premultiplication happens, so it could not have caught this. `unpremultipliesPartialAlpha` round-trips a real PNG holding a half-transparent red and a fully transparent pixel, and asserts `#FF000080` and `#00000000`.

apps/macos/Tools/jpdrive/Sources/DriveKit/Pixels.swift
JeanMertz: Half of this, with the other half declined — the two bounds are not the same case.

`--from` past the edge is a genuine defect, but not quite the one described: it already errored, just incoherently. With `from: 150` and no `to`, `to` clamps to 99 and the guard reports `--from 150 is past --to 99`, blaming a bound the caller never wrote. Fixed by checking `from` against the image before `to` is clamped, with its own message. `PixelsTests.refusesAStartOutsideTheImage` covers it.

`--to` past the edge I've left clamping. It reads naturally as "to the far edge", and the outcome is contained and visible rather than hidden: the report carries `width`, `height`, and every run's `start` and `count`, so it states exactly which pixels were read. A caller comparing against its own expectation sees the scan stopped at 99.

On the stated policy — that comment (`refusesALineOutside`) is about `at`, which selects *which* line to scan and has no sensible clamped reading. `from`/`to` bound a range along it, where clamping to the edge is the ordinary meaning. Worth documenting rather than changing, which I've done on `PixelOptions.to`.

apps/macos/Tools/jpdrive/Sources/DriveKit/WindowIDs.swift
JeanMertz: Agreed — fixed by renaming rather than by adding state.

You're right that the set difference proves only "not currently on screen", and minimizing is one keystroke. The field's own doc made the stronger claim ("not on the active Space"), so the report gave a confident wrong diagnosis for a common state, and the two conditions have different remedies — switch Space versus deminiaturize.

`CGWindowListCopyWindowInfo` carries no minimized flag, so distinguishing them from this source alone isn't possible; the second option in your comment would need a different source entirely. The field is now `offScreen` / `off_screen`, and its doc claims only what the difference supports, naming minimized, hidden and other-Space as the three that land there and pointing at `jpdrive windows`, which reads `AXMinimized` from the accessibility tree and separates the first of them.

No consumer to update — I checked, `other_spaces` is unreferenced in the tools that shell out to the driver.

apps/macos/Tools/jpdrive/Sources/DriveKit/AXElement.swift
JeanMertz: Real, and filed as T-05phhc9 rather than fixed here — but I'd amend the remedy.

The erasure is exactly as you describe: an all-`nil` return is indistinguishable from an element that genuinely reports none of the requested attributes, and `cannotComplete` from a busy target is reachable while driving an app that is doing work.

Where I'd diverge is "turn whole-call failures into `DriveError`" as a blanket rule, because it's wrong for two of the three consumers. `tree` and `dump` are diagnostics; aborting an entire walk because one element timed out mid-traversal is worse than reporting the gap. `find` is the one where an error is right, and it's also where the sharp harm is — a step reporting `identifier_not_found` for an element that exists, after which a script author "fixes" a correct identifier.

So the shape in the ticket is: `Reading` gains a `failed` flag distinct from the per-attribute nulls; `Tree` and `Dump` keep emitting the node with an explicit unreadable marker; `find` tracks whether any read failed and, when the identifier was not found *and* something failed, says so instead of reporting a clean miss.

On the single-value helpers — I looked, and I'd leave them. `point`, `size` and `flag` each have callers that correctly treat absence as a refusal (`click` on an element with no activation point, `drag` on one with no frame), so there is no erasure to undo there.

Out of this PR because it changes the `Element` protocol every file in the package uses, it's independent of the six fixes here, and it wants its own tests.

apps/macos/Tools/jpdrive/Sources/DriveKit/Act.swift
JeanMertz: Correct about the decoder, but I'm leaving it — and your own framing is most of the reason.

Contained and visible: `StepResult.step` names the verb that ran, so the output states exactly what happened rather than hiding it. And I can't name an input that produces it. The JSON is built by `debug_app_drive` constructing one verb; a model writing a step writes a step. A two-key object is a construction, not something a caller plausibly emits.

Weighed against that, enforcing it costs a permanent branch plus a test for a case that already reports itself. The worse result a reader gets by my declining: a hand-written two-key step runs its first verb and tells them which one.

Happy to be overruled if you've actually seen one produced.

apps/macos/Tools/jpdrive/Sources/DriveKit/Act.swift
JeanMertz: Accurate, and leaving it — contained and visible, as the comment itself grants.

The part that decides it for me: this failure mode already announces itself. The timeout error reports `(\(attempts) attempts over \(elapsed.milliseconds)ms)`, so a zero interval surfaces as an absurd attempt count in the very message the caller reads. Bounded by the timeout, ends with the process, nothing outlives the command.

Worse result from declining: a `wait_for` with a zero interval spins for up to its timeout and then tells you it made forty thousand attempts — which is both the symptom and the diagnosis.

Reasonable to revisit if a real run ever hits it, but I'd rather not add a validation branch for a case the output already explains.

apps/macos/Tools/jpdrive/Sources/DriveKit/AXElement.swift
JeanMertz: Correction to my previous reply: this is fixed in this PR, and the ticket is deleted rather than filed.

Deferring it was the wrong call and the reasoning was wrong too. I scoped it as a protocol upheaval because it touched `Element`; counted properly it was about seventy lines. And a ticket is for a pre-existing problem or a follow-up improvement — `AXElement.values` is new in this PR, so there was no version of "we didn't introduce it" available. Shipping new code with a known hidden-failure path and a note to fix it later is not a trade worth making.

What landed is the shape I described:

- `values(_:)` returns `nil` when the call fails as a whole, rather than a row of `nil`s that reads identically to an element reporting nothing.
- `Reading` carries `failed`, distinct from the per-attribute nulls.
- `Act.find` counts unreadable elements and, on a miss *with* a gap, throws a new `read_failed` naming the count instead of `identifier_not_found`.
- `Tree` and `Dump` keep emitting the node with `unreadable: true`, since aborting a diagnostic walk over one transient failure is worse than reporting the gap.

`ActTests` pins both directions: `distinguishesAGapFromAMiss` and `aReadableTreeReportsACleanMiss`, so the new branch cannot be firing on everything. `TreeTests.marksAnUnreadableElement` and `DumpTests.marksAnUnreadableElement` cover the walks.

Single-value helpers unchanged, for the reason given before.

apps/macos/Tools/jpdrive/Sources/DriveKit/Dump.swift
JeanMertz: Correction to my previous reply: there is a regression test now, and the ticket is deleted rather than filed.

You asked for one and I explained at length why it was impractical. The explanation was accurate — `Dump.node` took a concrete `AXElement` and called `names()`/`values(_:)`, neither on the `Element` protocol — but it described a fixable obstacle as a reason not to test. A fix that ships unverified is not finished, and this is new code, so there was nothing to inherit the excuse from.

`Element` gained `reportedAttributes(settable:)`, returning rendered name/value pairs so the CoreFoundation handling stays inside `AXElement`. `DumpAttribute` became `Attribute` on the protocol, and `Dump.node` is generic over `Element`, so `FakeElement` drives it like every other traversal in the package.

`DumpTests.swift` now covers the depth limit reporting its elision — the case that was broken — plus the sibling cap, a genuine leaf reporting nothing, structural attributes being dropped, settability being opt-in, and an unreadable element being marked rather than shown empty.

That also retires the note I made earlier about `Dump` being the one file in the package with no suite.

PR #1010, opened 2026-08-19
feat(macos): Add the UI test suite

JeanMertz:
The app's unit tests run inside the app process, which puts three things
out of reach: whether a menu item is enabled, what landed on the
pasteboard, and what survives a terminate and relaunch. Those are
exactly the behaviours that break without anyone noticing, and `QA.md`
carried them as a manual checklist. This is the half of that checklist a
machine can run.

A UI test drives the app through its accessibility tree from a separate
process, so `@testable import JP` is unavailable and must not be reached
for. Anything checkable in-process belongs in `JPTests`, where it costs
milliseconds instead of an app launch.

Three things make the suite fast enough to be worth having. A suite
shares one launched app, because launching costs about four seconds and
the work under test costs milliseconds — so a test leaves the app as it
found it, or the suite is ordered so that what one test leaves is what
the next expects. Waits are on conditions rather than the clock, and not
through `waitForExistence`, which reports an element about a second
after it appears whatever the element; asking again finds it in under
100ms, and the timeout becomes the price of a failure rather than of a
pass. And every animation goes through one lever the tests turn off,
because XCUITest waits for the app to stop moving before each action it
synthesizes.

Anything the tests need the app to do differently goes through
`DebugState`, gated on `#if DEBUG` and off unless an environment
variable says otherwise, so a release build has no way to reach it. The
pasteboard works this way and is enforced: no test may touch the system
pasteboard, a debug build copies wherever `JP_DEBUG_PASTEBOARD` names,
and `ClipboardPolicyTests` fails on any spelling of the general one.

`just test-app-ui` is the CI job and runs everything even after a
failure. While writing a test, run it by name instead. `fmt-app` and
`lint-app` arrive here rather than with the app because they name every
Swift directory in the repository, and this is the last one to exist.


apps/macos/UITests/PointerCursorTests.swift
```diff
@@ -0,0 +1,84 @@
...
+            // showing by anything at all would pass the assertion below without
+            // this app having done a thing. Establishing the arrow first turns "it
+            // is the right cursor" into "it changed to the right cursor".
+            //
+            // The conversation list rather than the transcript: the transcript does
+            // not exist until something is selected, and hovering a missing element
+            // fails without stopping the test — which is how an earlier version of
+            // this passed with no baseline at all.
+            driven.sidebar.hover()
+            let arrowFirst = driven.waitForCursor(.arrow)
```
JeanMertz: blocking: The ordinary input here is the system arrow after hovering the sidebar. `AppUnderTest.cursorIs` explicitly documents that `NSCursor.arrow.image` does not match the bytes returned by `NSCursor.currentSystem`, so this wait times out and returns before the divider check. Establish the baseline by waiting until the cursor is not `.columnResize` rather than trying to match `.arrow`.

apps/macos/UITests/PointerCursorTests.swift
```diff
@@ -0,0 +1,84 @@
...
+                """
+            )
+            guard arrowFirst else { return }
+
+            driven.divider.hover()
+
+            // Read into a `Bool` first: swift-testing reports the expression it
+            // evaluated, and `driven` holds an `XCUIApplication` whose description
+            // is the entire element tree.
+            let changed = driven.waitForCursor(.columnResize)
```
JeanMertz: blocking: A normal `just test-app-ui` run against the current app reaches the behavior documented on this test: the divider leaves the pointer unchanged. That makes `changed` false on every run, so the new test target stays red and masks future UI regressions behind this known failure. Land this test with the cursor fix, or disable it as a known issue until that behavior is implemented.

apps/macos/UITests/ConversationListTests.swift
```diff
@@ -0,0 +1,242 @@
...
+                    Transcripts.configPipeline, "the Config pipeline transcript")
+            else { return }
+
+            driven.app.typeKey(.escape, modifierFlags: [])
+            guard
+                driven.expectDisappears(driven.transcript, "the transcript, after Escape")
+            else { return }
+
+            let edit = driven.openMenu("Edit")
+            #expect(driven.menuItemIsEnabled("Copy Link", in: edit) == false)
```
JeanMertz: blocking: A plausible regression is that `Copy Link` is omitted while there is no selection and inserted once a row is selected. `menuItemIsEnabled` returns `false` for both a missing item and a disabled item, so that regression passes this whole test after the selected-path copy succeeds. Assert that the item exists first, then assert that it is disabled, so the test checks the menu contract rather than conflating absence with disablement.

apps/macos/UITests/ConversationListTests.swift
```diff
@@ -0,0 +1,242 @@
...
+            let hidden = driven.row(ConversationFixtures.readingList)
+            guard
+                driven.expectAppears(hidden, "the Reading list row"),
+                // Present whether or not there is anything to clear, so it is
+                // there to be found before a word has been typed.
+                driven.expectAppears(driven.filterClear, "the clear button")
+            else { return }
+
+            driven.filter.click()
+            driven.filter.typeText("Release")
```
JeanMertz: blocking: A filtering regression where typing `Release` fails to hide the Reading list row takes the guard below and leaves the shared app filtered. Because CI continues with the rest of this suite, later tests then fail while looking for rows this test hid, so one defect spreads into unrelated failures. Install cleanup immediately after mutating the field, preferably clearing it through keyboard input so cleanup does not depend on the clear button behavior this test also checks.

apps/macos/UITests/AppUnderTest.swift
```diff
@@ -0,0 +1,566 @@
...
+        if let failure = Quiescence.installation {
+            let message = "the quiescence waits could not be turned off: \(failure)"
+            Diagnostics.append("\(sourceLocation.fileName):\(sourceLocation.line): \(message)")
+            Issue.record("\(message)", sourceLocation: sourceLocation)
+        }
+
+        let app = XCUIApplication()
+        app.launchEnvironment = fixture.environment
+        if !keepingWindowState {
+            app.launchArguments = ["-ApplePersistenceIgnoreState", "YES"]
```
JeanMertz: blocking: Running this suite from a Mac whose preferred language is not English localizes system-owned menu titles such as `Edit`, `View`, `Window`, `Enter Full Screen`, and `Merge All Windows`, while the tests query their English titles. That makes otherwise-correct builds fail across the menu tests. Pin `AppleLanguages` and the locale in every launch, then append `-ApplePersistenceIgnoreState` conditionally.

apps/macos/UITests/ConversationListTests.swift
```diff
@@ -0,0 +1,242 @@
...
+
+        /// The URI lands on a pasteboard of the fixture's own, never the system
+        /// one — the app under test is told which to use, and
+        /// `ClipboardPolicyTests` holds the suite to it.
+        @Test("copies a conversation's URI from the context menu")
+        func contextMenuCopiesTheURI() {
+            driven.row(ConversationFixtures.readingList).rightClick()
+            driven.chooseContextMenuItem("Copy Link")
+
+            #expect(fixture.copiedText() == ConversationFixtures.readingList.uri)
```
JeanMertz: nit: This reads the pasteboard immediately after a synthesized menu click, although `AppUnderTest` documents that app work can occur after `click()` returns because quiescence waits are disabled. A normal run can therefore observe `nil` or the previous URI and fail even though Copy Link completes a moment later. Poll for the expected pasteboard value up to `AppUnderTest.timeout`, as the element and transcript assertions do.

apps/macos/UITests/Quiescence.swift
```diff
@@ -0,0 +1,96 @@
...
+            return "XCTest no longer has a class named \(className)."
+        }
+
+        let two: @convention(block) (AnyObject, Bool, Bool) -> Void = { _, _, _ in }
+        let three: @convention(block) (AnyObject, Bool, Bool, Bool) -> Void = { _, _, _, _ in }
+
+        let replacements = [
+            (
+                name: "waitForQuiescenceIncludingAnimationsIdle:isPreEvent:",
+                encoding: "v24@0:8B16B20",
```
JeanMertz: question: Is the UI suite intentionally arm64-only? On an Intel Mac, Objective-C `BOOL` uses the `c` encoding that the comment above calls out, so this exact `B` check fails and every `AppUnderTest.launch` records an issue before testing app behavior. If Intel source builds remain supported, this needs an architecture-correct replacement or the optimization should fall back without failing correctness tests; otherwise the target should declare its architecture restriction.

justfile
```diff
@@ -344,6 +344,36 @@ test-app: gen-app (build-ffi "debug")
...
+# Takes over the screen for the length of the run. This is the CI job; while
+# writing a test, run it by name through the `swift_test_ui` tool instead, which
+# stops at the first failure.
+#
+# Every test runs here even after one fails, which is what `CI` means to that
+# tool and what a run nobody is watching should do.
+[group('test')]
+[macos]
+test-app-ui: gen-app (build-ffi "debug")
+    CI=1 xcodebuild test -project apps/macos/JP.xcodeproj -scheme JP \
```
JeanMertz: nit: When a UI assertion fails, this direct `xcodebuild` path bypasses the `swift_test_ui` collection that `Diagnostics` relies on. Failure messages and PNGs remain in the xctrunner container, while reported `tmp/uitests/...` paths do not exist in the checkout, so the CI job loses the evidence these helpers captured. Route the all-tests run through equivalent collection logic, or give `xcodebuild` an explicit result bundle and copy the diagnostics out as CI artifacts.

PR #1011, opened 2026-08-19
feat(tools): Add check, format and test tools for the macOS app

JeanMertz:
The Rust side of this repository can be built, linted, formatted and
tested from inside a conversation. The Swift side could not, which made
the app the one place where a change had to be handed back to a human to
find out whether it even compiled. These close that gap: each shells out
to a toolchain binary from the repository root and reports diagnostics
rather than raw build logs, the way the `cargo_*` tools do.

Every tool that builds brings its own inputs up to date first — the
`jp_ffi` static library, its generated header, and the Xcode project —
so there is no setup step to forget and no failure mode where the tools
disagree with what `just` would have produced.

`xcodebuild` repeats the failing command line in full for every error,
so a broken build's tail is almost entirely noise; diagnostics are
capped and the head is what survives. `swift_check` builds the app,
`swift_format` formats or reports without rewriting, and `swift_test`
runs the unit tests and the `jpdrive` package.

`swift_test_ui` is separate and requires test names. A UI test takes
over the screen for as long as it runs, so the whole bundle is CI's job
rather than something a conversation should trigger by accident; asking
for the suite by name is the guard. It stops at the first failure and
closes the app the run left behind, unless `CI=1` says to finish and
report everything.

Stopping early needs the process interrupted rather than killed, which
`run_until` on the shared runner now supports. `SIGINT` gives
`xcodebuild` the chance to tear down its test session, and tearing down
that session is what stops the app the test was driving; killed
outright, it leaves that app sitting on the screen. Five seconds to
unwind, then a kill.

A failure copies the screenshots the run recorded into `tmp/uitests/`,
because a UI test that failed on something visual is unreadable from its
assertion alone.


.config/jp/tools/src/swift/test_ui.rs
```diff
@@ -0,0 +1,242 @@
...
+/// Stopped from out here because nothing inside can do it. swift-testing has no
+/// cancellation, and a test that exits its own process makes `xcodebuild`
+/// relaunch the runner, finish the remaining tests, and report the whole thing
+/// as passing.
+fn run<R: ProcessRunner>(
+    ctx: &Context,
+    tests: &[String],
+    runner: &R,
+) -> Result<(ProcessOutput, Stopped), std::io::Error> {
+    let filters: Vec<String> = tests.iter().map(|test| ui_bundle_filter(test)).collect();
```
JeanMertz: blocking: A call with one valid selector and one typo, for example `tests = ["UISuite/ConversationListTests/clickSelects()", "UISuite/ConversationListTests/clickSelcts()"]`, still produces a non-empty summary for the valid test. `xcodebuild` ignores the selector that matched nothing, and `outcome` only catches the case where the entire run has no summary, so this returns success without running everything requested. Validate every requested selector against enumeration, or compare the successful result bundle's executed test IDs with the requested selectors, and add a mixed valid/invalid regression test.

.config/jp/tools/src/swift/test_ui_tests.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/// CI has nobody waiting on it and wants every result from the one run it gets,
+/// so it opts out of stopping.
+#[test]
+fn ci_is_read_from_the_environment() {
+    // SAFETY: this test reads back only what it just wrote, and the tools crate
+    // runs its tests in one process where nothing else touches `CI`.
+    unsafe {
+        std::env::set_var("CI", "1");
```
JeanMertz: blocking: Running `cargo test -p tools` uses parallel test threads. Other tests in this file enter `swift_test_ui_impl` and read `CI` through `under_ci` while this test mutates the process environment. On Unix, `std::env::set_var`'s safety contract forbids concurrent environment readers, so this can invoke undefined behavior; the test also removes a pre-existing `CI` value at the end. Extract the value interpretation into a pure helper that accepts `Option<&str>` and test that without changing the process environment.

.config/jp/tools/src/swift/test_ui.rs
```diff
@@ -0,0 +1,242 @@
...
+        let mut message = "`tests` is required: name the suites or tests to run. Every one of \
+                           these launches the app and drives it through the screen, so there is \
+                           deliberately no way to ask for all of them; that is CI's job, through \
+                           `just test-app-ui`."
+            .to_owned();
+
+        // Asked of the bundle rather than read off a list someone maintains:
+        // a list in a document rots, and the one place that cannot is the
+        // bundle itself.
+        match enumerate(ctx, runner) {
```
JeanMertz: nit: On a fresh checkout, asking with no test names cannot enumerate the bundle here: `JP.xcodeproj` is uncommitted and generated by `xcodegen`, and the bridging header must exist before Xcode plans the build. This branch reaches `enumerate` before `prepare`, so the documented empty-list discovery path falls back to the generic naming example until some earlier command happened to prepare the app. Run the required generation steps before enumeration and cover the absent-project/header case.

.config/jp/tools/src/swift.rs
```diff
@@ -0,0 +1,160 @@
...
+            "Building `jp_ffi` failed:\n\n```\n{}\n```",
+            report(&ffi, "just build-ffi")
+        )));
+    }
+
+    let project = runner.run(
+        "xcodegen",
+        &["generate", "--spec", PROJECT_SPEC, "--project", PROJECT_DIR],
+        &ctx.root,
+    )?;
```
JeanMertz: nit: On a machine where `xcodegen` is not installed, spawning it returns `io::ErrorKind::NotFound` at this `?`, so execution never reaches the Homebrew hint below. The mock's `returns_error` path models a process that started and exited non-zero, not an executable that could not be spawned. Catch `NotFound` here and return the hinted failure outcome, with a runner test that returns a spawn error.

.config/jp/tools/src/swift/test_ui.rs
JeanMertz: Correcting my previous reply: it is implemented, and my reason for deferring it was wrong.

I said the shape of the `xcresulttool` JSON could not be read off the source and needed a real bundle. It is documented — `xcrun xcresulttool get test-results tests --help` embeds the JSON Schema of its own output. `TestNodeType` is a closed enum (`Test Plan`, `Unit test bundle`, `UI test bundle`, `Test Suite`, `Test Case`, …), and a `Test Case` node carries `nodeIdentifier` in the form `FooTests/fooTest()` — the suite path and function without the bundle prefix. That is exactly the selector this tool takes, since it adds `JPUITests/` itself, so the two compare with no reshaping. I declared it unknowable without checking, which was the wrong call.

Taking your second option, for the reason I gave: `bundle_document` now fetches the document once and both `bundle_issues` and the new `executed_tests` read it, so the check costs no extra `xcresulttool` call on any run. Validating against `enumerate` up front would have put a build on the hot path of a six-second loop.

- `executed_tests` walks for `Test Case` nodes and collects `nodeIdentifier`.
- `unmatched` reports requested names nothing answers to — equality for a whole test, a `/` boundary for a suite standing for what is under it, so `UISuite/ConversationList` does not match `UISuite/ConversationListTests`.
- A successful run with an unmatched name is an error naming them, with the summary of what did run beneath it.

The regression test you asked for is `names_the_requested_tests_that_never_ran`, with your exact example — `clickSelects()` alongside `clickSelcts()`. Four more cover suite matching, the partial-name case, and the unreadable-bundle case.

Two conservatisms worth flagging, both chosen so a shape surprise degrades to the old behavior rather than failing a passing suite: `nodeIdentifier` is optional in the schema, so a `Test Case` lacking one is skipped rather than counted as missing; and an empty executed list is read as "cannot tell", which skips the check.

One gap I will not paper over: the six lines wiring the check into `swift_test_ui_impl` have no test. `clear_result_bundle` removes the bundle directory before the run and the mock runner cannot recreate it, so no end-to-end test can have a bundle present — the same reason `collect_bundle_issues` has never been covered there. The walk, the matcher and the fetch are each tested; their composition is not.

The docs promise the guarantee again, since it is true again. **T-06zzq1w** is closed.

PR #1012, opened 2026-08-19
feat(tools): Add tools for driving and profiling the macOS app

JeanMertz:
Reading a GUI from a conversation is a different problem from reading a
CLI. A CLI hands back its output; an app puts pixels on a screen nobody
in the conversation can see. Guessing at what the app is doing, or
handing every question back to a human, is not a development loop.

These tools launch the app, drive it through its accessibility tree,
read back what is on screen, capture pixels, take screenshots, record
traces, and profile it. `debug_app_quit` stops it again, closing any
profile bracket that is still open and keeping the app's state so a
relaunch can test what it restores. Together they answer "what is the
app actually doing" without a human in the middle.

Profiling goes through `xct2cli`, which turns an Instruments `.trace`
bundle into symbolicated hotspots and a callgraph. Driving goes through
`jpdrive`, so the accessibility work stays in Swift where the API lives
and this side only shells out to it. `jpdrive` gains `--actions` and
`--menus` here, because there was no way to ask it for less: it read
every kept element's actions and walked the whole menu bar regardless,
and this side filtered the result, paying for a few hundred
accessibility round-trips to throw them away. `just test-drive` also
stops narrating every passing test, keeping the failures and the closing
count.

Two pieces move out of `debug_jp` into `util` because both halves need
them. The trace parser becomes `util::trace`: the same `tracing` output
is read whether the process under inspection is `jp` or the app, and one
parser reading one format is the point. Path shortening becomes
`util::paths`, and `debug_jp` switches from relativizing each value
before it goes into a report to shortening the finished report in one
pass. These reports quote a subprocess's stderr verbatim, render dhat
frames carrying source locations, and print trace fields naming whatever
was being read — there is no enumerating where an absolute path can turn
up, so enumerating was the wrong shape. One pass over the text also
covers the artifact paths in the footer, which is why nothing upstream
relativizes anything any more.


.config/jp/tools/src/debug_app/drive.rs
```diff
@@ -0,0 +1,515 @@
...
+    let run = marks::new_run();
+    let mut marked = Vec::new();
+
+    for (index, step) in steps.iter().enumerate() {
+        let began_ms = marks::now_ms();
+        let acted = act(&bin, session.pid, step, root, runner)?;
+        let after = if reads.is_none() {
+            String::new()
+        } else {
+            reading(&bin, session.pid, opts, root, runner)?
```
JeanMertz: blocking: A `menu`, `click`, or `drag` can move focus and the pointer successfully, then hit a failed post-step tree read here if the app exits or the accessibility read fails. This `?` returns before `ambient::restore`, so the failure leaves the user's desktop in the borrowed state. Put the restore behind a guard, or collect the loop result and restore before propagating it, so every exit path returns the ambient state.

.config/jp/tools/src/debug_app/capture.rs
```diff
@@ -0,0 +1,915 @@
...
+    }
+
+    /// Whether this bracket is still open.
+    ///
+    /// A stop stamp or a summary means it already closed, whatever else is on
+    /// disk.
+    /// Liveness of the recorder says only whether stopping it needs a signal,
+    /// not whether there is anything to stop for.
+    pub(crate) fn is_pending(&self, dir: &Utf8Path) -> bool {
+        self.stopped_unix.is_none() && !self.summary(dir).exists() && self.age() < PENDING_WINDOW
```
JeanMertz: blocking: A bracket left open for more than 48 hours while `xctrace` is still alive becomes non-pending solely because of its age. The next `start` or `quit` sweep then removes its sidecar and bundle without signalling the recorder, so that recorder keeps running and writing with no record through which it can be stopped. An open record needs to remain addressable while its recorded process identity is still valid, or expiry must explicitly stop and reap it before deleting its artifacts.

.config/jp/tools/src/debug_app/capture.rs
```diff
@@ -0,0 +1,915 @@
...
+
+    let mut out = Vec::new();
+    for entry in entries.flatten() {
+        let path = Utf8PathBuf::from_path_buf(entry.path()).unwrap_or_default();
+        if path.extension() != Some("json") {
+            continue;
+        }
+
+        if let Ok(raw) = fs::read_to_string(&path)
+            && let Ok(recording) = serde_json::from_str::<Recording>(&raw)
```
JeanMertz: blocking: An interrupted sidecar write can leave an empty or truncated `.json` beside a live or finalized trace. This silently drops that recording, while `orphaned_bundles` treats the mere presence of the `.json` as proof that the bundle is owned. The recorder and bundle then escape both `pending()` and every sweep, which can leave an `xctrace` process and a system-wide bundle containing process environments behind indefinitely. Write sidecars atomically and surface or quarantine malformed records so their bundles remain reclaimable.

.config/jp/tools/src/debug_app/session.rs
```diff
@@ -0,0 +1,442 @@
...
+
+/// Whether a process with `pid` is currently alive.
+///
+/// On non-unix targets, conservatively returns `true`: these tools are macOS
+/// only, and a wrong `false` would report a running app as gone.
+#[cfg(unix)]
+pub(crate) fn pid_is_alive(pid: u32) -> bool {
+    // `kill(pid, 0)` runs the kernel's permission and existence checks without
+    // sending a signal: 0 => alive; EPERM => alive but not ours; ESRCH => gone.
+    if unsafe { libc::kill(pid.cast_signed(), 0) } == 0 {
```
JeanMertz: blocking: A PID is not a stable process identity. If the app or recorder exits, its stale pid file or sidecar remains, and macOS later reuses that number for another process, this check reports the unrelated process as alive. `debug_app_quit` can then send it `SIGTERM` and `SIGKILL`, and profile stop can send it `SIGINT`. Persist a birth token such as process start time plus executable identity and verify it before accepting a session or sending any signal.

.config/jp/tools/src/debug_app/capture.rs
```diff
@@ -0,0 +1,915 @@
...
+
+    pub configuration: String,
+}
+
+impl Target {
+    /// What a running session says about the app a bracket is recording.
+    pub(crate) fn for_session(session: &Session) -> Target {
+        Target {
+            pid: session.pid,
+            binary: app_binary(&session.bundle),
```
JeanMertz: blocking: A retained recording stores paths to the slot's staged `JP.app` and the DerivedData dSYM, but both paths are overwritten by later builds and launches. A normal sequence of recording build A, rebuilding, then asking `mode: "report"` about build A will symbolicate its PCs against build B and can return plausible but incorrect function names. Preserve the matching binary and dSYM with the recording, or record their UUIDs and refuse the read when the files at these paths no longer match.

.config/jp/tools/src/debug_app/marks.rs
```diff
@@ -0,0 +1,190 @@
...
+}
+
+impl Mark {
+    /// Whether `at_ms` falls inside this step's window.
+    ///
+    /// Called with the moment an interval began.
+    /// See the module documentation for why that rather than the moment it
+    /// ended.
+    pub(crate) const fn holds(&self, at_ms: u64) -> bool {
+        at_ms >= self.began_ms && at_ms <= self.ended_ms
```
JeanMertz: blocking: Adjacent steps can share the same millisecond boundary because one mark's `ended_ms` is followed immediately by the next mark's `began_ms`. With both ends inclusive, an interval whose truncated millisecond timestamp equals that boundary satisfies both marks and is counted under both steps. That silently inflates the deterministic counts this feature relies on. Make attribution non-overlapping, for example with half-open windows or by assigning each interval to exactly one mark.

.config/jp/tools/src/debug_app/report.rs
```diff
@@ -0,0 +1,1473 @@
...
+/// The marks belonging to the run a scope selects.
+///
+/// A named recording picks the run it overlaps; otherwise the most recent run,
+/// which is what a caller asking about "the drive I just did" means.
+fn run_marks(all_marks: &[Mark], named: Option<&Recording>, window: Window) -> Vec<Mark> {
+    let Some(_) = named else {
+        return marks::latest_run(all_marks);
+    };
+
+    marks::overlapping(all_marks, window.from_ms, window.to_ms)
```
JeanMertz: blocking: A recording can span two ordinary `debug_app_drive` calls. This returns every overlapping mark from both runs, but `render_timeline` labels the result with only `steps[0].run` and treats duplicate step numbers as one run. A scoped report can therefore attribute and compare counts under the wrong run without saying so. Group overlapping marks by `run` and either render each run separately or require the caller to choose one.

.config/jp/tools/src/debug_app/report.rs
```diff
@@ -0,0 +1,1473 @@
...
+    );
+
+    out.push_str("| # | Step | View bodies | Δ | FFI calls | Δ |\n");
+    out.push_str("| -: | :--- | ---: | ---: | ---: | ---: |\n");
+
+    for mark in steps.iter().take(MAX_ROWS) {
+        let here = counts_for(selected, mark);
+        let there = against_steps
+            .iter()
+            .find(|other| other.step == mark.step)
```
JeanMertz: blocking: Two recordings with the same number of steps but different step lists are silently compared by position here. For example, a three-selection run compared against a three-resize run produces numeric deltas as if each pair measured the same action, and the warning below does not fire because the counts match. Verify the labels as well as the positions, and refuse or mark mismatched rows rather than presenting their deltas as comparable.

.config/jp/tools/src/debug_app/report.rs
```diff
@@ -0,0 +1,1473 @@
...
+    let intervals = stream::load(dir);
+    let sampled: Vec<&Interval> = intervals
+        .iter()
+        .filter(|interval| window.holds(interval.started_ms))
+        .filter(|interval| interval.footprint_mb.is_some())
+        .collect();
+
+    let mut out = String::new();
+
+    match (sampled.first(), sampled.last()) {
```
JeanMertz: blocking: `stream::load` sorts intervals by `started_ms`, but footprint samples are taken at `at_ms`, when an interval ends. With a normal outer `conversation.select` containing shorter nested intervals, `first()` can be the last sample taken and `last()` can be an earlier one, reversing the reported before/after trajectory and the retained-versus-transient conclusion. Select the earliest and latest samples by `at_ms`, as `stream::count` already does for its last footprint.

.config/jp/tools/src/debug_app/quit.rs
```diff
@@ -0,0 +1,342 @@
...
+        // recording.
+        let target = Target::for_session(session);
+        drop(recording.close(Some(target.clone()), dir));
+
+        // A system-wide bundle goes whatever the read did: it is credential
+        // material, and a failure that leaves one behind is the case nobody is
+        // watching.
+        let read = hotspots::summarize(&recording, dir, Some(&target), shortenings)
+            .map_err(|e| e.to_string());
+        drop(recording.retire(dir));
```
JeanMertz: blocking: A filesystem failure while storing the close stamp or deleting the trace is discarded here. In the system-wide case, `remove_dir_all` can fail because of permissions or an I/O error, yet quit still reports the summary and removes the session while the bundle containing every recorded process environment remains on disk. Propagate these cleanup failures, or include them in `Closed` and keep enough state to retry, rather than dropping both `close` and `retire` results.

.config/jp/tools/src/debug_app/quit.rs
```diff
@@ -0,0 +1,342 @@
...
+            &recording,
+            dir,
+            &session,
+            signals,
+            capture::FINALIZE_TIMEOUT,
+            &shortenings,
+        )
+    });
+
+    let termination = if session.is_running() {
```
JeanMertz: blocking: If `xctrace` is still finalizing after the five-minute timeout, `close` returns `Stop::Stuck`, but execution continues here and terminates the app anyway. That leaves the recorder running against a disappearing target and then removes the session needed to attribute or clean up the bracket, despite the comment above saying this state has undefined behavior. Return an error before stopping the app when the bracket is stuck, and keep the session record for a later retry.

.config/jp/tools/src/debug_app/profile.rs
```diff
@@ -0,0 +1,548 @@
...
+
+    let (id, started_unix) = new_id();
+    let recording = Recording {
+        id,
+        tiers: tiers.to_vec(),
+        scope,
+        recorder_pid: 0,
+        started_unix,
+        stopped_unix: None,
+        target: None,
```
JeanMertz: blocking: An attached bracket already knows its target here, but the sidecar records `None` until `stop`. If that app crashes and the user launches another app in the slot before stopping the bracket, `stop` reads the replacement session and attributes the old trace to the new PID, binary, slide, and dSYM. Store `Target::for_session` at start for `Scope::Attach`; only a system-wide bracket needs its target filled from a later launch.

.config/jp/tools/src/debug_app/tree.rs
```diff
@@ -0,0 +1,220 @@
...
+    if let Some(depth) = opts.depth {
+        args.push("--depth".to_owned());
+        args.push(depth.to_string());
+    }
+
+    if opts.frames {
+        args.push("--frames".to_owned());
+    }
+
+    args
```
JeanMertz: nit: A normal snapshot with `actions: false` and `menus: false` still pays for both. `jpdrive` currently reads `element.actions` for every retained node and walks the menu-bar subtree; these options are applied only while rendering the returned JSON, and neither is represented in this argv. On a whole-app read that means hundreds or thousands of extra Accessibility calls whose output is discarded. Pass fetch controls through the driver protocol so the default read avoids that work.

.jp/config/skill/debug-app.toml
```diff
@@ -0,0 +1,203 @@
...
+behind by an interrupted bracket is reclaimed without being asked, and retention is bounded by an \
+age window and a byte budget with the oldest evicted first. Never commit a bundle or attach one to \
+a bug report.
+
+The app's own stream is archived rather than truncated at each launch, so the per-step counts from \
+an earlier run stay readable. That is what `against` compares against, and it is worth more than \
+the bundles: those counts live there and nowhere else.
+
+Everything a run touches is scoped to a *slot*, named by `JP_DEBUG_APP_SLOT` and defaulting to \
+`default`. A slot owns its session record, its state and conversation-store directories, its \
```
JeanMertz: nit: In a normal tool invocation with no `JP_DEBUG_APP_SLOT`, `Slot::for_context` derives the slot from `ctx.conversation_id`; `default` is only the fallback when that ID has no usable characters. This instruction sends the agent to `tmp/debug-app/default` even though its artifacts live under the conversation-derived slot. Describe that default here, including the rare `default` fallback.

.config/jp/tools/src/util/paths.rs
```diff
@@ -0,0 +1,219 @@
...
+///
+/// The boundary test is what keeps `/Users/jean` from rewriting the front of
+/// `/Users/jeanne`.
+/// A path continues into `/`, and stops at anything that cannot be part of a
+/// name.
+fn replace_at_boundaries(text: &str, prefix: &str, label: &str) -> String {
+    let mut out = String::with_capacity(text.len());
+    let mut rest = text;
+
+    while let Some(at) = rest.find(prefix) {
```
JeanMertz: nit: This checks only the character after a match, so a known prefix can be replaced in the middle of another path. For example, a report containing `/Volumes/Backup/Users/jean/file` with `HOME=/Users/jean` becomes `/Volumes/Backup$HOME/file`. Require a valid boundary before `at` as well, and add a mounted-backup-path case to the tests.

crates/contrib/xct2cli/src/trace/toc.rs
```diff
@@ -0,0 +1,336 @@
...
+    loop {
+        buf.clear();
+        match reader.read_event_into(&mut buf)? {
+            Event::Empty(s) | Event::Start(s) if local_name(&s)? == "process" => {
+                let attrs = collect_attrs(&s)?;
+                let name = attrs.get("name").cloned().unwrap_or_default();
+                let pid = attrs.get("pid").and_then(|v| v.parse().ok()).unwrap_or(-1);
+                let path = attrs.get("path").cloned();
+                out.push(TocProcess { name, pid, path });
+                if matches!(reader.read_event_into(&mut Vec::new())?, Event::End(_)) {}
```
JeanMertz: theoretical: This unconditionally consumes one extra XML event after both `Event::Start` and `Event::Empty`. With valid compact input such as `<process .../><process .../>`, the first process consumes and discards the second one; the fixture happens to put a whitespace event between them. I could not name a current Xcode version that emits the compact form, so this is not merge-gating, but the parser should handle `Empty` without reading again and skip to the matching end only for `Start`.

.config/jp/tools/src/debug_app/profile.rs
```diff
@@ -0,0 +1,548 @@
...
+    .join(" ");
+
+    let retention = "\nThe bundle from an attached recording is kept, so `mode: \"report\"` can \
+                     ask it further\nquestions. A system-wide one is destroyed at stop: it embeds \
+                     the environment of every\nprocess it recorded.\n";
+
+    let allocations = if tiers.contains(&Tier::Allocations) {
+        "\nAllocation attribution needs `MallocStackLogging` in the app's environment, and \
+         libmalloc\nreads that at process start. Against a running session this only works if the \
+         app was\nlaunched for it; otherwise start the bracket first and launch into it.\n\nWhat \
```
JeanMertz: nit: The allocation preview recommends starting the bracket before launching, but `start` rejects `Tier::Allocations` whenever no app session exists. A caller following this preview enters a workflow the implementation cannot execute. Keep the preview aligned with the enforced order: launch with `allocation_stacks: true`, then start the attached bracket.

.config/jp/tools/src/debug_app/session.rs
```diff
@@ -0,0 +1,442 @@
...
+    /// Reduced to what a bundle identifier accepts: the slot ends up inside a
+    /// reverse-DNS identifier, which takes only letters, digits and hyphens.
+    /// Anything else is dropped rather than rejected, and a name left with
+    /// nothing usable falls through — a slot that cannot be spelled is a
+    /// naming problem, not a reason for the tools to stop working.
+    fn named(overridden: Option<&str>, conversation: &str) -> Slot {
+        for candidate in [overridden.unwrap_or_default(), conversation] {
+            let cleaned: String = candidate
+                .chars()
+                .filter(|c| c.is_ascii_alphanumeric() || *c == '-')
```
JeanMertz: blocking: Distinct user-supplied slots can collapse to the same identifier because invalid characters are deleted. For example, `JP_DEBUG_APP_SLOT=agent_one` and `JP_DEBUG_APP_SLOT=agentone` both become `agentone`, so two agents that deliberately chose different slots can overwrite one session and drive or quit the same app. Reject invalid override names or encode/hash the original value so distinct inputs remain distinct.

.config/jp/tools/src/debug_app/launch.rs
```diff
@@ -0,0 +1,787 @@
...
+        .map_err(|e| format!("Failed to spawn `open`: {e}"))?;
+    if !launch.success() {
+        return error(format!(
+            "`open` refused to launch {}:\n\n```\n{}\n```",
+            spec.bundle,
+            launch.stderr.trim_end()
+        ));
+    }
+
+    let pid = wait_for_pid(&spec.state_dir, PID_TIMEOUT)?;
```
JeanMertz: blocking: If the app launches but never writes the pid file, for example a slow startup crosses the timeout or the build lacks `JP_DEBUG_STATE_DIR` support, this returns an error while leaving the GUI process running and unrecorded. A later launch sees no session and starts another instance against the same state directory. On timeout, locate and terminate the staged bundle instance, or record enough launch identity before waiting to clean it up reliably.

.config/jp/tools/src/debug_app/capture.rs
JeanMertz: Taken. `stage_bundle` deletes and re-copies the staged bundle on every launch and Xcode overwrites the dSYM in place, so with a two-day retention window the ordinary record → edit → rebuild → report loop lands on exactly this. And it does not fail: it resolves each address to whatever now sits at that offset and hands back a table of plausible names for code that never ran, which is the same failure `hotspots` already warns about for a wrong slide, by another route.

`Target` now carries the binary's `LC_UUID`, read through `xct2cli::symbol::macho::BinaryInfo`, which was already parsing it to recover slides from image loads. `Target::stale()` compares it against the binary on disk, and `from_bundle` refuses before opening the trace rather than after symbolicating.

Chose the UUID over copying the binary and dSYM into the recording. Copying is bulletproof but a dSYM is large and the byte budget is already a live constraint here — and the UUID answers the actual question, since it changes on a rebuild even when nothing about the path does.

Unguarded in two cases, both deliberate: a recording written before the field existed, and a binary that cannot be read either time. Both read as they always did rather than being refused, so this cannot turn a readable recording into an unreadable one.

.config/jp/tools/src/debug_app/capture.rs
JeanMertz: Confirmed and fixed. Once `age()` passes `PENDING_WINDOW`, `is_pending` goes false, `sweep` stops skipping the recording, and `enforce_limits` finds it past `RETENTION_WINDOW` — the same 48 hours — and calls `discard`, which takes the sidecar and with it `recorder_pid`. An `xctrace` writing to a path that no longer exists, and nothing left able to name it.

`sweep` now takes a `&dyn Signals` and interrupts a recorder that is still alive before deleting the record that names it, reporting what it stopped in the swept note.

Your framing is what made the fix obvious, so it is now in the code: liveness deliberately does not decide whether a bracket is *pending* — a recorder that failed on its own still left a bundle and a reason worth reading, which is what `PENDING_WINDOW`'s doc argues. It does decide whether one can be *deleted*. Two questions, two answers, and conflating them is what produced this.

The identity check from the pid-reuse thread applies here too: the recorder is confirmed to be `xctrace` before anything is sent, so an expired record whose pid has been reused signals nothing.

.config/jp/tools/src/debug_app/capture.rs
JeanMertz: Both halves confirmed and both fixed. `store` was a plain `fs::write`, so an interrupt mid-write leaves a partial file; `recordings` skips what will not parse, so the recording drops out of `pending` and out of every sweep; and `orphaned_bundles` tested only that a file with the right name *existed*, so the bundle was treated as owned and left. For a system-wide recording that is every process's environment sitting on disk with nothing tracking it, which is a good deal worse than the recording being lost.

`store` now writes to `<id>.json.writing` beside the destination and renames into place, so the record is the old one or the new one and never half of either. Staged in the same directory rather than the system temp directory, so the rename stays on one filesystem and cannot degrade into a copy.

`orphaned_bundles` now parses the sidecar instead of counting it. A file that names nothing leaves its bundle as unowned as one with no file at all, and the unreadable sidecar is removed alongside it so the pair does not come back.

.config/jp/tools/src/debug_app/drive.rs
JeanMertz: Taken, and the comment eight lines below the bug already promised what the code did not: "Restored whichever way the run went, and a failure is when it matters most."

Two `?`s inside the loop, not one — `act` and `reading` — and they skipped more than the restore. `session.store` and `marks::append` sit under it too, so a driver that died mid-run also lost the console offsets and every step boundary for the steps that had run.

The loop body now collects into a `Result` and breaks, and the error is returned after the housekeeping instead of instead of it. Everything owed happens on both paths: the focus and pointer go back, the offsets record what was reported, the marks record when the steps ran.

.config/jp/tools/src/debug_app/launch.rs
JeanMertz: Taken. Worth adding that the error message names a cause that makes this permanent rather than occasional: "a build without `JP_DEBUG_STATE_DIR` support" never writes the pid at all, so every attempt would leave another window on screen with nothing able to address it, and the next launch would start a second instance against the same state directory.

On timeout the launch now stops the instance before returning. Identity is available without the pid file, which is the part that makes this tractable: the staged bundle is this slot's own copy, deleted and re-made a moment earlier, so `pkill -f <staged binary>` cannot reach the developer's own build or another slot's. The message says whether something was stopped or whether nothing had started.

`app_binary` was already in `capture` for the same purpose, so this uses that rather than growing a second copy of where an executable lives inside a bundle.

.config/jp/tools/src/debug_app/marks.rs
JeanMertz: Taken. Step *N*'s `ended_ms` and step *N+1*'s `began_ms` are separate `now_ms()` calls with only string formatting between them, and under `reads: "none"` that gap is comfortably under a millisecond — so the shared boundary is not rare, it is the normal case for the mode a caller uses precisely when they are measuring.

`Mark::holds` is now half-open, `>= began_ms && < ended_ms`, which also reads as the thing it means: an interval belongs to whichever step was running when the work began. `overlapping` and `steps_in` compare window against window rather than point against window, so neither needed touching.

One assertion changed and one test is new: `a_moment_two_steps_share_belongs_to_the_later_one` pins the boundary against exactly this, since the old test asserted `holds(2_000)` on a window ending at 2000 and would have kept passing either way.

Low frequency, as you say — it needs an interval starting in that particular millisecond. But it inflates the counts this whole feature exists to make comparable between runs, and it does it invisibly, which is the combination worth a one-line fix.

.config/jp/tools/src/debug_app/profile.rs
JeanMertz: Taken. The window is real and not narrow: if the app crashes mid-bracket, `is_running` goes false, `debug_app_launch` stops refusing, and a replacement session can be recorded before anything closes the bracket. Profiling a build that crashes is a reason people profile, so this is not an exotic ordering.

`start` now sets `target: Some(Target::for_session(session))` for `Scope::Attach`, and `stop` fills the target only when it is still `None` — which leaves the system-wide case working as it did, since a bracket opened with no app running has nothing to name yet by construction.

Your last sentence is the whole argument and it is now the comment on the field.

One thing this does not subsume: pinning the target earlier still stores paths that a later build overwrites, which is the separate finding on `Target`. The two together are what make a retained recording answer for itself.

.config/jp/tools/src/debug_app/quit.rs
JeanMertz: Taken. Three `drop(...)`s discarding `Result`s, and the one that matters is `retire`: `remove_dir_all` can fail on permissions or I/O, and quit would then print the summary, remove `session.json`, and report success with a system-wide bundle still on disk and nothing left pointing at it.

`Closed` now carries a `cleanup: Vec<String>` and the report renders it under a bold heading, before the summary reads as a tidy ending, saying what a system-wide bundle holds and that it has to be deleted by hand. That placement is deliberate — the session record naming it is gone by the time anyone reads the report, so this is the only chance to say where it is.

The `close` stamp is captured the same way rather than dropped: a recording that failed to record its own closing is worth knowing about, since it is what the next sweep will act on.

.config/jp/tools/src/debug_app/quit.rs
JeanMertz: Taken — and the module doc one screen above the bug is what makes it indefensible: "a recorder attached to a process on its way out has nothing useful to do and no defined behaviour." The code walked straight into the state its own comment describes, and then removed the session record, which was the last thing able to name either the app or the bracket.

`run` now returns before touching the app when `closed.stop == Stop::Stuck`, keeping `session.json` so a retry can still find both. The message `capture::stop` already writes for this case is the right text — it names the recorder's pid, the bundle, and the `kill -INT` to send by hand — so it is reused rather than rewritten.

.config/jp/tools/src/debug_app/report.rs
JeanMertz: Taken. `run_marks` returns `marks::overlapping(...)` for a named recording with nothing filtering by run, and a session is explicitly designed to hold several drives inside one bracket — so this is the ordinary case rather than an edge, and the output was a table headed with one run id containing another run's rows and repeating step numbers.

Went with grouping rather than refusing. `by_run` splits the marks, and `render_timelines` renders a table per run when there is more than one, opening with a line saying why they are kept apart. Refusing and asking the caller to pick would have made them do work the report can do itself, and the second run is usually the one they wanted.

Single-run windows render exactly as before, so the common path is untouched.

.config/jp/tools/src/debug_app/report.rs
JeanMertz: Taken. The guard was on `steps.len() != against_steps.len()`, which is exactly the case your example does not trip: three selections against three resizes matched by position and printed deltas as though each pair measured one thing.

Rows now pair on `(step, label)`. The labels carry the payload, so two runs of the same list match exactly and two different lists do not. A position that exists on both sides with a different label renders `(other step)` in place of a delta, and a note at the bottom says those rows hold the same position without doing the same thing. The count warning stays for the case it was written for.

Marking rather than refusing because the rest of the table is still worth reading — a run that diverges at one step is the interesting comparison, not a reason to decline the whole thing.

.config/jp/tools/src/debug_app/report.rs
JeanMertz: Taken, and the shape you describe is the normal one rather than a corner: an outer `conversation.select` containing shorter nested intervals starts first and ends last, so ordering by `started_ms` puts the newest sample at `first()`.

That inverts the headline figures of `view: "allocations"` and the conclusion drawn from them — the peak-above-where-it-ended paragraph reads "that much was transient" off numbers that have swapped ends.

The sampled intervals are now sorted by `at_ms` before the ends are taken. Your pointer to `stream::count` was the useful part: it already does this with its `at_ms >= sampled_at` check, and the comment there says exactly why. Having the same reasoning in two places and the correct behaviour in only one was the actual defect, so the new comment says what it is guarding against rather than restating the sort.

.config/jp/tools/src/debug_app/session.rs
JeanMertz: Fixed, though not quite for the reason given, and not by hashing.

The collision as stated needs two people to pick slot names differing only in characters the filter strips, which is a construction rather than something anyone does. The default path cannot reach it at all: the slot comes from a conversation id, which is already alphanumeric.

What is worth fixing is the other half of your example. `JP_DEBUG_APP_SLOT=my slot` silently becomes `myslot`, and the artifacts are not where the caller goes looking — the filter answers a different question than the one asked, without saying so.

So an override is now held to the character set rather than reduced to it, and refused by name when it does not fit. A conversation id is still filtered, because nobody chose it and there is nobody to report it back to. Your case falls out of that: `agent_one` is refused rather than quietly becoming `agentone`.

Not hashing, because a slot is a directory name and part of a bundle identifier, and both are meant to be read by a person looking for their own artifacts. Encoding would keep them distinct at the cost of making every one of them unrecognisable.

`Slot::for_context` returns a `Result` now, so the refusal reaches the caller as a tool error rather than a panic.

.config/jp/tools/src/debug_app/session.rs
JeanMertz: Taken, with a lighter mechanism than the one proposed.

The sequence is real: the app dies leaving its pid file, nobody relaunches — a relaunch would rewrite it, since `prepare_state_dir` clears the stale one — the pid space wraps past five digits, and `debug_app_quit` sends `SIGTERM` and then `SIGKILL` to a stranger. Low probability per occurrence, but the damage lands entirely outside the operation, which is what settles it.

`Signals` gains `is(pid, expected)`, checked before anything is sent. `quit` compares against the session's staged bundle path, which is slot-unique and freshly copied, so it cannot match another slot's app or the developer's own build. `sweep` uses the same call against `xctrace` before interrupting an abandoned recorder. An unreadable process reports as not matching, so the failure mode is refusing to signal — which degrades to the already-handled "it was gone" path.

Went with the executable path rather than a birth token because it needs no second file, no second thing that can go stale, and no timestamp parsing, and the staged bundle already gives an unambiguous answer. A start-time token is more robust in general; here it would be more machinery for the same guarantee.

The new `Termination::Reassigned` says what happened rather than reporting a clean stop, and `does_not_signal_a_pid_that_belongs_to_something_else_now` asserts the absence of any signal — which only a fake can do, since a real process cannot report what it was not sent.

One test needed changing rather than the code: `stops_a_running_process_with_sigterm` drives a real `sleep`, which is of course not the app the session names, so it now uses a `Signals` that delegates to the real one and answers the identity check. That test is about the escalation ladder against a process that really exits; identity is the subject of its own test.

.config/jp/tools/src/debug_app/tree.rs
JeanMertz: Confirmed on both sides and fixed in both. `tree::args` emitted `--identifier`, `--max-matches`, `--depth` and `--frames` and neither of these, and `Tree.swift` read `element.actions` for every retained node with no way to opt out and had no menu-bar branch at all — so the subtree was walked and then dropped by `render`.

I had triaged this as a follow-up on the grounds that `jpdrive` is not in this PR's diff. You asked for it here, and you were right to: the tool half alone cannot fix it, and splitting it would have left the wasteful version shipping while the flags landed elsewhere.

`TreeOptions` gains `actions` and `menus`. Actions are read only when asked for. The menu bar is left unwalked unless asked for, so its elements are never read — reading and discarding them costs the same couple of hundred round-trips as reading and showing them. The bar is still reported with its children counted as elided, so a reader can see it is there.

The Rust side passes `--actions` and `--menus`, and `render` no longer filters what it no longer receives. The menu note now comes from the elision count rather than from children it was about to discard, which keeps "pass `menus` for them" — the one elision a caller can undo.

`jpdrive menu` passes `menus: true` unconditionally, since walking the menu bar is the whole of what that subcommand does.

Two tests changed rather than the code, both of which had encoded the old division of labour: the fixtures assumed the driver always sends actions and menu children and the tool filters. They now model what each kind of read actually returns, and a new test pins that the flags are passed rather than applied after the fact.

The Swift half is unverified — `swift_format` and `swift_check` are not reachable from this branch, so `just fmt-app` and `just test-drive` are worth running before this goes out.

.config/jp/tools/src/util/paths.rs
JeanMertz: Taken. Only the trailing side was checked, so any occurrence of a known prefix mid-string was rewritten — your `/Volumes/Backup$HOME/file` is exactly it.

Both ends now use one predicate, `continues_component`, and the doc says why each side needs it: without the trailing test `/Users/jean` rewrites the front of `/Users/jeanne`, and without the leading one it rewrites the middle of a mounted copy.

A separator deliberately does not continue a component, which keeps `file:///Users/jean/notes.md` shortening while `/Volumes/Backup/Users/jean` does not. That distinction is worth a test of its own, so there are two: the backup path you named, and the URL case that the naive fix would have broken.

.jp/config/skill/debug-app.toml
JeanMertz: Taken. `Slot::for_context` reads the variable, falls through to `ctx.conversation_id`, and only reaches `FALLBACK_SLOT` when neither yields a usable character — so an agent following that paragraph looks in `tmp/debug-app/default` for artifacts sitting under its conversation's slot.

The paragraph now leads with the conversation id, names `default` as the fallback it is, and says outright that artifacts are under `tmp/debug-app/<conversation-id>/`, since being able to find them is the whole reason this is documented.

Its next sentence gets better rather than worse from the correction: two agents not colliding is true *because* the slot comes from the conversation, without either of them choosing anything. That now reads as the consequence it is.

Also mentions that an invalid override is refused, which is new behaviour from the slot-collision thread.

crates/contrib/xct2cli/src/trace/toc.rs
JeanMertz: This one is already the way you are asking for, which is what the `outdated` marker is telling us.

`toc.rs:247-255` handles `Event::Empty` by pushing the process and nothing else, and calls `skip_to_end` only in the `Event::Start` arm. The comment above it describes the hazard in your terms: "Self-closing `<process/>` has no end tag to consume. Consuming one anyway swallows whatever follows, which for adjacent processes is the next one and for the last is `</processes>`."

The file's last change is `81c0d78f` (#1002), which is in this branch's base rather than in it — `crates/contrib/xct2cli` is not in this PR at all. The duplicate commit the description mentions was dropped when this branch was rebased, so the description is stale too and I will fix that.

No change here.

PR #1013, opened 2026-08-19
feat(gui): Add the `jp gui` command plugin

JeanMertz:
Opening the macOS app at the workspace you are already working in meant
finding the bundle and passing it a path by hand. `jp gui` does it from
wherever `jp` itself is pointed.

The plugin resolves nothing. `InitMessage` carries a workspace root the
host has already settled, so `jp gui`, `jp -w ../other gui` and
`jp gui .` all arrive here answered; the plugin sends `Ready`, launches
the app at that root, and sends `Exit`. The app is found by bundle
identifier rather than by path, so it works wherever it is installed.

Run directly from a terminal it prints its usage and says it is a
plugin, rather than sitting silent waiting for a protocol message on a
stdin that will never carry one.

It joins `non_jp_excludes` alongside the other command plugins, so the
JP-only test scope stays JP-only.

### DIFF
--- a/crates/plugins/command/gui/Cargo.toml
+++ b/crates/plugins/command/gui/Cargo.toml
@@ -1,0 +1,35 @@
+[package]
+name = "jp-gui"
+
+authors.workspace = true
+description = "Open the current workspace in the JP macOS app."
+documentation.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license-file.workspace = true
+publish.workspace = true
+readme.workspace = true
+repository.workspace = true
+version.workspace = true
+
+[dependencies]
+jp_plugin = { workspace = true }
+
+serde_json = { workspace = true, features = ["std"] }
+
+[dev-dependencies]
+pretty_assertions = { workspace = true, features = ["std"] }
+
+[lints]
+workspace = true
+
+[package.metadata.jp-registry]
+id = "gui"
+command = ["gui"]
+description = "Open the current workspace in the JP macOS app"
+official = true
+repository = "https://github.com/dcdpr/jp"
+
+[[bin]]
+name = "jp-gui"
+path = "src/main.rs"
--- a/crates/plugins/command/gui/src/launch.rs
+++ b/crates/plugins/command/gui/src/launch.rs
@@ -1,0 +1,42 @@
+//! Launching the app, behind a trait so the rest can be tested without opening
+//! a window.
+
+use std::process::Command;
+
+/// Starts the app on a workspace.
+pub(crate) trait Launcher {
+    /// Open the app identified by `bundle_id`, showing the workspace at `path`.
+    fn launch(&self, bundle_id: &str, path: &str) -> Result<(), String>;
+}
+
+/// Launches through macOS Launch Services.
+pub(crate) struct SystemLauncher;
+
+impl Launcher for SystemLauncher {
+    /// Hands the workspace over as `JP_WORKSPACE`, which is what the app reads
+    /// when a window opens with no workspace of its own.
+    ///
+    /// `open -b` finds the app by bundle identifier, so nothing here depends on
+    /// where it was installed.
+    /// `-n` is deliberately absent: a second `jp gui` for a workspace already
+    /// on screen should bring that window forward rather than start a second
+    /// copy of the app.
+    fn launch(&self, bundle_id: &str, path: &str) -> Result<(), String> {
+        let status = Command::new("open")
+            .arg("-b")
+            .arg(bundle_id)
+            .arg("--env")
+            .arg(format!("JP_WORKSPACE={path}"))
+            .status()
+            .map_err(|e| format!("could not run `open`: {e}"))?;
+
+        if status.success() {
+            return Ok(());
+        }
+
+        Err(format!(
+            "could not open the JP app (`open` exited with {status}). Build and install it with \
+             `just build-app`, or open it once by hand so macOS knows where it is."
+        ))
+    }
+}
--- a/crates/plugins/command/gui/src/launch.rs
+++ b/crates/plugins/command/gui/src/launch.rs
@@ -12,6 +12,23 @@ pub(crate) trait Launcher {
 /// Launches through macOS Launch Services.
 pub(crate) struct SystemLauncher;
 
+/// Why `target_os` cannot open the app, or `None` if it can.
+///
+/// The plugin builds and publishes for every target the release workflow
+/// covers, because the registry has no way to say a plugin is macOS-only.
+/// So a Linux or Windows user can install it, and what they get should name the
+/// reason rather than fail reaching for a `macos` binary that was never going
+/// to be there.
+pub(crate) fn unsupported_platform(target_os: &str) -> Option<String> {
+    if target_os == "macos" {
+        return None;
+    }
+
+    Some(format!(
+        "`jp gui` opens the JP macOS app, which does not run on {target_os}."
+    ))
+}
+
 impl Launcher for SystemLauncher {
     /// Hands the workspace over as `JP_WORKSPACE`, which is what the app reads
     /// when a window opens with no workspace of its own.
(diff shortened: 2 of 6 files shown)


crates/plugins/command/gui/src/launch.rs
```diff
@@ -0,0 +1,42 @@
...
+    /// where it was installed.
+    /// `-n` is deliberately absent: a second `jp gui` for a workspace already
+    /// on screen should bring that window forward rather than start a second
+    /// copy of the app.
+    fn launch(&self, bundle_id: &str, path: &str) -> Result<(), String> {
+        let status = Command::new("open")
+            .arg("-b")
+            .arg(bundle_id)
+            .arg("--env")
+            .arg(format!("JP_WORKSPACE={path}"))
```
JeanMertz: blocking: A plausible sequence is: run `jp gui` in workspace A, then choose File > Open Workspace for B. `open --env` makes `JP_WORKSPACE=A` part of the app process environment. `WorkspaceWindow.chooseWorkspace` from #1008 consults that value before its scene-stored path on every load, so the attempt to open B is reset to A; new and restored windows stay pinned until the app quits. This state outlives the plugin command and affects every later window. The inverse also fails: if the app is already running, LaunchServices reuses that process and cannot replace its environment, so `jp -w B gui` can report success while merely activating A. The workspace needs to travel in a per-open channel the running app handles, such as a document or URL open event or IPC, rather than process environment.

crates/plugins/command/gui/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+pretty_assertions = { workspace = true, features = ["std"] }
+
+[lints]
+workspace = true
+
+[package.metadata.jp-registry]
+id = "gui"
+command = ["gui"]
+description = "Open the current workspace in the JP macOS app"
+official = true
```
JeanMertz: nit: A Linux or Windows user who runs `jp gui` gets an official downloadable plugin because the plugin workflow builds each `crates/plugins/command/*` manifest for its macOS, Linux, and Windows targets, and this registry entry has no platform restriction. The binary then tries macOS `open` and reports advice about installing a Mac app. This is contained and visible, but please avoid publishing non-macOS checksums for `jp-gui`, or fail with an explicit macOS-only diagnostic if the registry cannot express platform support.

JeanMertz: Confirmed, and taken — with the diagnostic rather than the registry half.

The workflow builds a four-target matrix (`plugins.yml:24-38`) and **hard-fails** if a registered plugin has no binary for one of them (`plugins.yml:76-79`), so `jp-gui` isn't merely publishable off macOS: the release job breaks unless it is. And the registry genuinely cannot say otherwise — `PluginMeta` in `build-registry/src/main.rs:24-34` carries `id`, `command`, `description`, `official`, `requires`, `suggests`, `repository`, and no platform field. `RegistryPlugin` doesn't either. So your fallback is the only branch available here.

`SystemLauncher::launch` now returns early with the reason:

```
`jp gui` opens the JP macOS app, which does not run on linux.
```

Done as a pure `unsupported_platform(target_os)` called with `std::env::consts::OS`, rather than a `#[cfg(not(target_os = "macos"))]` body — a cfg'd branch is compiled only on the platforms it targets, so neither this machine nor the macOS CI job would ever type-check it. The pure form is checked everywhere and unit-tested for `macos`, `linux` and `windows`. `Describe` still answers, so `jp plugin list` keeps explaining what the plugin is.

Adding a `platform` field so non-macOS checksums are never published is the actual fix, and it spans `jp_plugin::registry`, `build-registry`, the installer and the workflow — its own change, not this one.

Unrelated but found while verifying: **`jp-gui` did not compile on this branch.** `PluginToHost::Ready` has gained a `ReadyMessage` payload since the plugin was written, and `jp-gui` sends a bare `PluginToHost::Ready`. It went unnoticed because `jp-gui` is in `non_jp_excludes`, so the default test scope skips it and `cargo check` passes. Fixed here by taking the same `jp_plugin::ready(REQUIRED_PROTOCOL, init.version)` handshake the other three command plugins use, with `REQUIRED_PROTOCOL = 1` — this plugin reads the workspace root out of `init` and launches an app, both of which protocol 1 carries. Worth asking whether the excluded plugins should still get a compile check somewhere, since this is exactly the gap that hid it.

crates/plugins/command/gui/src/launch.rs
JeanMertz: The conclusion is right and the fix you name is the right one. One of the two mechanisms isn't, and it happens to be the one carrying the `blocking` label, so it's worth separating them.

**"consults that value before its scene-stored path on every load" — this doesn't hold.** `WorkspaceWindow.swift:490` passes `honorEnvironment: !hasResolvedWorkspace`, and line 499 sets the flag immediately after the first resolution; `chooseWorkspace` only reads `JP_WORKSPACE` when it's true. The comment at 494-498 records this as deliberate, for exactly the failure you describe, and `WorkspaceWindowTests.swift:106` pins it. So File ▸ Open Workspace for B is not reset to A, and that window doesn't stay pinned.

**"new and restored windows stay pinned until the app quits" — this does hold**, by a different route. `hasResolvedWorkspace` is `@State`, so it's per window rather than per process. A window opened with ⌘N after `jp gui A` starts with the flag false, finds `JP_WORKSPACE=A` still in the process environment, and prefers it over the more recent B. Same for each window restored at the next launch: every one's first resolution lets A outrank its own stored path.

**The already-running case is the real defect.** `open -b … --env` applies the environment only to a process it actually launches; without `-n` (deliberately omitted, `launch.rs:22-25`) an already-running app is merely activated. And there's no other way in — I grepped `apps/macos` for `CFBundleURLTypes`, `CFBundleDocumentTypes` and `onOpenURL` and found none of them. So with the app up on A, `jp gui` in B activates A, `open` exits 0, and the plugin sends `Exit { code: 0 }`. The user asked for B, got A, and was told it worked. The plugin cannot detect this, which is why no amount of error handling on the current channel fixes it.

On severity: the damage is contained and largely visible — the window titles itself with the workspace name. By the project's own rule that would make it a suggestion. I'd still treat it as blocking on a different ground: this plugin's one job is "open the current workspace in the app", and on the second ordinary use it silently opens a different one. That's a contract failure rather than a degraded result.

**Not fixed in this push, deliberately.** The fix is your per-open channel, and it's app-side: declare a URL type in `project.yml`'s `info` block, handle it so a *running* process receives it, and have the plugin send `open -b <id> "jp://workspace?path=…"` instead of `--env`. That also removes claim B from the `jp gui` path, since nothing would set the variable any more. `JP_WORKSPACE` should stay for `just run-app` and the UI-test harness, where it's a genuine launch instruction with tests behind it.

Two things I'd want settled before writing it, rather than after:

1. **`jp://` is already taken.** The app exports `jp://<conversation-id>` for copy and drag, and the CLI uses the same scheme for attachment URIs. Registering it claims it system-wide for the app, so `jp://workspace?path=…` has to sit alongside conversation URIs without either shadowing the other — or app control gets its own scheme. Worth a decision rather than a guess.
2. **Which window handles it.** `onOpenURL` sits on a view inside a `WindowGroup`, so with several windows open the delivery target isn't obvious, and the wrong answer is every window jumping to the new workspace. A single-claim store, or an `NSApplicationDelegate` adaptor, both work; they behave differently when no window is open.

I also can't verify URL delivery from here — it needs a running app driven from outside — so I'd rather land it as its own reviewed change than improvise it blind at the end of a session. Happy to take it next, or leave it to whoever owns the app side.

One more thing neither of us raised: `open -b` resolves the bundle identifier through LaunchServices, which only knows the app once it's been registered — from `/Applications`, or by being launched once from where it was built. `just build-app` leaves it in DerivedData and there's no install recipe, so on a clean machine `jp gui` fails even on macOS. The "build and install it" message happens to be the right advice, by accident. Worth an install step before this is advertised as official.

PR #1017, opened 2026-08-20
chore(ticket): Add ticket labels

JeanMertz:
Users can label tickets by system area when creating or updating work items, then filter listings by every requested label. Labels appear in CLI, tool, and documentation board views.

Boards define active and retired labels in `docs/ticket/.labels.json`. Writes reject unknown labels, while retired labels remain available on tickets that already carry them.


.config/jp/tools/src/ticket.rs
```diff
@@ -125,18 +145,19 @@ fn create(
...
-        title.trim(),
-        HANDLE,
-        &date,
+        title: title.trim(),
+        authors: HANDLE,
+        date: &date,
         implements,
-        &body.unwrap_or_default(),
-    )?;
+        labels,
```
JeanMertz: blocking: A plausible `ticket_create` call with `p:jp_attachment_agentic_shepherd`, `p:jp_attachment_mcp_resources`, and `p:jp_conversation` makes the `Labels` metadata line exceed the 80-column formatter width. The `reflow(&path)` immediately below treats that metadata line as Markdown list prose and wraps it onto an indented continuation, while `parse::metadata_range` and `parse::labels` only consume consecutive `- **Key**:` lines. The call reports success, but later reads retain only the labels on the first line and treat the continuation as description text. `ticket_comment` can corrupt an existing long label line the same way when it reflows the whole file. Please keep the metadata block out of `comfort` reflow, or give labels a formatter-stable multiline representation, and add a regression test with a label set wider than 80 columns.

justfile
```diff
@@ -2054,6 +2055,81 @@ ticket-list *ARGS: _install-ticket
...
+
+    source="docs/ticket/.labels.json"
+    target=".jp/mcp/tools/ticket/labels.toml"
+
+    if [ ! -f "$source" ]; then
+        echo "No label vocabulary at ${source}." >&2
+        exit 1
+    fi
+
+    names=$(jq -r '.active | keys[] | "  \"\(.)\","' "$source")
```
JeanMertz: blocking: A ticket that already carries a retired label such as `legacy-ui` cannot be updated through a schema-compliant `ticket_label` call. This active-only enum excludes the retired value even though the tool contract requires the caller to repeat that value to preserve it; omitting it silently removes persisted metadata. The same generated constraint prevents `ticket_list` from expressing its documented retired or deleted-label lookup. Please generate per-tool schemas: active labels for `ticket_create`, and active plus retired labels, or an open string schema, for `ticket_label` and `ticket_list`.

crates/internal/ticket/src/labels.rs
```diff
@@ -0,0 +1,337 @@
...
+    /// label is listed as both active and retired.
+    pub fn parse(source: &str) -> Result<Self, Error> {
+        // An empty file is an empty vocabulary rather than a syntax error: it
+        // is what `touch` leaves behind, and it means the same thing.
+        if source.trim().is_empty() {
+            return Ok(Self::default());
+        }
+
+        let document: Document =
+            serde_json::from_str(source).map_err(|error| Error::Malformed(error.to_string()))?;
```
JeanMertz: blocking: A plausible vocabulary entry such as `"api,web": "API and web work."` parses and resolves successfully here, but `render::ticket` and `set_labels` serialize labels with commas and `labels::split` reads that value back as two labels. The write reports success while the persisted ticket has different metadata, and the docs build later reports both fragments as unknown. Validate vocabulary names at this boundary so every accepted name round-trips through the ticket format, rejecting at least commas, line breaks/control characters, empty names, and names changed by trimming. A render-then-parse regression test would pin down the invariant.

crates/internal/ticket/src/labels.rs
```diff
@@ -0,0 +1,337 @@
...
+
+        let both: Vec<String> = document
+            .active
+            .keys()
+            .filter(|name| document.retired.contains_key(*name))
```
JeanMertz: blocking: A vocabulary with active `CLI` and retired `cli` passes this exact-key overlap check. Resolution is ASCII-case-insensitive and checks the active map first, so a fresh write of `cli` is accepted even though the same logical label is retired. Reject case-insensitive duplicates across the two maps, and within each map, before constructing `Vocabulary`; otherwise the canonical spelling and retirement policy depend on `BTreeMap` iteration order.

.jp/mcp/tools/ticket/create.toml
```diff
@@ -24,7 +24,7 @@ File a bug with a description:
 ```json
 {"kind": "bug", "title": "Tool call header misaligned", "body": "The header \
 renders one column left of the body when the terminal is narrower than 80 \
-columns."}
+columns.", "labels": ["cli"]}
```
JeanMertz: nit: Copying this documented call produces a visible refusal because the vocabulary added in this PR defines `c:cli`, not `cli`. The label and list tool examples, the RFD sample, and the new justfile examples also use undefined `cli`/`config` names. Please use pasteable values from `.labels.json`, likely `c:cli` and `p:jp_config`, throughout the user-facing examples.

.config/jp/tools/src/ticket.rs
```diff
@@ -53,20 +56,37 @@ pub fn run(ctx: Context, t: Tool) -> ToolResult {
...
+
+        "label" => {
+            let id = match id_arg(&t.req("id")?) {
+                Ok(id) => id,
+                Err(message) => return error(message),
+            };
+            label(
+                root,
+                id,
+                &t.opt::<Vec<String>>("labels")?.unwrap_or_default(),
```
JeanMertz: blocking: A model call such as `ticket_label({"id":"T-02wt0kx","labels":null})` clears every label on the ticket and reports success. `Tool::opt` turns JSON null into `None`, and this default turns that into the destructive empty-set operation. The host's `validate_arguments_against` checks that the required key exists but does not reject null, so this reaches the tool in normal dispatch. Read this required argument with `req::<Vec<String>>` and reject null/missing values before writing; reserve clearing for an explicit `[]`. Add a regression test asserting the ticket stays byte-for-byte unchanged on null.

docs/.vitepress/config.mts
```diff
@@ -199,6 +199,12 @@ const ticketBoardWriter = {
                     status: field('Status'),
                     kind: field('Kind'),
                     blockedBy: field('Blocked by'),
+                    // Repeated once per pair; a formatter escapes what would
+                    // otherwise be emphasis, so reading undoes that.
+                    labels: [...content.matchAll(/^- \*\*Label\*\*:\s*(.+)$/gim)]
```
JeanMertz: blocking: A ticket discussing label handling can quote `- **Label**: package=jp_config` in a fenced example in its description or a comment. This scan collects that line as actual ticket metadata, even when the header carries different labels or none. `TicketBoard` replaces its initial loader data with this endpoint's response on mount in dev mode, so the local board silently shows labels the ticket does not carry. Use the header-scoped `readLabels` reader here and cover a ticket containing a quoted label line.

docs/.vitepress/loaders/ticket-shared.mjs
```diff
@@ -48,6 +49,93 @@ export function parseTicket(content, filename) {
...
+// The field repeats once per pair rather than joining them onto one line, so no
+// line grows with the number of labels and the block survives `comfort`'s width
+// wrapping. A formatter escapes what would otherwise be emphasis (`jp_config`
+// becomes `jp\_config`), so reading undoes that.
+//
+// Read as written rather than checked against the vocabulary: a listing that
+// hid a label the file carries would disagree with the file.
+// `findUnknownLabels` below is what catches one the board doesn't define.
+export function readLabels(content) {
+    const block = content.match(/^# .+\n\n((?:- \*\*[^*]+\*\*:.*\n)+)/m)?.[1]
```
JeanMertz: blocking: A ticket saved with CRLF line endings, for example after editing it on Windows, returns no labels here. The header match requires literal `\n\n`, while the file contains `\r\n\r\n`; JavaScript `.` does not consume the `\r`. Rust's `str::lines()` still reads the labels, so CLI results and the published board disagree, and `findUnknownLabels` silently skips validation of that ticket's labels. Normalize line endings before matching or parse the header line by line. Please add a CRLF fixture with an unknown label as well as a known one.

crates/plugins/command/ticket/src/main.rs
```diff
@@ -82,6 +90,24 @@ enum Command {
...
     },
 
+    /// Replace a ticket's labels.
+    ///
+    /// The whole set is written: pass every label the ticket should end up
+    /// with, or none to clear them.
+    Label {
+        /// Ticket id: `T-02wt0kx`, `T02wt0kx`, or `02wt0kx`.
+        /// Omit it to choose.
+        id: Option<TicketId>,
```
JeanMertz: nit: `jp ticket label --label client=cli` in an interactive terminal does not offer the advertised ticket picker. `compose_missing` has no `Command::Label { id: None, .. }` branch, so execution reaches `required(id, "ticket id")` and reports that there is no terminal. Add a picker branch that preserves `labels`, following the existing Show/Edit branches, and a protocol-level test for the omitted-id path.

justfile
JeanMertz: nit: The active/retired split addresses the preservation problem on this revision. One list-schema mismatch remains: `ticket_list` documents `{"labels":["package"]}`, but its generated `items.enum` contains pairs and excludes bare `package` (and the other valued keys). A model following a strict schema cannot express that filter. Give `ticket_list` its own selector schema, preferably an open string schema so deleted-label lookups remain expressible too; keep the write enums separate.

.config/jp/tools/src/ticket.rs
JeanMertz: Confirmed and fixed. The chain is exactly as described: `Tool::opt` maps `Some(Value::Null)` to `None` (`.config/jp/tools/src/lib.rs:75`), and `validate_arguments_against` tests `!arguments.contains_key(name)` (`crates/jp_llm/src/tool.rs:1167`), which is false for a present-but-null key — so `required = true` doesn't stop it.

Two things make it worse than stated. `result = "unattended"` means no approval prompt, and this PR sets `style.inline_results = "off"` on `ticket_label`, so the `Cleared the labels on …` line is never rendered. The destructive write is silent.

Fixed in the `"label"` arm by refusing a missing-or-null list rather than defaulting it:

```rust
let Some(labels) = t.opt::<Vec<String>>("labels")? else {
    return error("`labels` must be an array. Pass `[]` to clear the labels.");
};
```

A tool `error()` rather than `req::<Vec<String>>`, because serde's `invalid type: null, expected a sequence` doesn't tell the model that `[]` is how you clear. `ticket_list` keeps `unwrap_or_default()` — null there means "no filter", which is harmless.

Two tests: `a_null_label_list_is_refused_and_writes_nothing` asserts the file is byte-for-byte unchanged after the refusal, and `an_empty_label_list_still_clears` pins that clearing stays reachable.

.config/jp/tools/src/ticket.rs
JeanMertz: Dismissing this one as resolved by the format change rather than by a fix at this line. GitHub kept the thread anchored because `ticket.rs:172` still exists, but the single joined `Labels` line it describes doesn't.

Labels are written one `- **Label**:` line per pair now, so the longest possible line is `- **Label**: package=jp_attachment_agentic_shepherd` — 51 columns, escapes included — however many labels a ticket carries. `no_label_line_grows_with_the_number_of_labels` (`crates/internal/ticket/src/store_tests.rs:654`) asserts no line in a three-long-package ticket exceeds 80.

The underlying point stood and is what drove the format. Worth recording why the fix isn't the one suggested: keeping the metadata block out of `comfort` was the other option and it doesn't hold, because every markdown file in the repo is formatted by `cargo comfort` and by CI, not just the ones written through this tool. The block has to survive the formatter rather than avoid it.

crates/plugins/command/ticket/src/main.rs
JeanMertz: Confirmed and fixed. `compose_missing` had arms for Add, Comment, Close, Show, Edit, Delete and Promote but none for Label, so `main.rs:1202` went straight to `required(id, "ticket id")` — which reports "no terminal to ask for one" even when there is one, while the subcommand's own doc comment says "Omit it to choose".

Added the branch next to the Show/Edit ones, carrying `labels` through.

On the test: a scripted `Compose`/`Composed` exchange has no precedent in this file and would pin a wire format that can drift silently, so I went narrower. `labelling_without_an_id_offers_the_picker` runs `compose_missing` against an empty board and asserts the picker's own `No tickets to choose from.` — reaching that error is what distinguishes the branch firing from falling through, and it fires before any protocol I/O. The `labels` pass-through is compiler-enforced by the struct literal, so a test adds nothing there.

justfile
JeanMertz: Confirmed and fixed. `labels.toml` gave `ticket_list` pairs only, while `list.toml` documents `{"labels": ["package"]}` as the way to find everything under a facet — the schema and the prose contradicted each other.

`ticket-labels-sync` now emits three lists rather than two: addable for `ticket_create`, plus-retired for `ticket_label`, and plus-retired-plus-bare-keys for `ticket_list`. I kept an enum rather than opening it to a free string, since the bare keys are a closed set too and the enum stays informative that way; the generated file's header note says why the three differ.

One correction worth recording, because it changes how both comments in this thread should be read: the enum isn't enforced anywhere. `validate_arguments_against` (`crates/jp_llm/src/tool.rs:1153-1200`) checks unknown parameter names and missing required keys, and never looks at `enum` or `items.enum`. So a model that sent a bare `package` — or a retired value, per the original comment — already succeeded. What the enum does is steer the model, and constrain it hard under providers that restrict decoding to the schema. The mis-steer was real; the "cannot be updated through a schema-compliant call" framing overstated it.

docs/.vitepress/config.mts
JeanMertz: Confirmed and fixed, with one severity correction.

The scan was `matchAll(/^- \*\*Label\*\*:\s*(.+)$/gim)` over the whole file with no header scoping, and `TicketBoard.vue:67-70` does replace the loader's data with this endpoint's response under `isDev`. The input is real rather than hypothetical: RFD 100 already carries a `- **Label**: client=cli` line inside a fenced block, and a ticket discussing labels would quote one the same way.

The correction: this is `configureServer` middleware, so it is dev-only and never reaches the published board, and the `PUT` path writes `Status` only — labels are read for display and nothing else. Wrong labels on a local dev board, nothing persisted, no write path. Contained and reversible by a reload, so I wouldn't gate the merge on it.

Fixed as you suggested: `config.mts` imports `readLabels` from `./loaders/ticket-shared.mjs` instead of carrying its own scan, which deletes the duplicate reader rather than patching it — and picks up the CRLF fix from the sibling thread for free.

Not covered by a test, and worth saying plainly: `docs/` has no test runner (`package.json` has no vitest/jest and no test script), so this fix and the CRLF one are both unverified by automation. Adding a harness for the loaders is worth doing and belongs in its own change.

docs/.vitepress/loaders/ticket-shared.mjs
JeanMertz: Confirmed and fixed, with a severity correction.

The mechanism holds: JavaScript `.` excludes `\r`, so `^# .+` stops before it and the required `\n\n` never matches `\r\n\r\n`. Rust's `str::lines()` strips `\r\n`, so the CLI reads labels the board can't see.

I checked whether this predates the PR, and it doesn't. `field()` in `metadata.mjs:14` is CRLF-safe by accident — `(.+)` stops at `\r` and `.trim()` cleans up the rest — so `readLabels` is the only CRLF-fragile reader in the docs build, and this PR added it.

Where I'd downgrade it: `.gitattributes` carries only linguist markers, no `text=auto eol=lf`, so CRLF *can* land in the repo — but it takes a Windows contributor with `core.autocrlf` off, and there's no sign of one. Worth fixing anyway, mostly for the second half of your point: `findUnknownLabels` skips a ticket whose labels read as empty, so an unknown label in a CRLF file bypasses the build check that exists to catch it.

Fixed by normalizing line endings before the header match. No CRLF fixture, for the reason in the sibling thread: `docs/` has no test runner, so there's nowhere to put one without introducing a harness first.

PR #1018, opened 2026-08-21
chore(tools): Align RUSTFLAGS and time cargo runs

JeanMertz:
Every cargo tool sets `RUSTFLAGS`, which replaces the `rustflags` list in `.cargo/config.toml` wholesale rather than adding to it. Only `cargo_check` and `cargo_format` were setting it, so `cargo_test` and `cargo_expand` compiled with a different flag set than their siblings. In a workspace relying on those config flags, alternating between the two groups invalidated a shared target directory and forced a full rebuild each time.

All four compiling tools now set the variable, and a new `options.rustflags` appends to the `-W warnings` base so a workspace can restate the flags it needs. A malformed value is refused rather than dropped, matching how `options.root` already behaves: compiling with flags the caller believes are in effect is worse than refusing to compile.

Each invocation is also timed, with the duration appended to successes and failures alike. "Check succeeded" reads identically after a warm cache and after a three-minute rebuild, and the elapsed time is the only thing that tells them apart.


.config/jp/tools/src/cargo.rs
```diff
@@ -93,10 +95,28 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
         ));
     }
 
+    // Flags to append to `RUSTFLAGS`, set via `options.rustflags`. Malformed
+    // values are refused for the same reason as `root`: silently ignoring them
+    // would compile with flags the caller believes are in effect.
+    let rustflags = match t.options.get("rustflags") {
```
JeanMertz: blocking: A plausible configuration is `conversation.tools.cargo_install_tools.options.rustflags = ["--cfg", "tokio_unstable"]`. `cargo_install_tools` compiles and installs a persistent binary, but this option is accepted here while the `install_tools` arm never passes `rustflags` to `just install-tools`. The call can report success even though the installed binary was built without the flags the caller believes were active. Please pass the computed value into `cargo_install_tools` and its subprocess environment, or reject `rustflags` for that subcommand as the `profile` option does. The non-compiling `update` arm should likewise reject the option rather than accept a no-op.

.config/jp/tools/src/cargo/check.rs
```diff
@@ -34,8 +37,7 @@ fn cargo_check_impl<R: ProcessRunner>(
     let clippy_scope = package.map_or("--workspace".to_owned(), |v| format!("--package={v}"));
     let profile_arg = profile.map(|name| format!("--profile={name}"));
 
-    // Prevent warnings from being treated as errors, e.g. on CI.
-    let mut env = vec![("RUSTFLAGS", "-W warnings")];
+    let mut env = vec![("RUSTFLAGS", rustflags)];
```
JeanMertz: blocking: A JP session launched with `CARGO_ENCODED_RUSTFLAGS` set is a plausible Cargo environment, especially when a flag contains whitespace. Cargo checks that variable before `RUSTFLAGS` ([documented precedence](https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags)), and `DuctProcessRunner` inherits variables not named here. In that case this assignment is ignored, so the base and `options.rustflags` can be silently absent while the command succeeds and leaves artifacts built with another flag set. Set `CARGO_ENCODED_RUSTFLAGS` to the merged argument list, or otherwise ensure the inherited higher-priority variable cannot override the effective flags. A regression test should set both variables and assert which flags Cargo receives.

.config/jp/tools/src/cargo.rs
JeanMertz: Confirmed and fixed, taking the reject arm rather than plumbing the value through.

The mechanics are as you describe: `cargo.rs` called `cargo_install_tools(&root)` with no flags argument, and `install_tools.rs` runs `runner.run("just", &["install-tools"], root)` — `run` uses `RunnerOpts::default()`, so no env entries are set. `cargo install` does honor `RUSTFLAGS`, so the binary got whatever ambient or `.cargo/config.toml` flags applied, and the tool then reported "Rebuilt and installed the `jp-tools` binary". Hidden and persistent, since that binary answers every later tool call.

I went with rejection over plumbing for two reasons. `install_tools.rs` states the design outright — "Going through `just` keeps the install flags defined in one place, so this and a hand-run `just install-tools` cannot drift apart" — and injecting `RUSTFLAGS` from tool config recreates exactly that drift: `just install-tools` by hand would then build differently from the tool. And rejection matches how `profile` already behaves for the same subcommand, so there's one rule rather than two.

You're right about `update` too, and there was a third case: `cargo_format` was being handed the computed value, but `cargo fmt` runs rustfmt, not rustc, so a configured flag was inert there as well. The guard now covers all three, mirroring the `profile` one five lines above:

```rust
if configured_flags.is_some() && !matches!(subcommand, "check" | "expand" | "test") {
    return error(format!(
        "The `rustflags` tool option has no effect on `cargo_{subcommand}`."
    ));
}
```

Parsing is split from the guard so an explicit `rustflags = null` reads as absent rather than configured, matching `profile.is_some()`. `format` still receives the base `-W warnings`, so nothing changes there for callers who set no option.

Three tests added: the `install_tools` case (returns before `just` is spawned), the `format` case, and the explicit-null case. I also narrowed the `rustflags()` doc comment, which claimed "Every compiling cargo tool sets the variable" while counting `format` as a compiler.

.config/jp/tools/src/cargo/check.rs
JeanMertz: Both mechanical claims hold, and I checked rather than assumed. The Cargo Book's `build.rustflags` entry lists four mutually exclusive sources checked in order, `CARGO_ENCODED_RUSTFLAGS` first and `RUSTFLAGS` second. And `runner.rs` confirms inheritance: `clean_env` is `false` for `run_with_env`, so only the named entries are added via `command.env(key, value)` and the rest of the parent environment passes through. (`clean_env: true` exists, but only `unix_utils` uses it.)

I'm not taking the change, though, because I can't name an input a JP user plausibly produces that gets there.

`CARGO_ENCODED_RUSTFLAGS` is `\x1f`-delimited and is essentially only ever set by Cargo itself when running a build script — at which point it also removes `RUSTFLAGS` (rust-lang/cargo#13981). Nothing in this repo sets it, no justfile recipe sets it, and reaching `jp-tools` means it was exported into the environment that launched `jp`, which then reaches the tool through `just serve-tools`. Contrast the variable that is live here: `.github/workflows/rust.yml` exports `RUSTFLAGS: -D warnings`, which is the case this code already handles and the reason `-W warnings` exists at all.

The damage is also narrower than the comment suggests. All four tools inherit the same ambient value, so they still agree with each other — the target-directory thrashing this PR fixes came from some tools setting `RUSTFLAGS` and others not, and that stays fixed. If the ambient flags are stricter, the tool fails loudly with the diagnostics. The genuinely hidden case is: a user sets `options.rustflags`, an encoded value is already exported, and the build succeeds with neither.

Against that, the fix isn't small. Switching to `CARGO_ENCODED_RUSTFLAGS` settles a policy question this PR hasn't raised — whether JP's flags should override an encoded value a Nix- or Yocto-style environment set deliberately — and changes the wire format to `\x1f`-separated. The regression test can't be written against the current fixtures either: `EnvCapturingRunner` records only explicitly-set entries, so proving which flags Cargo receives needs a real invocation or a runner that models inheritance.

So: declining, and the worse result a user gets is that in an environment exporting `CARGO_ENCODED_RUSTFLAGS`, a configured `options.rustflags` is silently ignored. Happy to be overruled if you read that environment as more common than I do. The cheap middle option, if you want the hidden case made visible without the format change, is to check for the variable once in `cargo::run` and refuse with a message naming the precedence.

One thing from your comment I do want to keep, separately: `rustflags()` joins with `" "`, so a single flag containing a space (`-Clink-args=-Wl,-rpath,/a b/c`) gets split into two by Cargo. That has a named input — a linker flag written in a config file — and it fails silently today. Filing it as a ticket rather than folding it into this PR, since it's fixable without touching the variable choice.

PR #1035, opened 2026-08-25
test: Resolve test fixture paths at runtime

JeanMertz:
Test fixtures and insta snapshots were located through `env!("CARGO_MANIFEST_DIR")`, which rustc resolves when the test binary is compiled. With a target directory shared between git worktrees, the binary that runs is not always the one the current worktree built, so the baked path can point at a sibling checkout. Tests then fail with "No fixtures found" against another worktree's path, or silently write snapshots into that checkout, since `.config/insta.yaml` sets `update: always` with `auto_accept_unseen`.

`jp_test::fixtures_dir()` derives the path from the process working directory instead, which both `cargo test` and `cargo nextest` set to the package root. `Vcr::new` loses its `manifest_dir` argument and calls it internally: it needs an absolute path because it feeds insta's `snapshot_path`, and insta resolves relative snapshot paths against the assertion's source file rather than the working directory. Call sites with no insta involvement use plain relative paths.

`lint-ci` rejects the macro under `crates` and `.config/jp/tools` so the pattern cannot creep back. The check is a plain string match, so Rust source that needs to discuss the macro must avoid spelling it exactly.

### DIFF
--- a/crates/jp_test/src/lib.rs
+++ b/crates/jp_test/src/lib.rs
@@ -8,22 +8,25 @@ pub use test_log::test;
 
 /// Absolute path to the calling package's `tests/fixtures` directory.
 ///
-/// Derived from the process working directory, which both `cargo test` and
-/// `cargo nextest` set to the package root.
+/// Read from the `CARGO_MANIFEST_DIR` environment variable, which both `cargo
+/// test` and `cargo nextest` set for the process running a test, keyed to the
+/// package that test belongs to.
 ///
-/// Resolving at runtime rather than baking in `CARGO_MANIFEST_DIR` keeps the
-/// path correct when a compiled test binary is shared between git worktrees
-/// through a common target directory: a compile-time path records whichever
-/// worktree built the binary, which is not necessarily the one running it.
-///
-/// Callers must not change the working directory before calling this.
+/// Reading it at runtime rather than through the compile-time macro keeps the
+/// path correct when a test binary is shared between git worktrees through a
+/// common target directory: the macro records whichever worktree built the
+/// binary, which is not necessarily the one running it.
 ///
 /// # Panics
 ///
-/// Panics if the working directory cannot be read.
+/// Panics if the variable is unset, which happens when a test binary is
+/// launched directly instead of through cargo or nextest.
+/// Failing here beats returning a path that quietly resolves outside the
+/// package.
 #[must_use]
 pub fn fixtures_dir() -> PathBuf {
-    env::current_dir()
-        .expect("readable working directory")
-        .join("tests/fixtures")
+    let root = env::var_os("CARGO_MANIFEST_DIR")
+        .expect("CARGO_MANIFEST_DIR: run tests via `cargo test` or `cargo nextest`");
+
+    PathBuf::from(root).join("tests/fixtures")
 }
--- a/crates/jp_test/src/lib.rs
+++ b/crates/jp_test/src/lib.rs
@@ -1,5 +1,29 @@
+use std::{env, path::PathBuf};
+
 pub mod macros;
 pub mod mock;
 
 pub type Result = std::result::Result<(), Box<dyn std::error::Error>>;
 pub use test_log::test;
+
+/// Absolute path to the calling package's `tests/fixtures` directory.
+///
+/// Derived from the process working directory, which both `cargo test` and
+/// `cargo nextest` set to the package root.
+///
+/// Resolving at runtime rather than baking in `CARGO_MANIFEST_DIR` keeps the
+/// path correct when a compiled test binary is shared between git worktrees
+/// through a common target directory: a compile-time path records whichever
+/// worktree built the binary, which is not necessarily the one running it.
+///
+/// Callers must not change the working directory before calling this.
+///
+/// # Panics
+///
+/// Panics if the working directory cannot be read.
+#[must_use]
+pub fn fixtures_dir() -> PathBuf {
+    env::current_dir()
+        .expect("readable working directory")
+        .join("tests/fixtures")
+}
--- a/justfile
+++ b/justfile
@@ -2624,6 +2624,20 @@ ci: lint-ci fmt-ci test-ci docs-ci coverage-ci deny-ci insta-ci shear-ci vet-ci
 # Lint the code on CI.
 [group('ci')]
 lint-ci: (_rustup_component "clippy") _install_ci_matchers
+    #!/usr/bin/env sh
+    set -eu
+
+    # `env!("CARGO_MANIFEST_DIR")` is resolved when the binary is compiled. With
+    # a target directory shared between git worktrees, the binary that runs is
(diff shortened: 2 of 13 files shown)


crates/jp_test/src/lib.rs
```diff
@@ -1,5 +1,29 @@
...
+/// Callers must not change the working directory before calling this.
+///
+/// # Panics
+///
+/// Panics if the working directory cannot be read.
+#[must_use]
+pub fn fixtures_dir() -> PathBuf {
+    env::current_dir()
+        .expect("readable working directory")
+        .join("tests/fixtures")
```
JeanMertz: blocking: An IDE or debugger that launches the compiled `jp_llm` test binary with the workspace root as its cwd makes this return `<workspace>/tests/fixtures`, not `crates/jp_llm/tests/fixtures`. The compaction snapshots then run under `update: always` and `auto_accept_unseen`, so they can create and accept snapshots in the wrong directory while the suite passes. That is the same hidden write this PR is intended to prevent. Please fail fast unless the derived fixture directory already exists, or otherwise verify that the cwd is the package root before giving the path to insta.

justfile
```diff
@@ -2624,6 +2624,20 @@ ci: lint-ci fmt-ci test-ci docs-ci coverage-ci deny-ci insta-ci shear-ci vet-ci
...
+    #!/usr/bin/env sh
+    set -eu
+
+    # `env!("CARGO_MANIFEST_DIR")` is resolved when the binary is compiled. With
+    # a target directory shared between git worktrees, the binary that runs is
+    # not always the one this worktree compiled, so the baked path can point at
+    # a sibling worktree: fixtures and snapshots then resolve against the wrong
+    # checkout. Resolve at runtime instead, via `jp_test::fixtures_dir()` or a
+    # path relative to the package root.
+    if grep -rn --include='*.rs' 'env!("CARGO_MANIFEST_DIR")' crates .config/jp/tools; then
```
JeanMertz: theoretical: This guard rejects only the exact one-argument spelling. A multiline invocation or `env!("CARGO_MANIFEST_DIR", "custom error")` bypasses it, so a later edit can restore the baked path while `lint-ci` stays green. No current source file uses either form, so this is a regression-guard gap rather than a current runtime failure. Since the ban is workspace-wide, consider matching `CARGO_MANIFEST_DIR` in Rust source regardless of surrounding syntax, or use a syntax-aware check.

crates/jp_test/src/lib.rs
JeanMertz: Confirmed, and fixed — though not the way you suggested, because a better option exists.

The chain checks out end to end. The workspace root has no `tests/` directory, so the derived path doesn't exist; `compaction_request_tests.rs:135-141` hands it straight to insta's `snapshot_path` with no existence check; `.config/insta.yaml` sets `update: always` with `auto_accept_unseen: true`. Wrong cwd therefore means insta creates and auto-accepts snapshots under `<workspace>/tests/fixtures/<provider>/compaction/` while the suite reports success. Your closing line is the one that decides it: this PR exists to stop silent cross-directory snapshot writes, and it opened a new route to exactly that.

Worth recording that the exposure was not uniform, since only one site was silent. `Vcr` playback already gates on `fixture.exists()` at `mock.rs:112` and errors with `Recording not found at …`; jp_md's globs trip `assert!(!paths.is_empty())`; the `.config/jp/tools` paths panic on a failed read. All loud. The direct-insta consumer was the exception.

I didn't take either proposed remedy. Requiring the directory to exist keeps the cwd dependence and only converts silence into noise, and it would break recording a new provider's fixtures, where `RECORD=1` plus `with_fixture_suffix` legitimately targets a directory that isn't there yet. Verifying the cwd is the package root has no robust test, since the workspace root also has a `Cargo.toml`.

Instead `fixtures_dir()` no longer consults the cwd at all:

```rust
let root = env::var_os("CARGO_MANIFEST_DIR")
    .expect("CARGO_MANIFEST_DIR: run tests via `cargo test` or `cargo nextest`");

PathBuf::from(root).join("tests/fixtures")
```

Reading the variable at runtime is the upstream-endorsed answer for this exact case. The Cargo book notes these variables apply "for running binaries with `cargo run` and `cargo test` as well", with `CARGO_MANIFEST_DIR` among them, and nextest sets it to "the directory containing the manifest of the test's package". Nextest's archiving docs recommend it verbatim: "For relocatable tests, use the value of CARGO_MANIFEST_DIR at runtime. This means `std::env::var("CARGO_MANIFEST_DIR")`, not `env!("CARGO_MANIFEST_DIR")`."

That keeps this PR's original property (nothing baked at compile time, so a binary shared through a common target directory still resolves correctly), removes the cwd assumption you found, and makes the IDE/direct-launch case a loud panic rather than a silent write.

It also let the `Callers must not change the working directory` caveat go, which is a real gain rather than a cosmetic one: `jp_cli/src/lib_tests.rs:551` calls `set_current_dir`, so that coupling was a live trap and is now gone.

Full suite green (4964). That is a meaningful check here and not a vacuous one — had the variable been absent under nextest, the `expect` would have panicked and every fixture-consuming test would have failed. `git diff` also shows no `.snap` or fixture file touched, so nothing was written outside its package.

justfile
JeanMertz: Gap is real, and the `theoretical:` label is right — no current file uses either form. Widened the pattern, but not all the way to a bare match, because that would have banned the fix for your other comment.

You also understated it slightly: `option_env!("CARGO_MANIFEST_DIR")` slips through as well.

Matching `CARGO_MANIFEST_DIR` regardless of surrounding syntax would reject three correct things:

1. `env::var("CARGO_MANIFEST_DIR")` — the runtime read that `fixtures_dir()` now uses to fix `id=3851604220`. The guard would have rejected its own remedy.
2. Prose. I already had to reword `fixtures_dir()`'s doc comment during this PR because it named the macro; a bare match means the identifier can't appear in documentation anywhere, including in the helper whose entire purpose is to replace it.
3. A future build script's `env::var("CARGO_MANIFEST_DIR")`, which is the correct API in that context.

So the guard stays anchored on the macro call and widens to the bypasses that matter:

```sh
grep -rnE --include='*.rs' '(env|option_env)!\s*\(\s*"CARGO_MANIFEST_DIR"' crates .config/jp/tools
```

That covers the two-argument form, `option_env!`, and whitespace variants, while leaving runtime reads and prose alone.

I left the newline-between-paren-and-string case open, on the grounds that `fmt-ci` closes it independently — `cargo fmt --check` collapses that invocation onto one line, so the spelling can't survive CI even if someone wrote it. Flagging that I reasoned about rustfmt's behaviour rather than running it; if you'd rather not lean on a second job for this, a syntax-aware check is the honest alternative, though clippy's `disallowed_macros` can't express it (it keys on the macro path, not its arguments, and `env!` has four legitimate uses in the workspace for `CARGO_PKG_VERSION` and friends).

Noted the reasoning in a comment above the check so the next person doesn't "simplify" it back to a fixed string.

PR #1037, opened 2026-08-25
fix(config): Record a changed model id whole in deltas

JeanMertz:
Continuing a conversation with a `--cfg` persona could fail before the
first request with `Missing required value for field
conversation.inquiry.assistant.model.id.provider`. A conversation's
config is layered back in as a diff against the top-level assistant,
and because the inquiry model (`haiku`) shared its provider with the
assistant model (`opus`), only the model name survived that diff. The
persona then set `assistant.model.id` to an alias, which carries no
provider for the half-recorded inquiry id to inherit.

The same diff could also produce a plausible-looking wrong model
instead of an error: an inquiry recorded as `claude-haiku-4-5`, layered
under an assistant on a different provider, resolved to
`openai/claude-haiku-4-5` and failed only once the request was sent.

A model id is one value spelled across a provider and a name, so a
changed id is recorded whole rather than field by field. An unchanged
id still drops out of the diff completely, so an inquiry that inherits
its model keeps following later `assistant` changes.

### DIFF
--- a/crates/jp_config/src/lib_tests.rs
+++ b/crates/jp_config/src/lib_tests.rs
@@ -364,6 +364,70 @@ fn inquiry_model_survives_an_aliased_assistant_layer() {
     );
 }
 
+/// A half-written inquiry model id takes its missing half from an assistant
+/// model named by alias.
+///
+/// `--cfg conversation.inquiry.assistant.model.id.name=<name>` sets one field
+/// of the pair.
+/// The other has to come from the assistant, which is reachable only once the
+/// alias naming it has been resolved.
+#[test]
+fn a_half_written_inquiry_model_id_inherits_from_an_aliased_assistant() {
+    use crate::{
+        model::id::{PartialModelIdConfig, PartialModelIdOrAliasConfig},
+        util::build,
+    };
+
+    let mut partial = PartialAppConfig::new_test();
+    partial.providers.llm.aliases.insert(
+        "opus".to_owned(),
+        PartialModelIdOrAliasConfig::from("anthropic/claude-opus-5"),
+    );
+    partial.assistant.model.id = PartialModelIdOrAliasConfig::Alias("opus".to_owned());
+    partial.conversation.inquiry.assistant.model.id = PartialModelIdConfig {
+        provider: None,
+        name: Some("claude-haiku-4-5".parse().unwrap()),
+    }
+    .into();
+
+    let config = build(partial).expect("valid config");
+
+    assert_eq!(
+        config
+            .conversation
+            .inquiry
+            .assistant
+            .model
+            .id
+            .resolved()
+            .to_string(),
+        "anthropic/claude-haiku-4-5",
+        "the inquiry keeps its own name and inherits the assistant's provider"
+    );
+    assert_eq!(
+        config.assistant.model.id.resolved().to_string(),
+        "anthropic/claude-opus-5",
+        "and the assistant keeps the model its alias names"
+    );
+}
+
+/// An alias the map cannot resolve is still reported as an unresolved alias,
+/// not as a missing field.
+#[test]
+fn an_unknown_assistant_alias_reports_itself() {
+    use crate::{model::id::PartialModelIdOrAliasConfig, util::build};
+
+    let mut partial = PartialAppConfig::new_test();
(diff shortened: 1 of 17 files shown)


crates/jp_config/src/lib_tests.rs
```diff
@@ -307,6 +307,127 @@ fn inquiry_inheritance_survives_a_partial_round_trip() {
     );
 }
 
+/// A later layer naming the assistant model by alias must not strand the
+/// inquiry's own model.
+///
+/// A resolved config becomes a layer again through `to_partial` (a stored
+/// conversation config, a `--cfg` baseline), which records the inquiry model as
+/// a diff against the assistant's.
+/// An inquiry sharing the assistant's provider records only its name, and
```
JeanMertz: nit: This describes the behavior the production change removes. The delta records both provider and name here, not only the name. The test also stops guarding that behavior after the second commit: because `opus` resolves to the same provider, pre-inheritance alias resolution makes the final assertion pass even if the delta drops the provider again. Please assert the provider on `round_tripped.conversation.inquiry.assistant.model.id` before replacing the assistant ID, or make the later alias use a different provider so the final assertion exercises the atomic-ID delta.

crates/jp_config/src/lib_tests.rs
```diff
@@ -307,6 +307,127 @@ fn inquiry_inheritance_survives_a_partial_round_trip() {
...
+fn an_unknown_assistant_alias_reports_itself() {
+    use crate::{model::id::PartialModelIdOrAliasConfig, util::build};
+
+    let mut partial = PartialAppConfig::new_test();
+    partial.assistant.model.id = PartialModelIdOrAliasConfig::Alias("nope".to_owned());
+
+    let error = build(partial).unwrap_err().to_string();
+
+    assert!(
+        error.contains("assistant.model.id"),
```
JeanMertz: nit: This assertion does not establish the behavior in the test name. A missing-field error such as `Missing required value for field assistant.model.id.provider` also contains `assistant.model.id`, so this test would stay green if the alias were converted into an incomplete ID. Assert the alias-resolution error text, or at least assert that the error does not contain `Missing required value`.

crates/jp_config/src/lib_tests.rs
JeanMertz: Both halves confirmed, fixed in 26173c6b.

The doc comment described the pre-fix diffing behavior in the present tense; rewritten to state the rule that holds now and the failure it guards.

On the sensitivity: reverting the `delta` hunk and running the test proved your point exactly — the alias resolves to `anthropic` before inheritance, so `fill_from` supplied the same provider the dropped field had, and the closing assertion passed regardless. Took your first option, reading the recorded diff at the seam before the aliased layer goes on:

```rust
assert_eq!(
    round_tripped.conversation.inquiry.assistant.model.id.to_string(),
    "anthropic/claude-haiku-4-5",
    "the recorded diff carries the whole id, not just the changed name"
);
```

`PartialModelIdConfig`'s `Display` renders the name alone when the provider is `None`, so a dropped provider makes this read `claude-haiku-4-5`. Against the reverted hunk it fails with exactly that.

Your second option became its own test rather than a change to this one, since it pins a different outcome — wrong model rather than build error — and that outcome had no coverage at all despite being named in the commit message. `an_inquiry_model_keeps_its_provider_when_the_assistant_changes_provider` points the later alias at `openai/gpt-5.6-sol`; reverted, it produces `openai/claude-haiku-4-5`, which is the model that would only fail once the request was sent.

crates/jp_config/src/lib_tests.rs
JeanMertz: Correct, fixed in 26173c6b.

Went with the exact message rather than the negative assertion, since it subsumes it:

```rust
assert_eq!(
    error,
    "assistant.model.id: model ID does not match a known alias nor matches \
     the full ID format <provider>/<model>",
    "an unresolvable alias is reported as one, not as a missing field"
);
```

Verified against the regression you described: making `finalize_in_place` fall back to a default `Id` on an unresolvable alias yields `Missing required value for field assistant.model.id.provider.`, which the old `contains("assistant.model.id")` accepted and the comparison now rejects.

PR #1040, opened 2026-08-26
refactor(conversation, llm): Record event patches as overlays

JeanMertz:
Provider-issued repairs rewrote the events they targeted and flushed the result to disk, so a signature stripped to satisfy one provider was gone for every future request, model and provider. RFD 064 accepted this as a temporary deviation from the append-only principle, blocked on the projection layer that has since landed.

A repair is now appended as an `event_overlay`, and applied when the stream is projected to build a request. Stored events keep their metadata. Matching is by metadata value rather than position, so an overlay is independent of where it sits and stays correct when compaction collapses turns or pruning removes them.

The persisted vocabulary is mirrored in `jp_conversation` rather than shared from `jp_llm`: the conversation crate owns what is stored, and cannot depend on the LLM crate to say what a stored patch means.

`apply_patches` becomes `record_patches` and reports how many events the patches change in the projection rather than how many it rewrote, accounting for overlays already recorded so an idempotent repair still reports no progress.

### DIFF
--- a/crates/jp_cli/src/cmd/conversation/summarize.rs
+++ b/crates/jp_cli/src/cmd/conversation/summarize.rs
@@ -11,7 +11,7 @@ use jp_conversation::{
 };
 use jp_llm::{
     Provider,
-    event::{Event, EventPatch, FinishReason, apply_patches},
+    event::{Event, EventPatch, FinishReason, record_patches},
     event_builder::EventBuilder,
     model::ModelDetails,
     provider,
@@ -206,7 +206,7 @@ async fn summarize_stream(
         // The patches target events in the local range stream, so applying them
         // here leaves the stored conversation untouched: repairing the user's
         // history is the query loop's job, not a side effect of summarizing.
-        let applied = apply_patches(&mut stream, &patches);
+        let applied = record_patches(&mut stream, &patches);
         if applied == 0 {
             return Err(Error::Summarize {
                 model: model_id.to_string(),
--- a/crates/jp_cli/src/cmd/conversation/summarize.rs
+++ b/crates/jp_cli/src/cmd/conversation/summarize.rs
@@ -57,11 +57,7 @@ pub async fn generate_summary(
     // is reused for provider lookup below.
     let model_id = model.id.resolved().clone();
 
-    let range_events = collect_range_events(events, range_from, range_to);
-
-    // Rebuild a clean stream with just the range events.
-    let mut stream = ConversationStream::new(events.base_config());
-    stream.extend(range_events);
+    let mut stream = build_range_stream(events, range_from, range_to);
 
     // Override the full assistant model (id plus parameters) so a
     // summary-specific model can also set max tokens, temperature, reasoning,
@@ -324,6 +320,28 @@ fn failure_reason(finish: Option<&FinishReason>) -> String {
         Some(FinishReason::Retry) => "the provider asked to rebuild the request".to_owned(),
         Some(FinishReason::Completed) | None => "the model returned an empty response".to_owned(),
     }
+}
+
+/// Build the throwaway stream a summary request is made from.
+///
+/// Holds the range's conversation events plus the source's patch overlays.
+/// The overlays repair metadata a provider already rejected; without them this
+/// request replays the stale metadata and pays for the same repair again.
+/// Overlays match by value, so copying all of them is safe: one whose target
+/// falls outside the range matches nothing.
+fn build_range_stream(
+    events: &ConversationStream,
+    range_from: usize,
+    range_to: usize,
+) -> ConversationStream {
+    let mut stream = ConversationStream::new(events.base_config());
+    stream.extend(collect_range_events(events, range_from, range_to));
+
+    for overlay in events.overlays() {
+        stream.add_overlay(overlay.patches.clone());
+    }
+
+    stream
 }
 
 /// Collect all events in the inclusive turn range `[range_from, range_to]`.
--- a/crates/jp_cli/src/cmd/query/interrupt/signals.rs
+++ b/crates/jp_cli/src/cmd/query/interrupt/signals.rs
@@ -154,30 +154,33 @@ pub fn handle_llm_event(
     retry_state: &mut StreamRetryState,
 ) -> (LoopAction, CommittedEvent) {
     // `Patch` is a side-channel instruction from the provider to fix bad events
-    // in the conversation stream. This can be handled directly instead of
-    // passing through the turn coordinator.
+    // in the conversation. This can be handled directly instead of passing
+    // through the turn coordinator.
     if let Event::Patch(patches) = event {
         let count = record_patches(conversation_stream, &patches);
-        let shrinks = patches.iter().all(|patch| patch.action.shrinks_stream());
+        let shrinks = patches
+            .iter()
+            .all(|patch| patch.action.shrinks_projection());
         retry_state.record_patch(count, shrinks);
 
         if !shrinks {
-            // A patch set that can grow the stream gives no guarantee the repair
-            // loop ends, so the rebuild below is refused whatever it changed.
+            // A patch set that can grow the projection gives no guarantee the
+            // repair loop ends, so the rebuild below is refused whatever it
+            // changed.
             tracing::warn!(
                 patches = patches.len(),
-                "History patches include an action that may not shrink the conversation stream."
+                "History patches include an action that may not shrink the projected conversation."
             );
         }
 
         if count > 0 {
-            tracing::debug!(count, "Applied history patches to conversation stream.");
+            tracing::debug!(count, "Recorded history patch overlay.");
         } else {
             // The rebuilt request would be byte-identical, so the rebuild that
             // follows is refused below rather than resent.
             tracing::warn!(
                 patches = patches.len(),
-                "History patches matched no events in the conversation stream."
+                "History patches change no events in the projected conversation."
             );
         }
 
--- a/crates/jp_cli/src/cmd/query/interrupt/signals.rs
+++ b/crates/jp_cli/src/cmd/query/interrupt/signals.rs
@@ -14,7 +14,7 @@ use jp_config::interrupt::{StreamingInterruptConfig, ToolInterruptConfig};
 use jp_conversation::ConversationStream;
 use jp_editor::EditorBackend;
 use jp_inquire::{ReplyEditMode, prompt::PromptBackend};
-use jp_llm::event::{Event, FinishReason, apply_patches};
+use jp_llm::event::{Event, FinishReason, record_patches};
 use jp_printer::Printer;
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, info, trace};
@@ -157,7 +157,7 @@ pub fn handle_llm_event(
     // in the conversation stream. This can be handled directly instead of
     // passing through the turn coordinator.
     if let Event::Patch(patches) = event {
-        let count = apply_patches(conversation_stream, &patches);
+        let count = record_patches(conversation_stream, &patches);
         let shrinks = patches.iter().all(|patch| patch.action.shrinks_stream());
         retry_state.record_patch(count, shrinks);
 
--- a/crates/jp_conversation/src/patch.rs
+++ b/crates/jp_conversation/src/patch.rs
@@ -1,0 +1,105 @@
+//! Event patch overlays.
+//!
+//! A provider can discover that events already in the stream carry metadata it
+//! will not accept back — most often a stale reasoning signature.
+//! Rather than rewrite those events, the repair is recorded as an overlay
+//! appended to the stream, and applied when the stream is projected for a
+//! request.
+//!
+//! The stored history stays byte-identical, so a repair made for one provider
+//! or model does not destroy metadata another one could still use.
+//!
+//! These types mirror the provider-facing vocabulary in `jp_llm`.
+//! The duplication is deliberate: the conversation crate owns what is
+//! persisted, and must not depend on the LLM crate to say what a stored patch
+//! means.
+
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
+
+/// An appended instruction to rewrite how earlier events are projected.
+///
+/// Overlays never modify or remove the events they target.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct EventOverlay {
+    /// When the overlay was recorded.
+    pub timestamp: DateTime<Utc>,
+
+    /// Patches to apply, in order.
+    pub patches: Vec<OverlayPatch>,
+}
+
+/// A single matcher/action pair.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct OverlayPatch {
+    /// Which events to target.
+    pub matcher: OverlayMatcher,
+
+    /// What to change about them.
+    pub action: OverlayAction,
+}
+
+/// Identifies which events a patch applies to.
+///
+/// Matching is by content rather than position, so an overlay is independent of
+/// where it sits in the stream and stays correct when earlier events are
+/// trimmed or compacted.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "matcher", rename_all = "snake_case")]
+#[non_exhaustive]
+pub enum OverlayMatcher {
+    /// Match events whose `metadata[key]` is the string `value`.
+    MetadataValue {
+        /// Metadata key to compare.
+        key: String,
+        /// Value the key must hold.
+        value: String,
+    },
+}
+
+/// The change a patch makes to a matched event's projection.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "action", rename_all = "snake_case")]
+#[non_exhaustive]
+pub enum OverlayAction {
+    /// Drop a metadata key from the projected event.
+    RemoveMetadata {
+        /// Metadata key to drop.
+        key: String,
+    },
+}
+
+impl OverlayPatch {
+    /// Whether this patch targets `metadata`, and would change it.
+    ///
+    /// A matcher that hits an event which does not carry the key the action
+    /// removes is not a change: matching alone is not progress.
+    #[must_use]
+    pub fn would_change(&self, metadata: &serde_json::Map<String, serde_json::Value>) -> bool {
+        let matched = match &self.matcher {
+            OverlayMatcher::MetadataValue { key, value } => metadata
+                .get(key)
+                .and_then(serde_json::Value::as_str)
+                .is_some_and(|v| v == value),
+        };
+
+        if !matched {
+            return false;
+        }
+
+        match &self.action {
+            OverlayAction::RemoveMetadata { key } => metadata.contains_key(key),
+        }
+    }
+
+    /// Apply this patch to `metadata`, returning whether it changed anything.
+    pub fn apply(&self, metadata: &mut serde_json::Map<String, serde_json::Value>) -> bool {
+        if !self.would_change(metadata) {
+            return false;
+        }
+
+        match &self.action {
+            OverlayAction::RemoveMetadata { key } => metadata.remove(key).is_some(),
+        }
+    }
+}
--- a/crates/jp_conversation/src/patch.rs
+++ b/crates/jp_conversation/src/patch.rs
@@ -23,6 +23,13 @@
 //! content, [`EventOverlay`] can gain an optional scope without changing what
 //! is already stored.
 //!
+//! The nested [`OverlayMatcher`] and [`OverlayAction`] vocabularies are read by
+//! exact tag: a build that meets a tag it does not know fails the conversation
+//! load rather than skipping the event.
+//! An addition older builds cannot understand therefore ships under a new
+//! top-level `type` tag, where the unknown-event path preserves it verbatim and
+//! round-trips it on save.
+//!
 //! These types mirror the provider-facing vocabulary in `jp_llm`.
 //! The duplication is deliberate: the conversation crate owns what is
 //! persisted, and must not depend on the LLM crate to say what a stored patch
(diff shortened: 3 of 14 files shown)


crates/jp_llm/src/event.rs
```diff
@@ -202,53 +200,41 @@ impl PatchAction {
...
 ///
-/// A return value of `0` means `stream` is byte-identical to what it was, so
-/// resending a request built from it would hit the same provider complaint.
-/// A patch whose matcher hits an event that does not carry the field the action
-/// removes counts as no change: matching alone is not progress.
+/// The patches are appended as an overlay: the stored events keep their
+/// metadata, and the removal happens when the stream is projected to build a
+/// request.
+/// A repair made for one provider therefore does not destroy metadata another
+/// provider or model could still use, and the conversation stays an append-only
```
JeanMertz: blocking: A user can generate signed reasoning with Anthropic model A, switch this conversation to model B (which rejects A's signature and records this repair), then switch back to A. Every provider calls `Thread::into_parts()`, which invokes `apply_projection`; `apply_overlays` has no provider or model context and removes the signature for A as well. Keeping the raw metadata on disk does not make it available to another model, and the downgrade to `<think>` text silently persists across future requests. Carry the rejecting provider/model as overlay scope and apply it only for matching requests, or otherwise make projection provider-specific.

crates/jp_conversation/src/stream/projection.rs
```diff
@@ -56,6 +56,38 @@ impl TurnOrigin {
     }
 }
 
+/// Apply every patch overlay to the events it matches, then drop the overlays.
+///
+/// Runs before compaction so both the compacting and non-compacting paths see
+/// patched metadata.
+/// Matching is by metadata value rather than position, so applying every
+/// overlay to every event is equivalent to applying each one to the events that
+/// preceded it, and stays correct when compaction reorders or replaces turns.
```
JeanMertz: theoretical: Matching by value is not generally equivalent to predecessor-only application. This implementation also modifies events after the overlay, so a later event with the same `(key, value)` loses metadata because of an earlier repair. I could not identify a normal current flow that reuses an opaque Anthropic or Google signature, so this is theoretical today. Either activate patches while walking the stream in order, or make all-time matching an explicit contract and constrain matcher values to identifiers guaranteed unique.

crates/jp_conversation/src/stream/projection.rs
```diff
@@ -107,6 +139,8 @@ struct ResolvedSummary {
 ///
 /// [`Compaction`]: crate::Compaction
 pub(super) fn apply(events: &mut Vec<InternalEvent>) -> Vec<TurnOrigin> {
+    apply_overlays(events);
```
JeanMertz: nit: This makes the public `ConversationStream::apply_projection` contract stale. Its docs say the method reads compaction overlays and leaves events unchanged when no compaction exists, but an overlay-only stream changes metadata and consumes the overlay here. The module preamble and the glossary's `Compacted View` definition also describe projection as compaction-only. Please update those descriptions and add `Event Overlay` to the ubiquitous language so the two overlay concepts remain distinct.

crates/jp_cli/src/cmd/query/interrupt/signals.rs
```diff
@@ -157,7 +157,7 @@ pub fn handle_llm_event(
     // in the conversation stream. This can be handled directly instead of
     // passing through the turn coordinator.
     if let Event::Patch(patches) = event {
-        let count = apply_patches(conversation_stream, &patches);
+        let count = record_patches(conversation_stream, &patches);
```
JeanMertz: nit: The termination measure is the projected request, not the conversation stream. `record_patches` grows the stored stream by appending an overlay, but `PatchAction::shrinks_stream`, `StreamRetryState::record_patch`, the warning below, and the debug message still describe shrinking or applying changes to the stream. Please reword these around shrinking the projection and recording an overlay so the retry invariant and diagnostics match the new behavior.

JeanMertz: Taken. Every citation was accurate, and the inconsistency you're pointing at is that this PR renamed `apply_patches` → `record_patches` for exactly this reason and then left its neighbours describing the model it replaced. A reader tracing the termination argument from `authorize_rebuild` back through `record_patch` to `shrinks_stream` was told three times that the property concerns a structure which no longer has it.

Renamed and reworded:

- `PatchAction::shrinks_stream` → `shrinks_projection`, and its doc now says why: "The measure is what the projection contains, not the stored stream: a patch is recorded as an appended overlay, so the stored stream only ever grows."
- `StreamRetryState::patch_changed_stream` → `patch_changed_projection`; `patch_sets_shrink`'s doc drops "the stream"; `record_patch`'s doc now reads "how many events it changes in the projection".
- `signals.rs`: the warning is "…may not shrink the projected conversation", the no-op warning is "…change no events in the projected conversation", and the debug is "Recorded history patch overlay."

No behavior change; `cargo check` found the call sites. Grepped afterwards to confirm nothing stale is left.

crates/jp_conversation/src/stream/projection.rs
JeanMertz: Right on the mechanism, and right that it's theoretical. I took your second option, minus one half.

What actually needed fixing was the doc comment rather than the code. It claimed:

> applying every overlay to every event is equivalent to applying each one to the events that preceded it

That equivalence holds only if `(key, value)` is unique across the stream, and the comment asserted it unconditionally — so the code was defended by an argument that omitted its own precondition. Now:

```
/// Matching is by metadata value and spans the whole stream, not only the events
/// recorded before the overlay. That is deliberate: an overlay has to survive
/// compaction replacing turns and pruning removing them, so it cannot be
/// anchored to a position.
/// The consequence is that an event recorded later carrying the same
/// `(key, value)` is patched too. The values matched are opaque provider
/// signatures, unique per reasoning block, so in practice an overlay recorded
/// for one event cannot collide with another.
```

I declined ordered activation: anchoring to a position is the thing the overlay has to avoid, since compaction can replace the turn an overlay was recorded in and pruning can remove it.

I also declined the "constrain matcher values to identifiers guaranteed unique" half. `OverlayMatcher::MetadataValue` takes a `String` because it mirrors `EventMatcher`; a uniqueness-carrying newtype is a type change across two crates to enforce a property no current caller can violate. Adding your reachability note to the collision argument: it needs a provider to emit a byte-identical signature for two distinct reasoning blocks, and nothing in the repair path copies metadata between events.

crates/jp_conversation/src/stream/projection.rs
JeanMertz: Both halves confirmed and fixed.

The second sentence was the worse of the two — "When no compaction events are present the events are left unchanged" is now falsified outright by an overlay-only stream, since `apply_overlays` runs first, rewrites metadata, and `retain`s the overlays away. `apply_projection` now describes both passes and says "When the stream carries neither patch overlays nor compactions…". The `projection.rs` preamble lists the two passes in order and notes why patches run first.

Glossary: **Compacted View** now reads "every Compaction and Event Overlay applied", and there's a new **Event Overlay** entry.

Your parenthetical about keeping the two overlay concepts distinct was the part that shaped the entry. The glossary already calls a Compaction "a non-destructive overlay", so "overlay" was taken — building a second term from it needs the disambiguation stated or it breaks the one-word-one-concept rule. Hence the "Not the same as" line: an Event Overlay changes the metadata of individual events; a Compaction reduces what a range of turns contributes.

One thing I got wrong twice while doing it: the entry is alphabetised between `EditorBackend` and `InlineReply` (`Edi` < `Eve`), and I initially placed it after `Compaction` and then before `EditorBackend`. Fixed, and the TOC matches.

crates/jp_llm/src/event.rs
JeanMertz: The plumbing description is exactly right, and the criticism of the PR's own claim lands — but I don't think the scenario is a loss, and I've documented the decision rather than scoping the overlay.

**Where you're right.** The description says a repair "does not destroy metadata another one could still use". That isn't what the code does: `Thread::into_parts()` calls `apply_projection()` unconditionally with no model context, so no other model or provider can use the preserved metadata either. I've corrected the module docs; the honest guarantee is that stored history stays intact for inspection, export, and later features — not that another model benefits this run.

**Where the scenario doesn't hold.** It needs step 3 — model B *rejects* A's signature. Measured against the live API:

- `opus-5` replaying its own prior thinking: 832 tokens vs 639 stripped — 193 billed, so it retains and reads them.
- `sonnet-5` replaying `opus-5`'s blocks: 707 vs 707 — zero difference, silently ignored.
- Real request to `claude-haiku-4-5` carrying `opus-5` thinking blocks: **200 OK**.

Anthropic's docs agree: "Other models silently ignore them rather than rejecting the request." No rejection means no `Event::Patch`, no overlay, nothing lost. One provider pair and one fixture, so not a proof — but it's the only evidence either way and it points against the scenario.

**Where scoping would hurt.** The trigger we've actually observed is a hand-edited stream: content changed underneath a signature, which invalidates it for *every* Anthropic model permanently. Scoping to the recording model means a conversation repaired under Opus 5 re-wedges the moment you switch away and back — reintroducing the failure this work exists to fix.

There's also a bound you didn't price: `OverlayMatcher::MetadataValue` matches a provider-specific key. An Anthropic repair strips `anthropic_thinking_signature`; Google reads `google_thought_signature` and is unaffected by construction. The cross-*provider* concern is nil; only cross-model-within-Anthropic is live, which is the case measured above.

And the cost: threading provider/model through `into_parts()` touches every provider, and recording it on `EventOverlay` puts provider identity into `jp_conversation` — the boundary the mirror exists to protect.

So: documented in `patch.rs` why unconditional is correct for the observed trigger, and noted that `EventOverlay` can gain an optional scope without changing what's already stored if a rejection ever proves model-specific rather than content-specific. Happy to reopen if you can construct a rejection that one Anthropic model raises and another doesn't — that would falsify the measurement above and change the call.

crates/jp_conversation/src/stream.rs
```diff
@@ -1486,7 +1566,9 @@ impl Iterator for IntoIter {
                         config: self.current_config.clone(),
                     });
                 }
-                InternalEvent::Compaction(_) | InternalEvent::Unknown(_) => {}
+                InternalEvent::Compaction(_)
+                | InternalEvent::Overlay(_)
```
JeanMertz: nit: A user who forks a repaired conversation gets the rejected metadata back. `fork_conversation` copies the source with `events.extend(new_events)`, which consumes this iterator; skipping `Overlay` drops every recorded repair while the copied raw events still carry their stale signatures. The fork's next query therefore repeats the provider rejection and rebuild, and a source with many repaired signatures can hit the rebuild cap. Preserve patch overlays when copying a full stream into a fork, for example through a full-stream replacement/copy API, and add a fork test whose source contains a recorded repair.

crates/jp_cli/src/cmd/conversation/summarize.rs
```diff
@@ -206,7 +206,7 @@ async fn summarize_stream(
         // The patches target events in the local range stream, so applying them
         // here leaves the stored conversation untouched: repairing the user's
         // history is the query loop's job, not a side effect of summarizing.
-        let applied = apply_patches(&mut stream, &patches);
+        let applied = record_patches(&mut stream, &patches);
```
JeanMertz: nit: Summarizing a range from an already repaired conversation repeats the repair. `generate_summary` rebuilds this local stream from `collect_range_events`, which copies only `ConversationEvent`s, so existing `EventOverlay`s are absent and the stale raw signatures reach the provider again. This line keeps the operation functional, but every later compaction over that range can incur the same rejection rounds. Build the raw range from a clone that retains patch overlays while dropping compactions, and cover a summary request made after a repair has been recorded.

crates/jp_conversation/src/patch.rs
```diff
@@ -0,0 +1,119 @@
...
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
+
+/// An appended instruction to rewrite how earlier events are projected.
+///
+/// Overlays never modify or remove the events they target.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct EventOverlay {
+    /// When the overlay was recorded.
+    pub timestamp: DateTime<Utc>,
```
JeanMertz: nit: A user editing `events.json` can naturally copy the timestamp format used by every other stored event (`2026-08-27 12:00:00.0`), but this field uses chrono's default serde and rejects that format. It also writes RFC 3339 while `ConversationEvent` and `Compaction` use the crate's storage format. Apply `serialize_dt`/`deserialize_dt` here as well, and make the roundtrip test include the storage-format timestamp rather than only roundtripping this field's own output.

crates/jp_conversation/src/patch.rs
```diff
@@ -0,0 +1,119 @@
...
+
+/// Identifies which events a patch applies to.
+///
+/// Matching is by content rather than position, so an overlay is independent of
+/// where it sits in the stream and stays correct when earlier events are
+/// trimmed or compacted.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "matcher", rename_all = "snake_case")]
+#[non_exhaustive]
+pub enum OverlayMatcher {
```
JeanMertz: theoretical: A conversation written by a future JP with a new matcher or action variant will fail to load in this build. `InternalEvent` dispatches every `event_overlay` directly through these serde enums, so an unknown nested tag errors before the raw `Unknown` preservation path can run. I cannot name a current producer because these are the only variants today. If `#[non_exhaustive]` is intended to permit persisted evolution, preserve overlays with unknown nested tags as raw events, or use a new top-level event type for incompatible additions, and add a downgrade roundtrip test.

crates/jp_llm/src/event.rs
JeanMertz: The API behavior supplied here removes the trigger from the named model-switch scenario: without a rejection, no overlay is recorded. I no longer consider that scenario merge-gating. One documentation mismatch remains at `event.rs:212-213`: it still says another provider or model can use the preserved metadata, although unconditional projection removes it before every provider. That is a wording nit, not the behavioral blocker described above.

crates/jp_conversation/src/stream/projection.rs
JeanMertz: `apply_overlays` documents the all-time matching contract, but `patch.rs:34` still says "earlier events" and `stream.rs:55` says "preceding events." Those descriptions conflict with the deliberate later-event behavior. Please describe matched events without the positional qualifier in both places.

crates/jp_conversation/src/stream/projection.rs
JeanMertz: One nearby glossary sentence remains stale: lines 103-106 still define the `jp_conversation::stream::projection` meaning as "applying compaction overlays." It should say conversation overlays, or patch and compaction overlays, to match the definition above.

crates/jp_cli/src/cmd/conversation/summarize.rs
JeanMertz: Confirmed and fixed — and this was a regression the PR introduced, not a pre-existing gap. Before the overlay migration the repair mutated the stored events, so `collect_range_events` copied already-repaired events and the summary request was clean. Now the events keep their metadata and the repair lives beside them, so every compaction over a repaired range re-ran the whole rejection cycle.

Rather than add the overlay copy inline, I extracted the construction so the tested path is the production one:

```rust
fn build_range_stream(
    events: &ConversationStream,
    range_from: usize,
    range_to: usize,
) -> ConversationStream {
    let mut stream = ConversationStream::new(events.base_config());
    stream.extend(collect_range_events(events, range_from, range_to));

    for overlay in events.overlays() {
        stream.add_overlay(overlay.patches.clone());
    }

    stream
}
```

The mock provider can't discriminate this — it doesn't validate signatures, so request count is identical either way — which is why the assertion is on the projected stream instead: `range_stream_carries_the_source_repairs` projects the range stream and asserts no `signature` metadata survives. Reverting the loop fails it with `stale signature reaches the summary request: [String("stale")]`.

Copying every overlay rather than filtering to the range is deliberate, and now stated in the doc comment: they match by value, so one whose target falls outside the range matches nothing.

New `overlays()` accessor on `ConversationStream`, mirroring the existing `compactions()`.

crates/jp_conversation/src/patch.rs
JeanMertz: Right on both counts, fixed:

```rust
#[serde(
    serialize_with = "crate::serialize_dt",
    deserialize_with = "crate::deserialize_dt"
)]
pub timestamp: DateTime<Utc>,
```

The asymmetry was worse than "inconsistent". `deserialize_dt` accepts both spellings (`lib.rs:95-102` tries `%Y-%m-%d %H:%M:%S%.f`, then RFC 3339) while chrono's default accepts only RFC 3339 — so every other event in `events.json` tolerated either and the overlay tolerated one, and it was the one that appears nowhere else in the file. Since overlays are recorded precisely on conversations someone has hand-edited, copying the timestamp format from the line above broke the load.

`test_internal_event_overlay_timestamp_accepts_storage_format` feeds `"2026-08-27 12:00:00.0"` in and asserts both the parse and the emitted form. Without the attribute it fails at the parse with `Error("premature end of input")` — the whole stream load, as you said, not just the field.

Backward-compatible: overlays already written in RFC 3339 still parse, because `deserialize_dt` accepts that form too.

I also checked whether the one-directional attribute was a wider pattern, since `stream.rs:241` and `:254` use `serialize_with` with no counterpart. They're fine — `ApplyDelta`/`ResetDelta` derive only `Serialize` and are read by the hand-written `deserialize_config_delta`, which calls `crate::parse_dt` directly. So there's no `Deserialize` to attach one to, and this was genuinely local to `EventOverlay`.

crates/jp_conversation/src/patch.rs
JeanMertz: Mechanism confirmed exactly: `event_overlay` is a known top-level tag, so `stream.rs:2048-2052` dispatches straight to `serde_json::from_value` and the `Unknown` preservation path at 2058-2070 — guarded by `!EventKind::TYPE_TAGS.contains(&tag)` — is unreachable for it.

I've taken your third option, and documented it in the `patch.rs` module docs rather than adding machinery:

> The nested `OverlayMatcher` and `OverlayAction` vocabularies are read by exact tag: a build that meets a tag it does not know fails the conversation load rather than skipping the event. An addition older builds cannot understand therefore ships under a new top-level `type` tag, where the unknown-event path preserves it verbatim and round-trips it on save.

Reasoning for not fixing it structurally now. Preserving an unknown nested tag needs the overlay to round-trip *verbatim*, which means carrying the original JSON alongside the typed value. Falling back to `Unknown` on any deserialize error is simpler but conflates version skew with corruption, and the comment at 2054-2057 separates those deliberately: "Corrupt *known* events still fail loudly below." A `#[serde(other)]` unit variant is worse again — it parses and then silently drops the unknown patch on the next save, which for an append-only log is the worst of the three.

One correction to your framing: `#[non_exhaustive]` is a Rust-API attribute about downstream matching and carries no serde meaning, so it isn't evidence of intent about persisted evolution either way. You posed that as a question rather than a claim, which was the right read.

Worth noting `compaction` at 2042-2046 has the identical shape, so this is a property of the dispatch design rather than something this PR introduced. If we ever want real nested forward-compat it belongs in one change covering both, which I'd rather file separately than bolt onto this.

crates/jp_conversation/src/stream.rs
JeanMertz: Confirmed, and fixed with the full-stream copy you suggested:

```rust
pub fn append_stream(&mut self, other: Self) {
    self.events.extend(other.events);
}
```

`fork_conversation` now calls that instead of `extend`.

Two things worth recording. First, compactions were being dropped too, and were before this PR — so the underlying bug is that `IntoIter` implements *iteration over conversation events* and using it to *copy a stream* silently discards everything global. Fixing it at that level covers both rather than special-casing the harmful one.

Second, the consequences differ sharply, which is why this one matters: losing a compaction only makes the fork see more history than the source, but losing a patch overlay replays metadata the provider already rejected, so the fork is wedged from its first query and has to re-run the entire repair. A source that needed six rounds costs the fork six rounds again.

`fork_inherits_patch_overlays` drives the real `fork_conversation` and asserts the fork's stream still carries the overlay. Reverting to `extend` fails it with `left: 0, right: 1`. `test_append_stream_preserves_overlays_and_compactions` covers the primitive directly, and its second half characterises what `Extend` does with the same input, so the distinction stays pinned if someone reaches for the wrong one later.

The `Extend`-versus-`append_stream` choice is documented on the new method: `Extend` recomputes config deltas and suits moving events between streams whose config state differs; `append_stream` appends verbatim and requires a shared base config, which the fork satisfies since the target is created from `new_events.base_config()`.

crates/jp_conversation/src/stream/projection.rs
JeanMertz: Both fixed — `patch.rs:34` and `stream.rs:55` now say "matched events".

You're right that this was the same failure as the contract fix it followed: I wrote the all-time behaviour into `apply_overlays` last round and left two summaries asserting the opposite, so a reader meeting either doc first would take away exactly the wrong model.

I left `stream.rs:51-52` alone. A `Compaction` genuinely does cover a preceding turn range, so "preceding" is accurate there, and making the two variants read identically would erase a real difference between them.

crates/jp_conversation/src/stream/projection.rs
JeanMertz: Fixed — "applying conversation overlays (`jp_conversation::stream::projection`)".

I updated the **Compacted View** definition three lines above and left the disambiguation paragraph beneath it describing the old behaviour, which is the third instance of the same pattern this round.

I went with "conversation overlays" rather than listing both kinds deliberately: that paragraph exists to disambiguate the word *projection*, and enumerating the overlay types there would quietly turn it into a second definition competing with the **Event Overlay** entry.

crates/jp_llm/src/event.rs
JeanMertz: Thanks for retiring the blocking part explicitly — and the wording nit is a fair catch, since it's the same claim I corrected in `patch.rs` last round left standing in the function that records the overlay. Now:

> The conversation therefore stays an append-only log and the original metadata remains available for inspection and export, but the removal itself applies to every request built from this conversation, whatever provider or model it targets. See `jp_conversation::patch` for why that is unconditional.

That states what the code guarantees and points at the reasoning rather than repeating it, so the two can't drift apart again.

PR #1041, opened 2026-08-26
fix(llm, anthropic): Downgrade whole turn of rejected thinking block

JeanMertz:
A rejection was answered by rewriting one thinking block, or every block in the single message holding it. Anthropic validates a turn as one logical assistant message, so a partial rewrite is refused with `thinking blocks ... cannot be modified`, and a turn spanning several of our messages needed one round trip per message. A conversation with six damaged turns exhausted the rebuild budget after eleven rounds without repairing the oldest one.

A rejected block now downgrades every thinking and redacted-thinking block in the assistant turn that holds it, which is the unit the API validates: one round per damaged turn. This applies to any turn rather than only the newest, because models that retain prior turns' thinking validate those turns too and refuse a partial rewrite anywhere.

An empty reasoning event left behind by a repair is dropped rather than sent as an empty `<think></think>` pair. Anthropic can return a thinking block carrying a signature and no text, and once the signature is stripped there is nothing left to replay.

### DIFF
--- a/crates/jp_llm/src/provider/anthropic.rs
+++ b/crates/jp_llm/src/provider/anthropic.rs
@@ -854,14 +854,18 @@ fn is_tool_result_message(msg: &types::Message) -> bool {
 /// On the next request rebuild, [`convert_event`] will fall through to the
 /// `<think>` tag path for events that no longer carry a signature.
 ///
-/// A block anywhere in the final assistant turn downgrades every thinking and
-/// redacted-thinking block in the message holding it.
-/// Anthropic treats that whole turn as one assistant message and requires its
+/// A rejected block downgrades every thinking and redacted-thinking block in
+/// the whole assistant turn holding it.
+/// Anthropic treats that turn as one assistant message and requires its
 /// thinking blocks back exactly as generated, so rewriting a subset is rejected
-/// with `thinking blocks ... cannot be modified`; rewriting all of a message's
-/// blocks leaves text and tool use, which the API accepts.
-/// Blocks in earlier turns are downgraded individually, keeping the rest of
-/// their reasoning native.
+/// with `thinking blocks ... cannot be modified`; rewriting all of them leaves
+/// text and tool use, which the API accepts.
+///
+/// This applies to any turn, not only the newest.
+/// Models that retain prior turns' thinking validate those turns too, so a
+/// partial rewrite anywhere is refused.
+/// The cost is the rest of that turn's reasoning, which buys a repair that
+/// converges in one round per broken turn instead of one round per message.
 ///
 /// When the error carries no usable position, [`fallback_thinking_block`] picks
 /// where to look based on `rejection`.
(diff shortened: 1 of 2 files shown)


crates/jp_llm/src/provider/anthropic.rs
```diff
@@ -882,18 +886,9 @@ fn build_thinking_patches(
...
-        last_assistant_turn(&request.messages).is_some_and(|turn| turn.contains(&msg_idx));
-
-    let content_indices = if in_final_turn {
-        thinking_block_indices(request, msg_idx)
-    } else {
-        vec![content_idx]
-    };
-
-    let patches: Vec<EventPatch> = content_indices
+    let patches: Vec<EventPatch> = turn_thinking_blocks(request, msg_idx)
```
JeanMertz: theoretical: The fallback below this block still returns a single `thinking_patch`. If a parseable API position resolves into a turn with no native thinking blocks, but `fallback_thinking_block` finds several blocks in another turn, the repair loses this whole-turn invariant and can require another rebuild. I cannot name a real Anthropic rejection that reports a resolvable non-thinking position, so this does not gate the fix. Consider running the fallback target through `turn_thinking_blocks` too, or documenting that such a position is impossible and removing the branch.

JeanMertz: Fixed by collapsing the two paths rather than special-casing the fallback.

`build_thinking_patches` now picks a **message** first and derives patches once:

```rust
let msg_idx = api_position
    .and_then(|(turn, flat)| resolve_turn_position(&request.messages, turn, flat))
    .map(|(msg_idx, _)| msg_idx)
    // A position can resolve into a turn that holds no native thinking, if
    // our turn model disagrees with the one the API reports against. Treat
    // that like an unusable position rather than patching nothing.
    .filter(|&msg_idx| !turn_thinking_blocks(request, msg_idx).is_empty())
    .or_else(|| fallback_thinking_block(request, rejection).map(|(msg_idx, _)| msg_idx))?;
```

Every route now ends in the same `turn_thinking_blocks` call, so a partial rewrite is structurally impossible rather than avoided by the branch you happen to take. Net −4 lines.

I took your first option over the second. Removing the fallback outright would also drop recovery for a resolvable-but-wrong position, and that's the case I'd least like to lose — `resolve_turn_position` returns `Some` whenever the flat index lands inside the turn's content, without checking that the content *is* a thinking block, so it's the most plausible way our turn model and Anthropic's could diverge.

On reachability: agreed, and I couldn't name a triggering input either. Across the eleven rounds in the field trace every position resolved to a real thinking block, and the `2n+1` turn mapping held against six independent rejections. The damage was also contained and self-healing — one extra round, visible in the logs, well inside the cap. By our severity rules that's a suggestion, not a gate.

I took it anyway because the fix is subtractive. It deletes a divergence rather than adding a guard, so it doesn't carry the usual cost of acting on a theoretical finding.

Test: `position_resolving_into_a_thinkingless_turn_still_downgrades_a_whole_turn`. A position resolving to a text-only turn while the final turn carries two signed blocks. Verified against the old shape first — it returned `[sig_a]`, the exact single patch you described.

Also folded in: the `debug!` now logs `patches.len()` instead of a `content_idx` that no longer exists, which is the more useful field anyway since the count is what tells you whether the whole-turn widening fired.

`jp_llm` 649/649, workspace 5024/5024, `cargo check` clean.

PR #1043, opened 2026-08-28
chore(tools): Add redirected roots to Git tools

JeanMertz:
Git and Cargo tools can run in a configured workspace-relative root, with capability checks and clear failures for invalid project roots.

Cargo tools also accept a `profile` option, keeping their build artifacts separate from a developer's concurrent default-profile builds.

### DIFF
--- a/.config/jp/tools/src/cargo.rs
+++ b/.config/jp/tools/src/cargo.rs
@@ -1,3 +1,4 @@
+use camino::Utf8Path;
 use jp_tool::Capability;
 use serde_json::Value;
 
@@ -6,6 +7,7 @@ use crate::{
     util::{
         ToolResult, error,
         root::{CARGO_MANIFEST, configured_root, note_root, resolve_root},
+        runner::{DuctProcessRunner, ProcessOutput, ProcessRunner},
         unknown_tool,
     },
 };
@@ -58,6 +60,12 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
         Err(message) => return error(message),
     };
 
+    if root != ctx.root
+        && let Err(message) = ensure_workspace_root(&root, &DuctProcessRunner)
+    {
+        return error(message);
+    }
+
     // Cargo profile to build under, set via `options.profile`. A profile of its
     // own gives these tools their own directory under `target/`, which is the
     // only way to keep their artifacts, fingerprints and build lock apart from
@@ -74,6 +82,16 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
             ));
         }
     };
+
+    // `format` and `update` never compile, and `install_tools` builds through a
+    // recipe that pins its own profile. Accepting the option for them would
+    // promise a target directory of its own and hand back the shared one, which
+    // is the outcome the option exists to avoid.
+    if profile.is_some() && !matches!(subcommand, "check" | "expand" | "test") {
+        return error(format!(
+            "The `profile` tool option has no effect on `cargo_{subcommand}`."
+        ));
+    }
 
     let outcome = match subcommand {
         "check" => {
@@ -117,6 +135,69 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
     }
 
     note_root(outcome, &root, "cargo")
+}
+
+/// Refuse a redirected root that cargo treats as a member of a larger
+/// workspace.
+///
+/// The manifest check that precedes this one only proves a `Cargo.toml` is
+/// present, and a member has one too.
+/// Cargo resolves the workspace from that manifest, so a member root puts
+/// `cargo check` and `cargo test` on `--workspace`, `cargo fmt` on `--all`, and
+/// `cargo update` on the enclosing lockfile — reaching a project the caller
+/// never named and the access policy never covered.
+///
+/// # Errors
+///
+/// Returns an error naming both directories when they differ, and when cargo
+/// cannot resolve the workspace at all.
+fn ensure_workspace_root<R: ProcessRunner>(root: &Utf8Path, runner: &R) -> Result<(), String> {
+    // Cargo is asked rather than the manifest parsed, so this answer cannot
+    // drift from the one the build commands themselves will resolve.
+    let ProcessOutput {
+        stdout,
+        stderr,
+        status,
+    } = runner
+        .run(
+            "cargo",
+            &[
+                "locate-project",
+                "--workspace",
+                "--message-format=plain",
+                "--manifest-path",
+                root.join("Cargo.toml").as_str(),
+            ],
+            root,
+        )
+        .map_err(|error| format!("Could not resolve the cargo workspace of `{root}`: {error}"))?;
+
+    if !status.is_success() {
+        return Err(format!(
+            "The `root` option resolved to `{root}`, whose cargo workspace could not be resolved: \
+             {}",
+            stderr.trim()
+        ));
+    }
+
+    let located = Utf8Path::new(stdout.trim());
+    let workspace = located.parent().unwrap_or(located);
+    // `root` arrives canonicalized, so the comparison only holds if this side is
+    // too. A path cargo reports for a directory that has since gone is left as
+    // printed, and fails the comparison below.
+    let workspace = workspace
+        .canonicalize_utf8()
+        .unwrap_or_else(|_| workspace.to_owned());
+
+    if workspace == root {
+        return Ok(());
+    }
+
+    Err(format!(
+        "The `root` option resolved to `{root}`, which is a member of the cargo workspace at \
+         `{workspace}`. Cargo would operate on that workspace instead. Name the workspace root, \
+         and scope the command with the `package` argument."
+    ))
 }
 
 /// Capabilities a cargo subcommand needs on the directory it runs in.
--- a/.config/jp/tools/src/cargo.rs
+++ b/.config/jp/tools/src/cargo.rs
@@ -1,11 +1,13 @@
-use camino::{Utf8Path, Utf8PathBuf};
-use jp_tool::{AccessPolicy, Capability, Outcome};
+use jp_tool::Capability;
 use serde_json::Value;
 
 use crate::{
     Context, Tool,
-    fs::utils::{authorize, resolve_workspace_path},
-    util::{ToolResult, error, unknown_tool},
+    util::{
+        ToolResult, error,
+        root::{CARGO_MANIFEST, configured_root, note_root, resolve_root},
+        unknown_tool,
+    },
 };
 
 mod check;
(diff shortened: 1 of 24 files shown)


.config/jp/tools/src/util/root.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// Where the program goes instead when the entry is absent.
+    ///
+    /// Appended to the refusal, so the reader learns what the check prevented
+    /// rather than only that it fired.
+    upward_search: &'static str,
+}
+
+/// Marks a directory cargo can be run in.
+pub const CARGO_MANIFEST: Marker = Marker {
+    entry: "Cargo.toml",
```
JeanMertz: blocking: A plausible configuration is `root = "crates/jp_tool"`, or any other workspace member. This marker accepts it, but Cargo still discovers the enclosing workspace from a member manifest. `cargo_check` and `cargo_test` then run with `--workspace`, `cargo_format` runs with `--all`, and `cargo_update` can rewrite the enclosing `Cargo.lock`, even though authorization covered only `crates/jp_tool`. The update tool also reads `root/Cargo.lock`, so that parent lockfile mutation can be reported as no change. Please validate that Cargo's actual workspace root, for example from `cargo locate-project --workspace --manifest-path ...`, equals the resolved root before dispatching these commands, or reject member directories explicitly.

.config/jp/tools/src/util/root.rs
```diff
@@ -0,0 +1,164 @@
...
+    upward_search: "Cargo would search parent directories and operate on the enclosing workspace \
+                    instead.",
+};
+
+/// Marks the top level of a git repository.
+///
+/// Only existence is checked: `.git` is a directory in an ordinary checkout and
+/// a file in a worktree or a submodule.
+pub const GIT_DIR: Marker = Marker {
+    entry: ".git",
```
JeanMertz: blocking: A normal submodule or linked worktree has `.git` as a file containing `gitdir: ...`. Checking only that this entry exists accepts such a root without resolving or authorizing the Git directory it names. For example, `root = "vendor/submodule"` with access granted only to that path lets `git_commit` write the index, refs, and objects under the parent repository's `.git/modules/...`; a linked worktree can point outside the workspace entirely. Parse and canonicalize the gitfile target, also accounting for a symlinked `.git` directory, then authorize the actual Git directory or reject it unless an external mount grants access.

.config/jp/tools/src/git.rs
```diff
@@ -118,7 +145,40 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
...
+
+/// Capabilities a git subcommand needs on the repository it runs in.
+///
+/// These gate whether git is spawned at all; they cannot bound what it does
+/// once running, because the subprocess is not sandboxed.
+///
+/// Reading commands are declared read-only even though git refreshes
+/// `.git/index` as it goes.
+/// Declaring that write would make every grant that allows reading history also
+/// allow rewriting it, which is the larger of the two inaccuracies.
```
JeanMertz: blocking: A plausible read-only policy for redirected `git_status`, with `read = true` and `update = false`, passes this check, but a normal `git status` may lock and rewrite `.git/index` to refresh cached stat data. Since the subprocess is not sandboxed, that persistent mutation happens silently despite the denied capability. Set `GIT_OPTIONAL_LOCKS=0` for read-only invocations, at least for `status`, or require the write capabilities the command can use.

.config/jp/tools/src/cargo.rs
```diff
@@ -39,40 +41,64 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
...
     };
 
+    // Cargo profile to build under, set via `options.profile`. A profile of its
+    // own gives these tools their own directory under `target/`, which is the
+    // only way to keep their artifacts, fingerprints and build lock apart from
+    // a developer's concurrent `cargo run` in the same workspace.
+    //
+    // Refused when malformed rather than ignored: a silently dropped value puts
+    // the build back in the shared profile without saying so.
+    let profile = match t.options.get("profile") {
```
JeanMertz: blocking: This parses and accepts `options.profile = "agent"` before dispatch, but the `cargo_install_tools` branch does not pass it onward. That tool runs `just install-tools`, whose recipe uses `cargo install --path ... --debug`; path installs use the workspace target directory. If a developer is concurrently building the dev profile, this invocation still uses the shared `target/debug` lock and fingerprints, which is exactly what the option promises to avoid, and success gives no indication that the profile was ignored. Thread the profile into the install recipe, or reject it explicitly for this subcommand.

.config/jp/tools/src/util/root.rs
JeanMertz: Agreed, and this was the strongest of the four. Verified every specific: `check.rs` and `test.rs` both `map_or("--workspace", …)`, `format.rs` uses `--all`. A member root clears the marker check and then cargo resolves the workspace from that manifest, so the command lands on the enclosing project — which is precisely the outcome the marker's own refusal text claims to prevent.

Fixed by asking cargo where it thinks the workspace is, rather than parsing the manifest, so the answer cannot drift from what the build commands themselves resolve:

```
cargo locate-project --workspace --message-format=plain --manifest-path <root>/Cargo.toml
```

`ensure_workspace_root` runs this whenever the root was redirected and refuses when the located manifest's parent is not the resolved root, naming both directories. It lives in `cargo.rs` rather than `root.rs`: it is cargo-specific and `root.rs` is shared with git. Covered by three tests — member refused, workspace root accepted, and a manifest cargo cannot resolve refused rather than falling through to a build that would walk up on its own.

One correction on the `cargo_update` detail: it does not report the parent mutation as "no change". `update.rs` joins `Cargo.lock` onto the configured root, a member has no lockfile there, so `lockfile_change` returns `Unknown` and the section is omitted entirely. The underlying point — an unauthorized write to the enclosing lockfile, unreported — stands, and is what the fix closes.

.config/jp/tools/src/cargo.rs
JeanMertz: Correct on the mechanism. `just install-tools` is `cargo install --locked --path .config/jp/tools --debug`, and the cargo book is explicit that with `--path` the build uses the `target` directory of the local crate's workspace; `--debug` selects `dev`. Since `.config/jp/tools` is a member of the root workspace, that is `<repo>/target/debug` — the same directory, build lock and fingerprints the option exists to stay clear of.

What settled it was the PR's own rule two lines above the anchor: malformed values are refused rather than silently dropped, because a dropped value puts the build back in the shared profile without saying so. Dropping a well-formed one does the same thing with better manners.

Took the second of your two options: `run` now refuses `profile` for any subcommand that cannot honor it (`format` and `update` never compile, `install_tools` builds through a recipe that pins its own profile). Threading it instead would mean `_install-tools` takes the profile and stops hard-coding `--debug`, which reaches into the justfile and past what this PR is doing — worth a follow-up if the isolation turns out to matter for `install_tools` too.

Test drives `run` with `options.profile` set on `cargo_install_tools` and asserts the refusal; it returns before `just` is spawned, so no fixture is needed.

.config/jp/tools/src/git.rs
JeanMertz: Took the first remedy, not the second.

`GIT_OPTIONAL_LOCKS=0` is now seeded in `env_from_options`, so every git invocation gets it and the `env` tool option is appended afterwards — the runner applies pairs in order, so a caller can still override it. Safe for the write commands: git's docs are specific that the variable suppresses *optional* sub-operations, and names `git status` refreshing the index as the example; `commit` and `apply` need their lock to do the requested work at all, so they still take one. Three unit tests pin the default, the append, and the override ordering.

Not taking the alternative. Declaring reads as writes is argued against three lines above the anchor: it would make every grant that allows reading history also allow rewriting it, which is the larger of the two inaccuracies. That reasoning still holds.

On severity, I'd downgrade this from blocking. The input is reachable but no tool config sets `root` today, and the effect is git refreshing its own stat cache: idempotent, no change to tracked content, history or the working tree. Worth the two lines because it makes the declaration honest, not because it gates the merge.

.config/jp/tools/src/util/root.rs
JeanMertz: Declining this one, and the symlink observation is fair — `.exists()` follows symlinks, so a symlinked `.git` directory passes the same way a gitfile does.

Three reasons the full remedy costs more than the risk it retires.

Accepting gitfiles is deliberate and tested (`a_worktree_whose_git_is_a_file_is_accepted`), because git works normally in a worktree or submodule. For the submodule case specifically there is no escape: writes landing under `<parent>/.git/modules/<name>/` is where that repository's git directory *is*, not the tool reaching somewhere it was not pointed.

The capability check gates spawning, not confinement — stated outright in the doc comment on `required_capabilities`. Authorizing the resolved git directory tightens bookkeeping around a subprocess that is not sandboxed and carries the process's full filesystem access regardless.

And the cost is not small: gitfile parsing, relative-gitdir resolution, symlink handling and worktree-vs-submodule branching, all pushed into a `Marker` whose entire job today is "does this entry exist", and all of it git-specific in a module shared with cargo. Against that, this repo has no `.gitmodules`, no `root` set in any tool config, and no `external` mount — and this is the repo's own maintenance tooling, not shipped surface.

What that leaves open: a linked worktree whose `gitdir:` points outside the workspace lets git write outside it, without the confinement `root` otherwise enforces. If you want that closed, the cheap version is one check inside `resolve_root` — when the marker entry is a file or a symlink, refuse unless its target resolves inside the workspace — rather than parsing and authorizing the git directory. Say the word and I'll add that instead.

.config/jp/tools/src/git.rs
```diff
@@ -118,24 +145,71 @@ pub async fn run(ctx: Context, t: Tool) -> ToolResult {
...
+        }
+        // Writes objects and refs, and runs hooks — arbitrary code carrying the
+        // process's own filesystem access.
+        "commit" => &[
+            Capability::Read,
+            Capability::Create,
+            Capability::Update,
+            Capability::Execute,
+        ],
+        _ => &[Capability::Read],
```
JeanMertz: blocking: A plausible redirected root is a Git LFS checkout granted read access, plus create/update access when staging. The diff-driven paths behind `git_diff`, `git_diff_file`, `git_list_patches`, `git_stage_patch`, and `git_stage_patch_lines` run `git diff-files -p`. Git converts modified worktree content through configured clean/filter-process drivers; for an LFS file that starts `git-lfs` and can populate `.git/lfs/objects`. `GIT_OPTIONAL_LOCKS=0` does not suppress filters. These capability sets therefore spawn code, and may persist objects, without `Execute` (or, for read commands, `Create`) being granted. Require those capabilities for commands that can invoke filters, or make the inspection path filter-free before treating it as read-only.

PR #1045, opened 2026-08-31
fix(workspace): Refresh conversation locks

JeanMertz:
JP reloads cached conversation data after acquiring its lock, so a command that waited for another writer preserves turns and metadata written before the lock became available.

Missing backing-store data retains an unpersisted in-memory conversation, while other load errors stop the operation before stale data is written.


crates/jp_workspace/src/lib.rs
```diff
@@ -861,6 +865,46 @@ impl Workspace {
...
+            }
+        }
+
+        if let Some(cell) = self.state.events.get(id) {
+            match cell.get() {
+                None => maybe_init_events(&*self.loader, (id, cell)),
+                Some(arc) => match self.loader.load_conversation_stream(id) {
+                    Ok(stream) => *arc.write() = stream,
+                    Err(error) if error.kind().is_missing() => {}
+                    Err(error) => return Err(error.into()),
```
JeanMertz: blocking: A plausible race is `jp conversation compact ID` caching the stream while it waits, followed by another process completing `jp conversation rm ID`. Once the removal releases the flock, both loads report missing; these arms retain the cached metadata and events, so compact acquires the lock and its next `flush()` recreates the deleted conversation. Archiving can similarly leave cached active data available after the directory moved. Missing data is only safe to preserve for a conversation known to be newly created in memory. Please track that state, or return the missing error for indexed conversations, and add a removal or archive regression test.

crates/jp_workspace/src/lib.rs
```diff
@@ -706,12 +715,7 @@ impl Workspace {
             return Ok(LockResult::AlreadyLocked(handle));
         };
 
-        if let Some(cell) = self.state.conversations.get(&id) {
-            maybe_init_conversation(&*self.loader, (&id, cell));
-        }
-        if let Some(cell) = self.state.events.get(&id) {
-            maybe_init_events(&*self.loader, (&id, cell));
-        }
+        self.sync_conversation(&id)?;
```
JeanMertz: blocking: The payload is refreshed here, but `state.presence` is not. For example, process A can index a projected conversation and wait in `jp conversation compact`, while process B runs `jp conversation edit --local` and removes the workspace projection. A then acquires the flock with fresh events but `lock_projection()` still returns `Projected`, so its flush silently recreates the projection and reverses B's change. The reverse race can remove a projection B just enabled. Refresh the conversation's storage presence under the lock as well, and cover a locality change between the initial read and acquisition.

crates/jp_workspace/src/lib.rs
```diff
@@ -861,6 +865,46 @@ impl Workspace {
...
+    /// in-memory conversation that was never persisted is the whole truth about
+    /// itself.
+    /// Any other load failure is returned rather than leaving a stale copy in
+    /// place for the next write to persist.
+    fn sync_conversation(&self, id: &ConversationId) -> Result<()> {
+        if let Some(cell) = self.state.conversations.get(id) {
+            match cell.get() {
+                None => maybe_init_conversation(&*self.loader, (id, cell)),
+                Some(arc) => match self.loader.load_conversation_metadata(id) {
+                    Ok(metadata) => *arc.write() = metadata,
```
JeanMertz: theoretical: A caller can bind `let cached = workspace.events(&handle)?;` and then pass the handle to `lock_conversation`. This in-place `write()` refresh waits for every read guard to drop; when the same thread still owns `cached`, it waits forever. I did not see a current CLI path retaining a guard through acquisition, so I cannot name a user command that triggers it. At minimum, document that guards returned by `events` and `metadata` must be dropped first. Prefer `try_write` with a typed error, or a cache shape that can swap the `Arc`, so misuse cannot hang the process.

crates/jp_workspace/src/lib.rs
JeanMertz: Confirmed and fixed.

The chain is exactly as described: `state.presence` is written only by `load_conversation_index` (`lib.rs:366-369`), `create_conversation_with_projection`, and `unarchive_conversation`, so `lock_projection` returned whatever the pre-wait index scan saw. And the projection is not advisory — `Storage::persist_conversation` (`jp_storage/src/lib.rs:213-219`) writes the workspace copy under `Projected` and calls `remove_conversation_dirs` under `LocalOnly`, so a stale value reverses B's change in either direction.

I did not update `state.presence` itself: `lock_conversation` takes `&self` and `presence` is a plain `HashMap`, so that would force `&mut self` and cascade through `Ctx`. Instead the projection is resolved from the store at acquisition, reusing the index-scan idiom already in `unarchive_conversation`:

```rust
/// The write projection for a lock being acquired on `id`.
///
/// Read from the store rather than the workspace index: another process can
/// change a conversation's locality (`jp conversation edit --local`) while
/// this one waits for the flock, and the projection decides whether the
/// next write creates or drops the workspace copy.
/// A conversation the store does not have falls back to its recorded
/// presence, which for an unpersisted conversation is the intent it was
/// created with.
fn acquired_lock_projection(&self, id: &ConversationId) -> Projection {
    self.loader
        .load_conversation_index(ConversationFilter::default())
        .into_iter()
        .find(|entry| entry.id == *id)
        .map_or_else(
            || self.lock_projection(id),
            |entry| Projection::from(entry.presence),
        )
}
```

Used in `lock_conversation` only; `lock_new_conversation` keeps `lock_projection`, since a fresh conversation's projection is the caller's stated intent rather than a stored fact.

Cost is one index scan per acquisition, on top of the one every command already runs at startup. If it ever shows up in a profile the answer is a per-id `LoadBackend::load_conversation_presence`, not a cache.

`lock_reads_the_projection_from_the_store_after_a_wait` covers the locality change between the initial read and acquisition, on a two-root backend (the single-root fixture ignores projection entirely, per `persist_conversation`'s `None` arm). It asserts both `lock.projection()` and that a subsequent flush leaves the workspace projection gone. Verified red before the fix: `left: Projected, right: LocalOnly`.

crates/jp_workspace/src/lib.rs
JeanMertz: Taking the documentation, declining the two stronger options.

The hazard is real and it is new here — pre-PR `maybe_init_*` only called `cell.set()` and never took a write guard. `events()` returns a guard borrowed from `&self` and `lock_conversation` also takes `&self`, so holding one across the other compiles, and `parking_lot::RwLock` is not reentrant.

It is not a merge gate, and your own note is why: no input reaches it. Triggering it takes *new code*, and that code hangs deterministically on its first run, in dev, at the line that wrote it. Contained and visible.

Documented on `lock_conversation`:

```rust
/// A value read through [`events`] or [`metadata`] before the lock was held
/// is superseded by that read, and the guard it returned must be dropped
/// first: the refresh takes the write side of the same lock.
```

On the alternatives:

`try_write` + a typed error converts a dev-time hang into a permanent error variant that every caller has to handle and no correct caller can produce. Worse trade than a doc line.

The `Arc`-swap I'd actively reject. In-place mutation is *why* the refresh works: `ConversationLock` and every `ConversationMut` hold clones of that `Arc`, and the `sync_conversation` doc leans on it ("scopes already sharing them observe the refreshed data"). Swapping the cell's `Arc` would leave any live scope writing the pre-refresh value — the class of bug this PR closes, reintroduced one level down.

crates/jp_workspace/src/lib.rs
JeanMertz: The analysis holds, including the invariant — missing data is only safe to preserve for a conversation known to be newly created in memory. `jp c rm` holds the flock across `remove_conversation_with_lock` (`rm.rs:102-109`), so a waiting `compact` resumes against a deleted conversation and flushes it back.

Deferring the behavior change to T-0b7d3w0, for two reasons.

It is not a regression. Pre-PR, `maybe_init_conversation` returned early on a populated cell, so the cached copy survived the deletion and the flush recreated it — same outcome, reached by omission instead of by an explicit arm. Merging this strictly reduces stale-data damage; blocking it on an adjacent bug of equal severity keeps the far more common event-loss case live in `main`.

And the fix needs a discriminator we do not have. Neither `state.presence` nor `state.conversations` membership separates "loaded from the store" from "seeded by `create_conversation`" — creation writes both. It needs provenance tracked across create → first successful persist, which is new state on the highest-risk path in the crate. The ticket sketches a `never_persisted` set and notes that `sync_conversation` takes `&self`, so it needs interior mutability or an owner on the persist path.

One correction to the archive half: the two loads disagree about the archive partition, so it is a mixed state rather than a clean resurrection. `FsStorageBackend::load_conversation_metadata` *does* fall back to the archive (`fs.rs:240`), so metadata refreshes and brings `archived_at` with it, while `load_conversation_stream` has no fallback (`fs.rs:250-255`) and keeps the cached stream. The waiting writer then persists into the live partition, producing a live conversation carrying `archived_at`. Recorded on the ticket.

What I did change here is the comment, which overclaimed:

```rust
/// A conversation the store does not have keeps its cached copy, covering
/// an in-memory conversation that was never persisted.
/// This does not distinguish that case from a conversation another process
/// removed or archived while this one waited, which the next write
/// recreates.
```

JeanMertz: blocking: The triggering input is still two ordinary commands: `jp c compact ID` waiting while `jp c rm ID` completes. The next compact flush silently recreates the deleted conversation. That the behavior predates this PR changes where it originated, but not the merge gate under the named-input rule: the damage persists on disk, and the user is told neither that removal was reversed nor that stale content was written. The provenance work is larger than the refresh, but `sync_conversation` is the point that decides whether cached state is safe to write. Please distinguish never-persisted state here, or make a missing indexed conversation abort acquisition.

PR #1064, opened 2026-09-03
fix(config)!: Stop duplicating prompt blocks across layers

JeanMertz:
A config file with a `assistant.system_prompt` field reached from both the workspace config and a `--cfg <file>` argument contributed its prompt block twice. Each layer resolves its own `extends` graph, so a shared file is loaded once per graph, and the value-level dedup that should have caught the repeat only recognized a block when the text around it happened to be joined with the same separator the block itself declared. A prompt assembled from sources that chose different separators failed that check. Matching now anchors on line breaks, which every `paragraph`- and `line`-joined block carries regardless of what its neighbours used.

`assistant.system_prompt.dedup` takes a mode rather than a boolean, so how a repeat is recognized is stated separately from how values are joined: `block` (the default) skips a value present as a whole line-bounded block, `contains` skips one present anywhere including inside a line, `exact` skips only a value equal to the whole string, and `off` merges unconditionally. `true`, `false` and `"inherit"` still parse, as `block`, `off` and "no opinion". A skipped value is reported at debug level, since dropping it is otherwise indistinguishable from never having supplied it.

Appending to an empty value no longer emits a leading separator.

BREAKING CHANGE: `separator` defaults to `paragraph` instead of `none`

An `append` or `prepend` that does not state a `separator` now puts a blank line between the two values instead of concatenating them. This applies to `assistant.system_prompt` and to any other string merged with a strategy. A config that builds one value out of fragments needs an explicit `separator = "none"` to keep joining them directly. The old default also made dedup nearly inert: with no separator there is no boundary to recognize a repeated block by.

### DIFF
--- a/crates/jp_config/src/internal/merge/string.rs
+++ b/crates/jp_config/src/internal/merge/string.rs
@@ -3,8 +3,11 @@
 #![expect(clippy::unnecessary_wraps, clippy::trivially_copy_pass_by_ref)]
 
 use schematic::MergeResult;
+use tracing::debug;
 
-use crate::types::string::{MergedStringStrategy, PartialMergeableString, PartialMergedString};
+use crate::types::string::{
+    MergedStringStrategy, PartialMergeableString, PartialMergedString, StringDedup,
+};
 
 /// Merge two ` PartialMergeableString  ` values.
 pub fn string_with_strategy(
@@ -39,21 +42,35 @@ pub fn string_with_strategy(
         }
     };
 
-    // Skip an append or prepend whose value is already present, unless a config
-    // explicitly opts out. Applying the same config source twice — through an
-    // `extends` diamond, or by re-supplying `--cfg` on a conversation that
-    // already merged it — must not duplicate its contribution.
-    let dedup_active = dedup.unwrap_or(true);
+    // Skip an append or prepend whose value is already present. Applying the
+    // same config source twice — through an `extends` diamond, or by
+    // re-supplying `--cfg` on a conversation that already merged it — must not
+    // duplicate its contribution.
+    let dedup_mode = dedup.unwrap_or_default();
 
-    // An unstated strategy means `append`, matching `MergedStringStrategy`'s
-    // default. Only the computation resolves it; the stored `strategy` keeps
-    // the unstated form so later merges can still express their own.
+    // An unstated strategy means `append` and an unstated separator means
+    // `paragraph`, matching the two types' defaults. Only the computation
+    // resolves them; the stored fields keep the unstated form so later merges
+    // can still express their own.
     let resolved_strategy = strategy.unwrap_or_default();
-
-    let sep = separator.as_ref().map_or("", |sep| sep.as_str());
+    let sep = separator.unwrap_or_default().as_str();
     let value = match (prev_value, next_value) {
         (_, n) if resolved_strategy == MergedStringStrategy::Replace => n,
-        (Some(p), Some(n)) if dedup_active && contains_block(&p, &n, sep) => Some(p),
+        // Nothing on one side means nothing to separate from.
+        (Some(p), Some(n)) if p.is_empty() => Some(n),
+        (Some(p), Some(n)) if n.is_empty() => Some(p),
+        (Some(p), Some(n)) if is_present(&p, &n, dedup_mode) => {
+            // The only path that drops a value the author wrote. Nothing
+            // downstream can tell it apart from a value never supplied, so a
+            // reader hunting a missing prompt section has this line and
+            // nothing else.
+            debug!(
+                mode = %dedup_mode,
+                value = %excerpt(&n),
+                "Skipping a value already present in the merged string."
+            );
+            Some(p)
+        }
         (Some(p), Some(n)) if resolved_strategy == MergedStringStrategy::Append => {
             Some(format!("{p}{sep}{n}"))
         }
@@ -82,7 +99,7 @@ pub fn string_with_strategy(
 }
 
 /// The explicit `dedup` opinion carried by a value, if any.
-const fn dedup_flag(v: &PartialMergeableString) -> Option<bool> {
+const fn dedup_flag(v: &PartialMergeableString) -> Option<StringDedup> {
     match v {
         PartialMergeableString::String(_) => None,
         PartialMergeableString::Merged(m) => m.dedup,
@@ -93,7 +110,10 @@ const fn dedup_flag(v: &PartialMergeableString) -> Option<bool> {
 /// a `replace` strategy so the flag survives the next merge.
 ///
 /// A `None` opinion is left implicit and the value's shape is unchanged.
-fn with_dedup_flag(v: PartialMergeableString, dedup: Option<bool>) -> PartialMergeableString {
+fn with_dedup_flag(
+    v: PartialMergeableString,
+    dedup: Option<StringDedup>,
+) -> PartialMergeableString {
     if dedup.is_none() {
         return v;
     }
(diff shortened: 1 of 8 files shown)


crates/jp_config/src/internal/merge/string.rs
```diff
@@ -39,21 +42,35 @@ pub fn string_with_strategy(
...
-        (Some(p), Some(n)) if dedup_active && contains_block(&p, &n, sep) => Some(p),
+        // Nothing on one side means nothing to separate from.
+        (Some(p), Some(n)) if p.is_empty() => Some(n),
+        (Some(p), Some(n)) if n.is_empty() => Some(p),
+        (Some(p), Some(n)) if is_present(&p, &n, dedup_mode) => {
+            // The only path that drops a value the author wrote. Nothing
+            // downstream can tell it apart from a value never supplied, so a
+            // reader hunting a missing prompt section has this line and
+            // nothing else.
+            debug!(
```
JeanMertz: nit: A repeated `--cfg` prompt reaches this branch during ordinary loading, but emitting here makes `jp_config::internal::merge` depend on the process-global tracing subscriber. These merge primitives are the pure core; callers should be able to evaluate them without observable I/O. Please either keep this silent or return or collect an explicit merge diagnostic that the loading orchestration logs at its boundary.

crates/jp_config/src/types/string.rs
```diff
@@ -192,29 +199,36 @@ pub struct MergedString {
...
     /// `"inherit"` (or omitting the field) means "no opinion" — inherit from
-    /// the previous merge, falling back to `true`.
+    /// the previous merge, falling back to `block`.
     #[setting(
         skip_serializing_if = "Option::is_none",
-        deserialize_with = "deserialize_dedup"
+        deserialize_with = "deserialize_string_dedup"
     )]
-    pub dedup: Option<bool>,
+    pub dedup: Option<StringDedup>,
```
JeanMertz: nit: An existing config with `dedup = false` still deserializes through the custom visitor, but the schema inferred from `Option<StringDedup>` advertises only the named enum values (plus null). Schema validators and editor integrations will therefore reject a supported compatibility form. Please describe the boolean and `"inherit"` input shapes in the field schema as well, and add a schema-shape test like the one for `EnableConfig`.

crates/jp_config/src/types/string.rs
```diff
@@ -192,29 +199,36 @@ pub struct MergedString {
...
-    /// Set to `false` to append the value unconditionally.
-    /// Accepts `true`, `false`, or `"inherit"`.
+    /// `true` and `false` are accepted as shorthand for `block` and `off`.
     ///
-    /// A value counts as present when it appears in the existing string as a
-    /// whole `separator`-delimited block.
-    /// Partial matches inside a block do not count.
-    /// With `separator = "none"` there are no block boundaries to match
-    /// against, so only an exact match of the whole string counts.
+    /// `block` and `exact` need a line break around the value to recognize it,
```
JeanMertz: nit: This is incorrect for `exact`: `is_present` only compares `accumulated == value`, so line breaks and the separator do not affect that mode. Since this doc becomes config documentation and schema text, please reserve the boundary explanation for `block` and describe `exact` as whole-string equality regardless of separator.

crates/jp_config/src/util_tests.rs
```diff
@@ -1005,6 +1005,101 @@ fn test_load_partial_at_path_dedups_load_paths_across_files() {
...
+        .assistant
+        .instructions
+        .iter()
+        .map(|i| i.title.as_deref())
+        .collect();
+    assert_eq!(instructions, [Some("Tool Usage")]);
+
+    // `web.toml`'s block is followed by `coding.toml`'s, joined with the `line`
+    // separator that file appends with rather than the `paragraph` one
+    // `web.toml` carries — so the block-boundary check misses it and the block
```
JeanMertz: nit: This says the check misses the block and appends it twice, which is the opposite of the assertion below. The regression proves that the line-boundary check recognizes `web.toml` despite the different separator and skips the second contribution. Please describe that current invariant so the test does not document the bug it prevents as present behavior.

crates/jp_config/src/types/string.rs
JeanMertz: Correct, fixed in c0476a4e.

`is_present` reads `StringDedup::Exact => accumulated == value`, so no boundary is involved and the separator is irrelevant to that mode. The sentence now describes `block` on its own:

```
/// `block` recognizes a value inside a longer string by the line breaks
/// around it, which a value merged with `separator = "space"` or `"none"`
/// does not have; such a value is only recognized when it equals the whole
/// string.
/// `contains` is the mode that recognizes a value merged inside a line.
```

I also fixed the same imprecision one field up: the `separator` doc said the line-break rule was "what `dedup` matches on", stated unconditionally where it holds for the default mode and not for `contains`. Both texts ship as config documentation, so leaving one half-true would have traded one wrong summary for another.

crates/jp_config/src/util_tests.rs
JeanMertz: Correct, fixed in c0476a4e.

The comment was written while this test was the red reproduction and never updated when the fix landed, so it described the bug as current behavior directly above an assertion proving otherwise. It now states the invariant:

```
// `web.toml`'s block is followed by `coding.toml`'s, joined with the `line`
// separator that file appends with rather than the `paragraph` one
// `web.toml` carries. Anchoring the match on the line break rather than on
// either separator is what recognizes the repeat and skips it.
```

crates/jp_config/src/types/string.rs
JeanMertz: The mechanics are right and I've filed it as T-0dkct9c rather than fixing it here. `deserialize_with` doesn't reach the schema (`Field::generate_schema_type` builds it from the field type), so the field advertises four mode names plus null while the deserializer also takes `true`, `false` and `"inherit"` — which my own doc comment advertises.

Three reasons it isn't in this PR.

No consumer exists. Nothing exports a JSON Schema for `AppConfig`: `SchemaBuilder::build_root::<AppConfig>()` appears only in RFD 061 and RFD 063, both unimplemented, `.config/workspace-schema.json` is Cargo's manifest schema, and `jp init` uses `ConfigEnum` for variant listing rather than the schema renderer. There is nothing today for an editor to validate against.

The gap is wider than this field. `MergedVec::dedup` is `Option<bool>` with `deserialize_dedup`, which accepts an `"inherit"` the schema likewise omits. This PR doesn't touch it, and fixing only the string side leaves the identical bug next door.

The suggested fix has no mechanism at field level. `schema_union_with` is a `FromDeriveInput` argument on `MacroArgs`, which is why `EnableConfig` can use it and a plain field cannot. Matching `test_enable_schema` means either adding field-level union support to `schematic_macros` or promoting `dedup` to its own `Config` container with a hand-written `Deserialize` — the latter also changes the persisted shape of a field that reaches conversation config deltas. The ticket records both options and prefers the first, since it fixes both fields at once.

crates/jp_config/src/internal/merge/string.rs
JeanMertz: Keeping the log. The purity instinct is fair, but the specific claim doesn't hold.

`tracing::debug!` with no subscriber installed is a no-op: the macro checks the global dispatcher and returns, performing no I/O and mutating no state. A caller can evaluate `string_with_strategy` without observable output, which is exactly what every non-`test_log` test in `string_tests.rs` does.

The boundary this appeals to also isn't the one the crate draws. `jp_config` already logs from `util.rs` (five levels), `fs.rs`, and `editor.rs` — the last inside a `*Config` module. The documented rule for the crate is no file I/O, no CLI parsing, no terminal rendering, no workspace discovery; a `debug!` is none of those, and the project's architecture notes rank observability a must with an explicit bias toward logging decisions rather than staying quiet.

The alternative isn't expressible at this signature either. Schematic's merge contract is `fn(prev, next, &Context) -> MergeResult<T>` where `MergeResult<T> = Result<Option<T>, MergeError>` and `Context = ()` for every JP config type, so there is no channel for a non-fatal diagnostic without changing schematic itself — a much larger change than the four lines here.

And "keep this silent" gives back the thing the line exists for. The dedup arm is the only path that drops text the author wrote, indistinguishable downstream from a value never supplied. That invisibility is what made the originating bug report undiagnosable: a user could see prompt content going missing and had nothing to grep for.

PR #1066, opened 2026-09-03
feat(cli, config): Accept comma-separated tool names

JeanMertz:
`--tool` and `--no-tool` take several tool names in one flag, separated by commas, so a query that opens up three tools no longer needs three flags:

```sh
jp query -t cargo_check,cargo_test,fs_read_file "Does this compile?"
```

Space around a name is ignored, repeating the flag still works and composes with the comma form, and each name keeps its place in the left-to-right order that `--no-tool --tool=write` relies on. A bare `--tool` or `--no-tool` is unchanged: it still covers every configured tool.

A missing value is refused rather than guessed at, wherever a comma separates values. `-t read,,write` reports the stray comma instead of reading it as the bare flag and enabling everything, and a `--cfg` list does the same instead of quietly dropping the gap. Assigning nothing at all is untouched: `conversation.labels.crate=` still names no values and remains the way to clear a list, and a value that genuinely contains a comma is written with the JSON form, `crate:="feat,exp"`.

The tool-use documentation is corrected alongside. It described `--tool=false`, `--tool=true`, and `--tool=<name>`, which are spellings of `--tool-use` and `--no-tool-use`; `--tool=false` asks for a tool named "false" and errors.

BREAKING CHANGE: A tool name cannot contain a comma

Configuring a tool whose name contains a comma is rejected when the configuration resolves, naming the offending key:

    conversation.tools.read,write: a tool name cannot contain a comma,
    because `--tool` and `--no-tool` read it as a separator between names

Rename the tool to clear the error.

### DIFF
--- a/crates/jp_config/src/assignment.rs
+++ b/crates/jp_config/src/assignment.rs
@@ -1076,6 +1076,10 @@ fn flatten_recursive(
 }
 
 /// Try to parse a comma-separated list of values, using the given parser.
+///
+/// Space around a value is ignored.
+/// An empty value between commas is an error; to name a value that contains a
+/// comma, use the JSON form of the assignment.
 fn try_parse_vec<'a, T, E>(
     key: &KvKey,
     s: &'a str,
@@ -1084,13 +1088,26 @@ fn try_parse_vec<'a, T, E>(
 where
     E: Into<BoxedError>,
 {
+    // Nothing at all is how a list is cleared, and stays distinct from a value
+    // that is missing between two commas.
+    if s.trim().is_empty() {
+        return Ok(vec![]);
+    }
+
     s.split(',')
         .map(str::trim)
-        .filter(|s| !s.is_empty())
         .enumerate()
-        .map(|(i, s)| {
-            parser(i, s)
-                .or_else(|err| assignment_error(key, Value::String(s.to_owned()), err.into()))
+        .map(|(i, value)| {
+            if value.is_empty() {
+                return assignment_error(
+                    key,
+                    Value::String(s.to_owned()),
+                    format!("empty value in '{s}'").into(),
+                );
+            }
+
+            parser(i, value)
+                .or_else(|err| assignment_error(key, Value::String(value.to_owned()), err.into()))
         })
         .collect::<std::result::Result<Vec<_>, _>>()
 }
--- a/crates/jp_config/src/assignment.rs
+++ b/crates/jp_config/src/assignment.rs
@@ -835,7 +835,9 @@ impl KvAssignment {
     ///
     /// Three shapes reach the list: a JSON array is taken element-wise, a JSON
     /// string is one element, and a bare string is split on commas, with each
-    /// element trimmed and empty ones dropped.
+    /// element trimmed.
+    /// An empty element is an error; a value that is empty in its entirety
+    /// names no elements at all.
     /// The comma shorthand is confined to the bare form, so `key:="a,b"` names
     /// one element and `key=a,b` names two.
     pub(crate) fn try_vec<T>(
--- a/crates/jp_config/src/conversation/tool.rs
+++ b/crates/jp_config/src/conversation/tool.rs
@@ -41,7 +41,8 @@ pub struct ToolsConfig {
     /// Tool config
     ///
     /// This section configures individual tools.
-    /// The key is the tool ID.
+    /// The key is the tool ID, and cannot contain a comma: a comma separates
+    /// one tool ID from the next wherever several are named at once.
     #[setting(nested, flatten, merge = merge_nested_indexmap)]
     tools: IndexMap<String, ToolConfig>,
 }
--- a/crates/jp_config/src/conversation/tool.rs
+++ b/crates/jp_config/src/conversation/tool.rs
@@ -188,8 +188,27 @@ impl ToolsConfig {
 impl Validator for ToolsConfig {
     /// Validate cross-field invariants on the tools configuration.
     fn validate(&self) -> Result<(), ConfigError> {
+        reject_comma_in_tool_names(self)?;
         reject_access_on_non_local_tools(self)
     }
+}
+
+/// Reject a tool whose name contains a comma.
+///
+/// The `--tool` and `--no-tool` flags read a comma as the separator between
+/// tool names, so such a tool could never be enabled or disabled from the
+/// command line.
+fn reject_comma_in_tool_names(tools: &ToolsConfig) -> Result<(), ConfigError> {
+    for name in tools.tools.keys() {
+        if name.contains(',') {
+            return Err(HandlerError::new(format!(
+                "conversation.tools.{name}: a tool name cannot contain a comma, because `--tool` \
+                 and `--no-tool` read it as a separator between names"
+            ))
+            .into());
+        }
+    }
+    Ok(())
 }
 
 /// Reject `access` on tools whose finalized source is `builtin` or `mcp`.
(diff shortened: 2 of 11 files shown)


docs/features.md
```diff
@@ -148,15 +148,38 @@ explicit instructions into the prompt.
...
+gives one to a tool.
+
+```sh
+jp query -t cargo_check,cargo_test "Does this compile?"
+jp query --no-tool=fs_modify_file "Review this function."
+```
+
+Both flags apply to the whole conversation: the change carries into every later
+query until you change it back.
+Without a value they cover every configured tool, and they are evaluated left to
```
JeanMertz: nit: A conversation with `enable = "explicit"` is a counterexample to this claim: bare `--tool` skips that tool because its `if_named` policy does not accept a bulk directive. Likewise, a locked-on tool survives bare `--no-tool`. Could this describe these as bulk directives whose effect remains subject to `allow_toggle`, and align the CLI long help's "all configured tools" wording with that behavior?

docs/features.md
```diff
@@ -148,15 +148,38 @@ explicit instructions into the prompt.
...
+jp query -T -t fs_read_file "What does this module do?"
+```
+
 ### Disable Tool Use
 
-You can disable the use of tools by using the `--tool=false` flag.
-Alternatively, use the `--no-tool` flag, or the short-hand `-T` flag.
+You can disable the use of tools for a single query with the `--no-tool-use`
+flag, or the short-hand `-U` flag.
+The next query has its tools back; use `--no-tool` to turn them off for good.
```
JeanMertz: nit: For a user deciding between these flags, "for good" implies a permanent or global change, but `--no-tool` only persists on this conversation and can be reversed by a later `--tool`. "For subsequent queries in this conversation" would state the actual scope.

docs/features.md
```diff
@@ -168,22 +191,26 @@ API, this flag will still work.
...
 
+Write the value with `=`, or put a bare `--tool-use` after your query.
+A bare flag before the query reads the first word that follows it as a tool
+name.
+
 ### Force Specific Tool Use
 
-You can force the use of a specific tool by using the `--tool=<name>` flag.
+You can force the use of a specific tool by using the `--tool-use=<name>` flag.
+The named tool runs even if it is disabled, and the choice applies to this query
```
JeanMertz: nit: A tool configured with `enable = { state = false, allow_toggle = "never" }` does not run here: `validate_tool_use` rejects it as locked off. Please include that exception, as the `--tool-use` CLI help already does, so users do not read this as bypassing the tool's policy.

crates/jp_config/src/assignment.rs
```diff
@@ -1076,6 +1076,10 @@ fn flatten_recursive(
 }
 
 /// Try to parse a comma-separated list of values, using the given parser.
+///
+/// Space around a value is ignored.
+/// An empty value between commas is an error; to name a value that contains a
```
JeanMertz: nit: A bare list such as `providers.mcp.example.arguments=a,,b` reaches this path and is rejected, but `KvAssignment::try_vec` still documents that bare strings have their "empty ones dropped" (line 838). Please update that caller-facing comment to say empty elements are rejected, so it does not contradict this behavior.

crates/jp_config/src/conversation/tool.rs
```diff
@@ -188,10 +188,29 @@ impl ToolsConfig {
...
         reject_access_on_non_local_tools(self)
     }
 }
 
+/// Reject a tool whose name contains a comma.
+///
+/// The `--tool` and `--no-tool` flags read a comma as the separator between
+/// tool names, so such a tool could never be enabled or disabled from the
+/// command line.
+fn reject_comma_in_tool_names(tools: &ToolsConfig) -> Result<(), ConfigError> {
```
JeanMertz: nit: A user defining `[conversation.tools.'read,write']` through the generated config or JSON Schema will not see this restriction in the field description, which currently only says that the key is the tool ID. Since `ToolsConfig` field docs feed both of those user-facing surfaces, please add the comma restriction to the `tools` field documentation as well.

crates/jp_config/src/assignment.rs
JeanMertz: Accepted, fixed. `try_vec`'s doc now reads:

```rust
/// Three shapes reach the list: a JSON array is taken element-wise, a JSON
/// string is one element, and a bare string is split on commas, with each
/// element trimmed.
/// An empty element is an error; a value that is empty in its entirety
/// names no elements at all.
```

The second sentence keeps the clear-a-list case explicit, since that is the one empty value `try_parse_vec` still accepts.

One correction to the example: the assignment key is `args`, not `arguments` (`providers/mcp.rs:126` matches on `kv.p("args")`), so the reaching path is `providers.mcp.example.args=a,,b`. The substance holds, that key goes through `try_some_vec_of_strings` into `try_vec`.

crates/jp_config/src/conversation/tool.rs
JeanMertz: The gap is real and the field doc now carries the restriction. The wording is not the one suggested, though: naming `--tool` and `--no-tool` there would put CLI vocabulary into `jp_config`, and this doc lands in the generated `config.toml` and the JSON Schema as the definition of what a tool ID may be. The constraint holds regardless of which surface names several tools at once, so it is stated in those terms:

```rust
/// Tool config
///
/// This section configures individual tools.
/// The key is the tool ID, and cannot contain a comma: a comma separates
/// one tool ID from the next wherever several are named at once.
```

This follows the precedent in `conversation/label.rs:58-66`, where the key grammar explains its excluded characters by role rather than by flag.

The error raised by `reject_comma_in_tool_names` still names both flags, deliberately: it is only ever read by someone who already hit the problem, and there the fastest fix is knowing which flags do the splitting.

docs/features.md
JeanMertz: Correct, and worse than described: the example was wrong for every workspace, not only ones configuring `explicit` tools. `describe_tools` is injected into every query with `PartialEnableConfig::LOCKED_ON` (`builtins.rs:20`, `query.rs:2167`), so `jp query -T -t fs_read_file` leaves `fs_read_file` **and** `describe_tools`, never "`fs_read_file` as the only tool".

Rewritten:

> Without a value they apply to every tool that permits it: a tool configured with `enable = "explicit"` responds only when it is named, and one configured with `enable = "always"` (such as the built-in `describe_tools`) can never be turned off.
> They are evaluated left to right, so this enables `fs_read_file` and nothing else that is optional:

I did not adopt "bulk directives" and `allow_toggle` here. `features.md` is the introductory tour, and those are the internal scope names; the config values a user actually writes carry the same fact.

Both long helps are aligned, since `--no-tool` claimed the same thing: "If no arguments are provided, every tool that allows it is enabled; a tool set to `explicit` or `always` is unaffected", and the matching clause on `--no-tool`.

docs/features.md
JeanMertz: Agreed, fixed:

> The next query has its tools back; use `--no-tool` to turn them off for the rest of the conversation.

The scope was already stated correctly four lines up, which made "for good" the one loose phrase in the section a reader lands on directly.

docs/features.md
JeanMertz: Confirmed on both paths and fixed. `validate_tool_use` probes `apply_directive_to_tool(..., ToggleScope::Named, true)` on a clone (`query.rs:2240`) and propagates the refusal, and `tool_definitions` drops a locked-off tool even when it is forced (`jp_llm/src/tool.rs:1206-1219`). The `-u` help already carried the exception (`query.rs:301-303`), so `features.md` was the only surface promising something the code refuses. Added:

> A tool locked off with `enable = { state = false, allow_toggle = "never" }` is refused instead of forced.

PR #1071, opened 2026-09-04
refactor(cli): Split promptability from stdout-is-a-terminal

JeanMertz:
[RFD 048] specifies two independent checks: `stdout.is_terminal()` for whether the consumer of JP's output can handle ANSI escapes, and controlling-terminal availability for whether a user is present to answer a prompt. The code answered both questions with a single `Term::is_tty`, so any change to one silently moved the other.

Add `Term::interactive` and move every prompting call site onto it: the editor re-open confirmation, label resolution, the title candidate picker, lock-timeout handling, plugin approval and binary resolution, and the tool coordinator's permission decisions. Rendering keeps `is_tty`: output format resolution, the retry line's cursor control, the waiting indicator, the tool renderer, and the OSC terminal title.

`run_turn_loop` fanned one flag into both kinds of consumer, so it takes both now. The derivation is unchanged — `interactive` starts as `stdout.is_terminal()` — leaving this commit behaviour-preserving. The two values diverge once `--non-interactive` can override the signal, and again when it moves to `/dev/tty` availability.

### DIFF
--- a/crates/jp_cli/src/cmd/lock.rs
+++ b/crates/jp_cli/src/cmd/lock.rs
@@ -49,7 +49,9 @@ pub(crate) enum LockOutcome {
 pub(crate) struct LockRequest<'a> {
     pub workspace: &'a Workspace,
     pub handle: ConversationHandle,
-    pub is_tty: bool,
+
+    /// Whether a user is present to answer the contention prompt.
+    pub interactive: bool,
     pub session: Option<&'a Session>,
     pub printer: &'a Printer,
     pub signals: &'a SignalRouter,
@@ -69,7 +71,7 @@ impl<'a> LockRequest<'a> {
         Self {
             workspace: &ctx.workspace,
             handle,
-            is_tty: ctx.term.interactive,
+            interactive: ctx.term.interactive,
             session: ctx.session.as_ref(),
             printer: &ctx.printer,
             signals: &ctx.signals,
@@ -108,7 +110,7 @@ pub(crate) async fn acquire_lock(mut r: LockRequest<'_>) -> Result<LockOutcome> 
     // First attempt — no timer yet.
     r.handle = match r.workspace.lock_conversation(r.handle, r.session)? {
         LockResult::Acquired(lock) => return Ok(LockOutcome::Acquired(lock)),
-        _ if !r.is_tty => return Err(Error::LockTimeout(id)),
+        _ if !r.interactive => return Err(Error::LockTimeout(id)),
         LockResult::AlreadyLocked(handle) => handle,
     };
 
--- a/crates/jp_cli/src/cmd/lock.rs
+++ b/crates/jp_cli/src/cmd/lock.rs
@@ -69,7 +69,7 @@ impl<'a> LockRequest<'a> {
         Self {
             workspace: &ctx.workspace,
             handle,
-            is_tty: ctx.term.is_tty,
+            is_tty: ctx.term.interactive,
             session: ctx.session.as_ref(),
             printer: &ctx.printer,
             signals: &ctx.signals,
--- a/crates/jp_cli/src/cmd/plugin/dispatch.rs
+++ b/crates/jp_cli/src/cmd/plugin/dispatch.rs
@@ -314,7 +314,7 @@ pub(crate) fn run_plugin(
         editor: crate::editor::build_editor_backend(&config.editor),
         // The configured mode, as every other inline reply in the CLI uses.
         edit_mode: reply_edit_mode(config.editor.inline.edit_mode),
-        is_tty: ctx.term.is_tty,
+        is_tty: ctx.term.interactive,
     };
 
     let PluginProcess {
@@ -1696,7 +1696,8 @@ pub(crate) async fn run_external(args: &[String], ctx: &mut Ctx) -> cmd::Output 
     }
 
     let config = ctx.config();
-    let Some(binary) = resolve_plugin_binary(subcommand, &config.plugins, ctx.term.is_tty).await?
+    let Some(binary) =
+        resolve_plugin_binary(subcommand, &config.plugins, ctx.term.interactive).await?
     else {
         return Err(unknown_subcommand_error(subcommand));
     };
--- a/crates/jp_cli/src/cmd/plugin/dispatch.rs
+++ b/crates/jp_cli/src/cmd/plugin/dispatch.rs
@@ -72,7 +72,7 @@ pub(crate) struct Composer<'a> {
 
     editor: Option<Arc<dyn EditorBackend>>,
     edit_mode: ReplyEditMode,
-    is_tty: bool,
+    interactive: bool,
 }
 
 impl Composer<'_> {
@@ -84,8 +84,8 @@ impl Composer<'_> {
             values: vec![],
         };
 
-        if !self.is_tty {
-            debug!("Plugin asked to compose without a terminal to ask on.");
+        if !self.interactive {
+            debug!("Plugin asked to compose without a user to ask.");
             return response;
         }
 
@@ -314,7 +314,7 @@ pub(crate) fn run_plugin(
         editor: crate::editor::build_editor_backend(&config.editor),
         // The configured mode, as every other inline reply in the CLI uses.
         edit_mode: reply_edit_mode(config.editor.inline.edit_mode),
-        is_tty: ctx.term.interactive,
+        interactive: ctx.term.interactive,
     };
 
     let PluginProcess {
@@ -1308,7 +1308,7 @@ pub(crate) fn find_any_plugin_binary(name: &str) -> Option<Utf8PathBuf> {
 pub(crate) async fn resolve_plugin_binary(
     name: &str,
     plugins_config: &PluginsConfig,
-    is_tty: bool,
+    interactive: bool,
 ) -> Result<Option<Utf8PathBuf>, cmd::Error> {
     let plugin_cfg = plugins_config.command.get(name);
 
@@ -1327,14 +1327,14 @@ pub(crate) async fn resolve_plugin_binary(
     }
 
     // 2. Check registry.
-    if let Some(path) = try_registry_install(name, plugins_config, is_tty).await? {
+    if let Some(path) = try_registry_install(name, plugins_config, interactive).await? {
         return Ok(Some(path));
     }
 
     // 3. Check $PATH with run policy.
     let segments: Vec<&str> = name.split('-').collect();
     if let Some(path) = find_plugin_binary(&segments) {
-        check_run_policy(name, &path, plugin_cfg, is_tty)?;
+        check_run_policy(name, &path, plugin_cfg, interactive)?;
         return Ok(Some(path));
     }
 
@@ -1368,7 +1368,7 @@ fn verify_checksum(
 async fn try_registry_install(
     name: &str,
     plugins_config: &PluginsConfig,
-    is_tty: bool,
+    interactive: bool,
 ) -> Result<Option<Utf8PathBuf>, cmd::Error> {
     let Some(reg) = registry::load_cached() else {
         return Ok(None);
@@ -1419,7 +1419,7 @@ async fn try_registry_install(
             )));
         }
         RunPolicy::Ask => {
-            if !is_tty {
+            if !interactive {
                 return Err(cmd::Error::from(format!(
                     "plugin `{id}` requires approval. Run `jp plugin install {id}` first, or set \
                      plugins.command.{id}.run = \"unattended\" in config."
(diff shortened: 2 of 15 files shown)


crates/jp_cli/src/cmd/query/turn_loop.rs
```diff
@@ -843,7 +844,7 @@ pub(super) async fn run_turn_loop(
                         mcp_client,
                         root,
                         &tool_renderer,
-                        is_tty,
+                        interactive,
```
JeanMertz: blocking: A user running `jp query ... >answer.txt 2>chrome.log` from a controlling terminal while a slow tool executes will have `is_tty = false` and `interactive = true` once the documented `/dev/tty` derivation lands. `execute_with_prompting` still uses this one argument both for prompt routing (`handle_tool_result`) and for the tool progress ticker (`if is_tty` at `coordinator.rs:984`), so passing `interactive` here writes `\r\x1b[K` progress control sequences into non-TTY chrome output. Thread both signals into `execute_with_prompting`: use `interactive` only for question/result prompts and `is_tty` for progress rendering.

crates/jp_cli/src/ctx.rs
```diff
@@ -82,8 +82,25 @@ pub(crate) struct Term {
...
+    /// format resolution, spinners, cursor control, OSC sequences.
+    /// For "can a user answer a prompt?", use [`Self::interactive`] ([RFD
+    /// 048]).
+    ///
+    /// [RFD 048]: https://jp.computer/rfd/048
     pub(crate) is_tty: bool,
 
+    /// Whether a user is present to answer prompts.
+    ///
+    /// Gates every interactive prompt: tool permissions, label resolution,
```
JeanMertz: question: Is “every interactive prompt” the intended contract? `confirm_and_remove` in `conversation/rm.rs` and `confirm_archive` in `conversation/archive.rs` still open prompts without consulting `Term::interactive`, so a future `--non-interactive` override would not suppress them. If this field only covers prompts that were formerly keyed off stdout TTY state, please narrow this documentation. Otherwise those prompt paths need to use the same signal.

crates/jp_cli/src/cmd/query/turn_loop_tests.rs
```diff
@@ -3178,6 +3199,7 @@ async fn test_waiting_indicator_shows_during_delay() {
             &mcp_client,
             root,
             true, // is_tty = true to enable the indicator
+            true, // interactive
```
JeanMertz: theoretical: No production input reaches a mixed state in this commit because `interactive` is initialized from `is_tty`, but the updated `run_turn_loop` tests also keep both values equal. The separation itself is therefore untested. This rendering-only test can pass `interactive = false` while retaining `is_tty = true`; a permission-prompt test should exercise the inverse pair. Those cases would catch either signal being threaded to the wrong consumer before the production derivations diverge.

crates/jp_cli/src/cmd/plugin/install.rs
```diff
@@ -48,7 +48,7 @@ impl Install {
 
         // Third-party plugins need explicit approval.
         if !plugin.official {
-            if !ctx.term.is_tty {
+            if !ctx.term.interactive {
```
JeanMertz: blocking: `jp plugin install <third-party> 2>prompt.log` is a plausible current input: stdout remains a terminal, so this gate passes, but the notice and `InlineSelect` below are both rendered through `std::io::stderr()`. The user sees no approval prompt and the command waits for input. Splitting promptability from output TTY state makes the channel choice matter here; route this prompt through `ctx.printer.prompt_writer()` (or the shared prompt backend) so redirection cannot hide it.

crates/jp_cli/src/cmd/plugin/dispatch.rs
```diff
@@ -1696,7 +1696,8 @@ pub(crate) async fn run_external(args: &[String], ctx: &mut Ctx) -> cmd::Output
     }
 
     let config = ctx.config();
-    let Some(binary) = resolve_plugin_binary(subcommand, &config.plugins, ctx.term.is_tty).await?
+    let Some(binary) =
+        resolve_plugin_binary(subcommand, &config.plugins, ctx.term.interactive).await?
```
JeanMertz: blocking: `jp <unapproved-plugin> 2>prompt.log` passes `interactive = true` while stdout is a terminal, but both approval branches reached through `resolve_plugin_binary` render their `InlineSelect` directly to stderr. The approval is hidden in the file while the process waits. Passing only the new boolean cannot enforce the prompt channel; this boundary also needs a prompt writer/backend so registry-install and `$PATH` approvals render through `/dev/tty`.

crates/jp_cli/src/cmd/lock.rs
```diff
@@ -69,7 +69,7 @@ impl<'a> LockRequest<'a> {
         Self {
             workspace: &ctx.workspace,
             handle,
-            is_tty: ctx.term.is_tty,
+            is_tty: ctx.term.interactive,
```
JeanMertz: blocking: If another process holds the conversation lock and the user runs `jp query ... 2>chrome.log`, stdout is still a terminal so this gate allows prompting. After the timeout, `prompt_contention` renders its `Select` through `printer.err_writer()`, leaving the user with an apparently hung command while the prompt sits in the redirected file. Use `prompt_writer()` for the contention menu; `interactive` should gate a prompt on the controlling-terminal channel, not a prompt on stderr.

crates/jp_cli/src/cmd/query/turn_loop.rs
JeanMertz: Fixed. `execute_with_prompting` takes both signals: `is_tty` stays on the progress ticker (`coordinator.rs:990`), `interactive` goes to `handle_tool_result`, whose parameter is renamed to match and whose two uses are the result prompt and the user-targeted question.

The divergence is closer than "once the `/dev/tty` derivation lands": #1072 sets `interactive = !non_interactive && is_tty`, so `jp query --non-interactive` at a terminal already produces the mixed pair, and the ticker would have gone silent there.

While in the area, every callee that receives `ctx.term.interactive` now names its parameter `interactive` rather than `is_tty` — `LockRequest`, `Composer`, `resolve_plugin_binary` and friends, `Resolver`, and the coordinator's permission path. The vocabulary was lying at each of those boundaries, which is how the ticker got wired to the wrong signal in the first place.

crates/jp_cli/src/cmd/query/turn_loop_tests.rs
JeanMertz: Taken, with a different shape.

`test_permission_prompt_follows_interactive_not_is_tty` runs the pair you'd get from `jp query > answer.txt` at a terminal (`is_tty = false, interactive = true`) against an Ask-mode tool whose prompt is answered `n`. The tool then runs only if the prompt never happened, so `mock output` in the response is the failure signal. I checked it red by pointing `resolve_tool_call_decision` and `run_permission_phase` at `is_tty`: it fails with `The prompt should have run and been declined: mock output`.

The inverse pair is on the two waiting-indicator tests rather than only the one at this line: `shows_during_delay` now runs `is_tty = true, interactive = false` and `not_shown_for_non_tty` runs `is_tty = false, interactive = true`. Together they pin the indicator to `is_tty` in both directions, which one test alone doesn't do.

No test for the tool progress ticker. Nothing covers it today, and building a slow-tool fixture for it is more machinery than the remaining risk justifies now that the parameter is named for what it gates.

crates/jp_cli/src/ctx.rs
JeanMertz: "Every interactive prompt" was too strong, and the doc now says what the field actually covers, with a line naming `conversation rm` and `conversation archive` as the confirmations that gate on `--force` / `--confirm` instead.

The behavioural half belongs in #1072, where the promise is made: `--non-interactive`'s help says "Anything JP would normally ask about resolves without asking", and `jp --non-interactive conversation rm <id>` still prompts. Either those two consult the signal, or that help text needs narrowing. Raising it there.

One more for that PR while it's in view: the same help says tool permission requests "decline". `decide_permission` returns `Approved` for `RunMode::Ask` and `Edit` when the signal is false (`coordinator.rs:763-766`), so `--non-interactive` runs ask-gated tools unprompted. That is today's no-terminal behaviour, unchanged by either PR, but the flag turns it into a documented promise that reads backwards.

crates/jp_cli/src/cmd/lock.rs
JeanMertz: Correct about the symptom, deferring the fix.

Two axes are in play. Which signal decides whether to prompt, and which channel the prompt renders on. This PR moves the first; the stderr rendering is the second, and the two are independent — `conversation rm`, `archive`, `title`, `label/resolve`, the tool prompter, and the interrupt handler already render on `prompt_writer()` under a mix of gates, while `lock.rs`, `plugin/{dispatch,install}.rs`, `bootstrap.rs:435`, `cmd/target.rs` (×4), and `workspace/target.rs` (×2) render on a raw stderr handle regardless of their gate. With `2>chrome.log` the contention menu is equally invisible before and after this commit.

So this is a half-finished RFD 048 migration, not something the split introduces or worsens. Fixing three of those ten sites here would leave the other seven, and two of them (`bootstrap::resolve`, `resolve_plugin_binary`) need a `&Printer` threaded through call chains that don't have one — a much larger diff than the rename this PR is.

Follow-up PR: route every prompt through the prompt channel, auditing all ten against RFD 048.

crates/jp_cli/src/cmd/plugin/dispatch.rs
JeanMertz: Same call as the lock contention menu: real, pre-existing, and going into the prompt-channel follow-up rather than here.

Confirmed both branches: `try_registry_install` (`dispatch.rs:1429-1440`) and `check_run_policy` (`dispatch.rs:1504-1515`) each build their own `std::io::stderr()`. And you're right that the boolean can't fix it — but the fix is bigger than swapping a writer. `resolve_plugin_binary` and both callees are free functions with no printer, so routing them through `/dev/tty` means threading a `&Printer` or `&dyn PromptBackend` down that chain. That is a signature change across the plugin resolution path, in a PR whose entire current diff is a rename plus one parameter.

The gate itself is right here: this line passes the promptability signal, which is what the split is for.

crates/jp_cli/src/cmd/plugin/install.rs
JeanMertz: Confirmed at `install.rs:47,58,68` — the notice and the `InlineSelect` both go to `std::io::stderr()`, so `2>prompt.log` hides the approval and the command waits.

Here the fix genuinely is one line (`ctx.printer.prompt_writer()`), which makes this the most tempting of the three to just do. Doing it alone is what I want to avoid: it would leave the same bug in the two `dispatch.rs` approval branches, `bootstrap.rs:435`, `cmd/target.rs`, and `workspace/target.rs`, with no test and no principle recorded about which sites were done and why. All of them go together in the prompt-channel follow-up.

crates/jp_cli/src/cmd/query/turn_loop_tests.rs
JeanMertz: theoretical: Production still derives both values from `stdout.is_terminal()`, so no current user input reaches this mixed pair. The new inverse test does not actually pin the rendering side, though: `test_waiting_indicator_not_shown_for_non_tty` discards `err`, while the waiting indicator is chrome written to stderr. Its stdout assertion still passes if the indicator is incorrectly enabled by `interactive`. Keep the stderr buffer and assert that it contains neither `Waiting…` nor `\r\x1b[K`.

crates/jp_cli/src/cmd/lock.rs
JeanMertz: blocking: The broader prompt-channel audit can be separate, but this touched path has a local fix and the named input still leaves JP waiting on an invisible prompt. The fact that it predates this refactor does not make the hidden failure safe to carry through a promptability migration. Route this contention menu through `prompt_writer()` here; untouched prompt sites can remain in the follow-up.

crates/jp_cli/src/cmd/plugin/install.rs
JeanMertz: blocking: The full audit is broader, but this prompt is already in a touched file and the fix is local. `2>prompt.log` still makes this command wait without showing the approval, so I would route this one through `ctx.printer.prompt_writer()` here and leave the untouched sites to the follow-up.

crates/jp_cli/src/cmd/plugin/dispatch.rs
JeanMertz: blocking: The plumbing cost is larger here, but this PR is changing this boundary's promptability input. Leaving it as only a boolean encodes whether prompting is allowed without carrying the channel on which that prompt can be seen. That cannot satisfy RFD 048 when the signals diverge. Thread a narrow prompt writer/backend through this path here, or leave plugin resolution on the old signal until the prompt-channel follow-up changes the boundary as a unit.

PR #1072, opened 2026-09-04
feat(cli): Add `--non-interactive` to suppress every prompt

### LINKED ISSUE #1068: `jp conversation ls` stopped to ask for user input while in a pipe
### JP version

10a03efc

### What happened?

Commands that will ever be in pipes cannot collect keystrokes.
The keystroke-friendly variant has to be a separate command.

JeanMertz:
`jp c ls | rg foo` could stop and ask which workspace to use. The
workspace-selection prompts from [RFD 087] fire whenever stdin is a
terminal, which a piped stdout does not change, so a command that only
prints a table blocked on a question the user never expected. The only
way out was redirecting stdin from `/dev/null`.

`--no-interactive` declares that nobody is available to answer.
Everything JP would ask about resolves the way it does when JP runs
without a terminal: the session layer is ignored and the workspace
picker fails with its candidates listed, the conversation pickers fail
with keyword help, the lock-timeout prompt fails instead of offering to
wait, and installing a third-party plugin fails rather than asking for
approval. A command with no answer to fall back on fails too: `jp init`,
and any removal or archive it would have to confirm.

    jp --no-interactive conversation ls | rg fix

Setting `JP_NONINTERACTIVE=1` applies the same to every invocation in a
script or CI job. Only `1` and `true` count, so `JP_NONINTERACTIVE=0`
turns the behaviour off rather than on.

One prompt does not decline: a tool whose `run` mode is `ask` or `edit`
runs unconfirmed when nobody can be asked. That branch predates this
flag — it read `stdout().is_terminal()` before [RFD 048]'s split, so a
piped `jp query` already auto-approved — and changing it changes every
non-interactive run, not just the flagged ones. It is [RFD 049]'s `deny`
policy and lands separately. The flag's help says what happens rather
than implying otherwise.

Three prompts gained a guard on the way, none of which consulted any
interactivity signal before: `jp init`, and the `conversation rm` and
`conversation archive` confirmations. A run with nobody watching either
blocked on the controlling terminal — `inquire` opens `/dev/tty` when
stdin is not a tty — or, for `archive`, read the failed prompt as "skip"
and reported success having archived nothing. Both confirmations now
fail and name `--no-confirm`. `init` writes nothing until the wizard has
every answer, so an interrupted run no longer leaves a directory
carrying a workspace ID and no config.

Output formatting is untouched: `--no-interactive` says nothing about
whether the consumer of stdout can render ANSI, which stays `--format`'s
question ([RFD 048]).

*The flag is `--no-interactive`; `--non-interactive` is accepted as an alias.*

Closes: #1068

[RFD 049]: https://jp.computer/rfd/049
[RFD 087]: https://jp.computer/rfd/087

### DIFF
--- a/crates/jp_cli/src/ctx.rs
+++ b/crates/jp_cli/src/ctx.rs
@@ -136,9 +136,7 @@ impl Ctx {
         let is_tty = io::stdout().is_terminal();
         let width = printer.output_width().columns();
 
-        // Same derivation as `is_tty` for now; the two diverge when the
-        // promptability signal moves to `/dev/tty` availability.
-        let interactive = is_tty;
+        let interactive = crate::interactive(args.non_interactive, is_tty);
 
         Self {
             exec,
--- a/crates/jp_cli/src/ctx.rs
+++ b/crates/jp_cli/src/ctx.rs
@@ -136,7 +136,7 @@ impl Ctx {
         let is_tty = io::stdout().is_terminal();
         let width = printer.output_width().columns();
 
-        let interactive = crate::interactive(args.non_interactive, is_tty);
+        let interactive = crate::interactive(args.no_interactive, is_tty);
 
         Self {
             exec,
--- a/crates/jp_cli/src/lib.rs
+++ b/crates/jp_cli/src/lib.rs
@@ -159,8 +159,8 @@ struct Globals {
     /// environment, such as a script or a CI job.
     ///
     /// This does not change output formatting; use `--format` for that.
-    #[arg(long, global = true)]
-    non_interactive: bool,
+    #[arg(long, global = true, visible_alias = "non-interactive")]
+    no_interactive: bool,
 
     /// The output format.
     #[arg(
@@ -349,22 +349,22 @@ pub(crate) enum CliFormat {
 
 /// Whether a user is present to answer prompts.
 ///
-/// A terminal is the evidence that someone is watching; `non_interactive` is
-/// the user overriding that evidence.
+/// A terminal is the evidence that someone is watching; `no_interactive` is the
+/// user overriding that evidence.
 /// Deliberately independent of whether output can carry ANSI escapes ([RFD
 /// 048]) — a piped `jp c ls` still has a user behind it.
 ///
 /// [RFD 048]: https://jp.computer/rfd/048
-const fn interactive(non_interactive: bool, terminal: bool) -> bool {
-    !non_interactive && terminal
+const fn interactive(no_interactive: bool, terminal: bool) -> bool {
+    !no_interactive && terminal
 }
 
 /// Whether a user is present to answer a workspace or conversation picker.
 ///
 /// Both resolve before a [`Ctx`] exists, so they cannot read
 /// `Term::interactive` and take the terminal signal from stdin themselves.
-pub(crate) fn stdin_interactive(non_interactive: bool) -> bool {
-    interactive(non_interactive, io::stdin().is_terminal())
+pub(crate) fn stdin_interactive(no_interactive: bool) -> bool {
+    interactive(no_interactive, io::stdin().is_terminal())
 }
 
 /// Whether `name` names an environment variable set to an opt-in value.
@@ -424,7 +424,7 @@ pub fn run() -> ExitCode {
 
     // Folded into the flag once here, so every consumer downstream reads one
     // field instead of re-reading the environment.
-    cli.globals.non_interactive |= env_opt_out("JP_NONINTERACTIVE");
+    cli.globals.no_interactive |= env_opt_out("JP_NONINTERACTIVE");
 
     let format = cli.globals.format.resolve(is_tty);
 
@@ -572,7 +572,7 @@ fn run_inner(cli: Cli, format: OutputFormat) -> Result<()> {
                 session.as_ref(),
                 cli.globals.persist,
                 cli.globals.workspace.as_ref(),
-                cli.globals.non_interactive,
+                cli.globals.no_interactive,
             )
             .map_err(Into::into);
 
@@ -593,7 +593,7 @@ fn run_inner(cli: Cli, format: OutputFormat) -> Result<()> {
         };
 
         return args
-            .run(&printer, cli.globals.non_interactive)
+            .run(&printer, cli.globals.no_interactive)
             .map_err(Into::into);
     }
 
@@ -607,7 +607,7 @@ fn run_inner(cli: Cli, format: OutputFormat) -> Result<()> {
     let exec = bootstrap::resolve(
         cli.globals.workspace.as_ref(),
         session.as_ref(),
-        cli.globals.non_interactive,
+        cli.globals.no_interactive,
     )?;
     trace!(
         root = %exec.root,
@@ -655,7 +655,7 @@ fn run_inner(cli: Cli, format: OutputFormat) -> Result<()> {
         &mut workspace,
         session.as_ref(),
         fs_backend.as_deref(),
-        stdin_interactive(cli.globals.non_interactive),
+        stdin_interactive(cli.globals.no_interactive),
     )?;
     let config = Arc::new(config);
     let runtime = build_runtime(cli.root.threads, "jp-worker")?;
--- a/crates/jp_cli/src/lib.rs
+++ b/crates/jp_cli/src/lib.rs
@@ -148,9 +148,12 @@ struct Globals {
     /// Assume no user is available to answer prompts.
     ///
     /// Anything JP would normally ask about resolves without asking: the
-    /// workspace-selection prompts, tool permission requests, plugin approval,
-    /// and lock-timeout handling all behave as they do when JP is run without a
-    /// terminal.
+    /// workspace and conversation pickers, tool permission requests, plugin
+    /// approval, and lock-timeout handling all behave as they do when JP is run
+    /// without a terminal.
+    ///
+    /// A command with no answer to fall back on fails instead: `jp init`, and
+    /// any removal or archive it would have to confirm.
     ///
     /// Set `JP_NONINTERACTIVE=1` to apply this to every invocation in an
     /// environment, such as a script or a CI job.
@@ -354,6 +357,14 @@ pub(crate) enum CliFormat {
 /// [RFD 048]: https://jp.computer/rfd/048
 const fn interactive(non_interactive: bool, terminal: bool) -> bool {
     !non_interactive && terminal
+}
+
+/// Whether a user is present to answer a workspace or conversation picker.
+///
+/// Both resolve before a [`Ctx`] exists, so they cannot read
+/// `Term::interactive` and take the terminal signal from stdin themselves.
+pub(crate) fn stdin_interactive(non_interactive: bool) -> bool {
+    interactive(non_interactive, io::stdin().is_terminal())
 }
 
 /// Whether `name` names an environment variable set to an opt-in value.
@@ -581,7 +592,9 @@ fn run_inner(cli: Cli, format: OutputFormat) -> Result<()> {
             unreachable!("every workspace-free command has a dedicated run path");
         };
 
-        return args.run(&printer).map_err(Into::into);
+        return args
+            .run(&printer, cli.globals.non_interactive)
+            .map_err(Into::into);
     }
 
     // The pre-workspace bootstrap (RFD 087): session identity and the
@@ -642,6 +655,7 @@ fn run_inner(cli: Cli, format: OutputFormat) -> Result<()> {
         &mut workspace,
         session.as_ref(),
         fs_backend.as_deref(),
+        stdin_interactive(cli.globals.non_interactive),
     )?;
     let config = Arc::new(config);
     let runtime = build_runtime(cli.root.threads, "jp-worker")?;
@@ -948,6 +962,11 @@ fn parse_error(error: cmd::Error, format: OutputFormat) -> (u8, String) {
 /// 6. Consume `default_id` so it doesn't leak into the runtime config.
 /// 7. Build the final [`AppConfig`].
 ///
+/// Step 3 can open the conversation picker, which happens here rather than in
+/// the command because the conversation it selects supplies a config layer.
+/// `interactive` decides whether that prompt is available; without it, a target
+/// that resolves to nothing fails with keyword help.
+///
 /// [RFD 038]: https://jp.computer/rfd/038
 pub(crate) fn resolve_config(
     command: &Commands,
@@ -956,6 +975,7 @@ pub(crate) fn resolve_config(
     workspace: &mut Workspace,
     session: Option<&jp_workspace::session::Session>,
     fs: Option<&FsStorageBackend>,
+    interactive: bool,
 ) -> Result<(
     AppConfig,
     Vec<jp_workspace::ConversationHandle>,
@@ -982,6 +1002,7 @@ pub(crate) fn resolve_config(
         session,
         default_id,
         command.allows_new_from_picker(),
+        interactive,
     )?;
     let handles = outcome.handles;
 
--- a/crates/jp_cli/src/lib.rs
+++ b/crates/jp_cli/src/lib.rs
@@ -145,6 +145,20 @@ struct Globals {
     #[arg(short = 'q', long, global = true)]
     quiet: bool,
 
+    /// Assume no user is available to answer prompts.
+    ///
+    /// Anything JP would normally ask about resolves without asking: the
+    /// workspace-selection prompts, tool permission requests, plugin approval,
+    /// and lock-timeout handling all behave as they do when JP is run without a
+    /// terminal.
+    ///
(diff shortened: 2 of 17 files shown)


crates/jp_cli/src/lib.rs
```diff
@@ -145,6 +145,20 @@ struct Globals {
...
+    /// workspace-selection prompts, tool permission requests, plugin approval,
+    /// and lock-timeout handling all behave as they do when JP is run without a
+    /// terminal.
+    ///
+    /// Set `JP_NONINTERACTIVE=1` to apply this to every invocation in an
+    /// environment, such as a script or a CI job.
+    ///
+    /// This does not change output formatting; use `--format` for that.
+    #[arg(long, global = true)]
+    non_interactive: bool,
```
JeanMertz: blocking: A user can run `jp --non-interactive init /tmp/project` from a terminal and still get every prompt in the init wizard. `run_inner` sends `Init` down the workspace-free path as `args.run(&printer)`, so this value never reaches `Init::run`; that function starts prompting only after it has created the workspace directory and stored its ID. The command can therefore block despite the explicit opt-out and leave a partially initialized workspace behind. Pass the interaction policy into the init path and fail before writing anything if the required model/tool choices cannot be resolved non-interactively.

crates/jp_cli/src/lib.rs
```diff
@@ -550,7 +591,11 @@ fn run_inner(cli: Cli, format: OutputFormat) -> Result<()> {
     trace!("Resolving session identity.");
     let session = session::resolve();
 
-    let exec = bootstrap::resolve(cli.globals.workspace.as_ref(), session.as_ref())?;
+    let exec = bootstrap::resolve(
+        cli.globals.workspace.as_ref(),
+        session.as_ref(),
+        cli.globals.non_interactive,
```
JeanMertz: blocking: `jp --non-interactive conversation show` still opens the conversation picker when run inside a workspace with no session-active conversation and `conversation.default_id = "ask"`. The flag reaches workspace bootstrap here, but conversation resolution happens later in `resolve_config`, before `Ctx` exists, and `cmd::target::{resolve_picker, resolve_multi_picker, ...}` still decide solely from `stdin().is_terminal()`. Thread the effective interaction policy through `resolve_config`/`resolve_request` so every pre-`Ctx` picker fails as it would without a terminal. Otherwise this prompt can hang the script or CI job that the flag is meant to protect.

crates/jp_cli/src/ctx.rs
```diff
@@ -132,9 +132,7 @@ impl Ctx {
         let is_tty = io::stdout().is_terminal();
         let width = printer.output_width().columns();
 
-        // Same derivation as `is_tty` for now; the two diverge when the
-        // promptability signal moves to `/dev/tty` availability.
-        let interactive = is_tty;
+        let interactive = crate::interactive(args.non_interactive, is_tty);
```
JeanMertz: blocking: `jp --non-interactive conversation rm <id>` still asks `Are you sure?` on a terminal because `confirm_and_remove` does not consult `ctx.term.interactive`; `confirm_archive` has the same gap for pinned, active, bulk, or `--confirm` archives. These direct confirmation paths bypass the policy computed here, so an otherwise deterministic command can wait for input indefinitely. Gate them on `ctx.term.interactive` and use the same safe non-terminal outcome (decline/error unless the user supplied `--yes` or `--no-confirm`). Add a regression test that makes prompting observable rather than testing only the boolean helper.

crates/jp_cli/src/lib.rs
```diff
@@ -145,6 +145,20 @@ struct Globals {
...
     quiet: bool,
 
+    /// Assume no user is available to answer prompts.
+    ///
+    /// Anything JP would normally ask about resolves without asking: the
+    /// workspace-selection prompts, tool permission requests, plugin approval,
+    /// and lock-timeout handling all behave as they do when JP is run without a
+    /// terminal.
+    ///
+    /// Set `JP_NONINTERACTIVE=1` to apply this to every invocation in an
```
JeanMertz: nit: The workflow guide still tells users to run `jp --no-prompt ...` and says stdout redirection automatically selects non-interactive mode (`docs/README/workflow-integration.md:13-17`). A user following that page gets an unknown-argument error, and the stdout claim contradicts the interaction/output split this flag documents. Update that example to `--non-interactive` and describe `JP_NONINTERACTIVE=1` while adding the public flag.

crates/jp_cli/src/lib.rs
JeanMertz: Confirmed and fixed. The five raw `io::stdin().is_terminal()` guards in `cmd/target.rs` (748, 786, 801, 1038, 1072) were the flag's real blind spot: they read the exact signal `--non-interactive` overrides, and `resolve_config` reaches them before `Ctx` exists.

`interactive` is now a parameter, threaded `run_inner` → `resolve_config` → `resolve_request` → `resolve_targets` → the four picker guards. `resolve_picker` / `resolve_archived_picker` are also called from `c use` and `c unarchive`, which pass the same stdin-derived value rather than `ctx.term.interactive` — those two share the pre-`Ctx` helpers, and `Term::interactive` is currently derived from *stdout*, which this PR's own `interactive()` doc calls the wrong signal for promptability. Keeping the pickers on stdin leaves them behaviour-preserving apart from the flag. `read_stdin_ids` keeps its raw check: that one asks whether data is piped in, not whether we can prompt.

Regression test in `target_tests.rs`: `a_picker_without_a_user_fails_instead_of_prompting`, covering the single- and multi-select paths. It also removes the ambient dependence those tests had — before this, running the suite from a terminal could have opened a real picker.

crates/jp_cli/src/ctx.rs
JeanMertz: Right that neither function consults `ctx.term.interactive`; both are gated now. Two corrections to the framing, and one thing worse than what you found.

Not a regression this PR introduces: `confirm_and_remove` and `confirm_archive` never consulted *any* interactivity signal, so `--non-interactive` matches the no-terminal behaviour exactly, which is "prompts anyway". `inquire` reaches the prompt through `crossterm::tty_fd()`, which falls back to opening `/dev/tty` when stdin is not a tty, so a redirected-stdin run blocks on the controlling terminal too. What's inaccurate is the flag's headline sentence, which I've narrowed.

The worse case is in `archive.rs`: with no `/dev/tty` at all (CI), `InlineSelect::prompt` errors, `_ => Ok(false)` reads that as "skip", and `jp c archive` reports success having archived nothing. Silent, not visible. `rm` was fine there — `Err(_) => Err(1.into())` fails loudly.

So both now error rather than decline, naming `--no-confirm` as the way through. Declining would have kept the silent-skip.

Declining the "make prompting observable" harness: both call sites use `jp_inquire` types directly rather than the `PromptBackend` seam, so observability means moving them onto that trait — its own change. The tests pin the branch and the exact message instead (`removing_without_a_user_fails_unless_confirmation_is_waived`, `a_required_confirmation_without_a_user_fails_rather_than_skipping`), including that `--no-confirm` still goes through without asking.

Behaviour note: these use `ctx.term.interactive`, which is stdout-derived until it moves to `/dev/tty`, so `jp c rm <id> > log` from a terminal now errors instead of prompting. A destructive command getting more conservative, visibly, with the flag to pass in the message.

crates/jp_cli/src/lib.rs
JeanMertz: Confirmed: `lib.rs:584` was `args.run(&printer)`, and `Init::run` writes the directory (`init.rs:47`) and the workspace ID (`:55`) before the first question at `:58`.

Took the first half of the fix: `Init::run` takes the flag and returns an error before `create_dir_all`, so nothing is written. Test `init_without_a_user_writes_nothing` asserts both the message and that the root does not exist.

Declining the second half here. "Resolve the model/tool choices non-interactively" needs `--model` and `--tools-run` to exist first, which is an open item in RFD 044 (`044-workspace-initialization.md:488`). Until they land, `jp --non-interactive init` can only ever fail — erroring up front is at least truthful, and it fixes the half-initialized workspace, which was reachable without the flag too (any environment where the wizard fails).

crates/jp_cli/src/lib.rs
JeanMertz: Correct on both counts — `--no-prompt` appears nowhere outside that line, and the stdout claim is exactly the conflation 469ddb5 set out to remove. Updated: `--non-interactive`, a `JP_NONINTERACTIVE=1` example, and the redirection line rewritten to say it switches off pretty-printing and nothing else.

Left the rest of the page alone: `jp --detach` on line 27 is not implemented either, and three of the six `docs/README/*.md` links in the root README point at files that don't exist. That page needs its own pass, not one folded into this PR.

crates/jp_cli/src/lib.rs
```diff
@@ -145,6 +145,23 @@ struct Globals {
     #[arg(short = 'q', long, global = true)]
     quiet: bool,
 
+    /// Assume no user is available to answer prompts.
+    ///
+    /// Anything JP would normally ask about resolves without asking: the
+    /// workspace and conversation pickers, tool permission requests, plugin
+    /// approval, and lock-timeout handling all behave as they do when JP is run
```
JeanMertz: blocking: A plausible input is a workspace with `conversation.tools.'*'.run = "ask"` and a `jp --non-interactive query` where the model requests a destructive tool. This sets `ctx.term.interactive` to false, but `ToolCoordinator::decide_permission` treats `!is_tty` plus `Ask` or `Edit` as `PermissionDecision::Approved`, so the tool runs without confirmation. That contradicts the PR description that permission requests decline, and the resulting workspace mutation persists without the approval the config requires. Non-interactive `Ask`/`Edit` calls need to be skipped or failed, or the detached-policy part of RFD 049 needs to land with this flag.

crates/jp_cli/src/lib.rs
```diff
@@ -145,6 +145,23 @@ struct Globals {
     #[arg(short = 'q', long, global = true)]
     quiet: bool,
 
+    /// Assume no user is available to answer prompts.
+    ///
+    /// Anything JP would normally ask about resolves without asking: the
```
JeanMertz: blocking: A plausible input is a long-running `jp --non-interactive query` receiving `SIGINT`, either from Ctrl-C or a script timeout. `run_turn_loop` always registers the streaming and tool interrupt handlers, whose default actions are `prompt`; neither handler consults this interaction policy before opening its menu. With a controlling terminal this still prompts, and without one the first signal becomes `PromptFailed` and leaves the operation running. Thread `interactive` into interrupt dispatch so a non-interactive run applies a deterministic non-prompt action or lets the signal reach shutdown.

crates/jp_cli/src/lib.rs
JeanMertz: blocking: The explicit flag case is covered, but the ambient no-terminal case still reaches the writes because `run_inner` passes the raw `non_interactive` flag rather than effective promptability. A plain `jp init /tmp/project` in CI or cron creates the directory and stores its workspace ID, then the first `inquire` call fails for lack of a controlling terminal. That leaves the same half-initialized workspace the new doc and test say is avoided. Pass an effective `interactive` value into `Init::run` and reject before writing when it is false.

crates/jp_cli/src/lib.rs
JeanMertz: nit: The replacement's “and nothing else” claim is still broader than the implementation. `Ctx::new` derives `term.interactive` from `stdout().is_terminal()`, so `jp query ... > out.txt` disables tool prompts, plugin approval, lock-timeout prompts, and other post-`Ctx` interaction; `jp c rm <id> > log` now errors rather than asking. Narrow this sentence to the workspace/conversation pickers, or defer the general claim until promptability moves to controlling-terminal availability.

crates/jp_cli/src/lib.rs
JeanMertz: Right that the guard only covers the flag, and that the doc comment claimed more than that. Fixed, but not by passing effective promptability in.

No available signal answers "can the wizard prompt". `inquire` reaches the terminal through `crossterm::tty_fd()`, which opens `/dev/tty` when stdin is not a tty, so gating on `stdin_interactive` would reject `jp init < /dev/null` from a terminal — a run that works today, and the workaround the reporter of #1068 said they had learned to use. Gating on stdout rejects `jp init > log`, which also works. Either one trades a working case for the CI case.

The half-initialized workspace was the actual defect, and it needed no signal: `init` now asks first and writes second. Nothing between the old `create_dir_all` and `id.store` fed the wizard — the three `ask_*` calls touch neither `root`, `storage`, nor `id` — so the writes moved into `Init::create_workspace`, which takes the answers as arguments and therefore cannot run before they exist. That covers every way the wizard can stop: CI, cron, Ctrl-C at the first question, EOF, an unreadable `$HOME`. The `Workspace::in_memory_with_id(...).with_backend(fs)` binding that sat in the middle of it went too; it was constructed, dropped, and `Workspace` has no `Drop`.

Doc comment narrowed to say the `--non-interactive` check is what errors. On tests: I did not add one that drives a failing wizard. Driving the real wizard from a test either opens a prompt on the developer's terminal or passes only because CI happens to have no `/dev/tty` — ambient either way. The ordering is a dataflow property now rather than something a test has to catch; `create_workspace_writes_the_id_and_the_collected_answers` pins what the write step produces.

crates/jp_cli/src/lib.rs
JeanMertz: Correct, and I contradicted myself: the reply two comments up says `jp c rm <id> > log` now errors instead of prompting, and then I wrote a doc line claiming stdout redirection changes nothing but pretty-printing. `Ctx::new` derives `interactive` from `stdout().is_terminal()` (`ctx.rs:132-135`) and every Ctx-side prompt reads it.

Narrowed to the half that is true:

```sh
# Redirecting stdout switches off pretty-printing; pass --non-interactive to
# also declare that nobody is there to answer a prompt
jp query "..." > out.txt
```

The general claim comes back when promptability moves to controlling-terminal availability, not before.

crates/jp_cli/src/lib.rs
JeanMertz: Confirmed, and this is the worst of the batch. `coordinator.rs:763-766` approves `Ask` and `Edit` outright, and it is reached with the interaction policy, not a rendering flag: `turn_loop.rs:672` → `resolve_tool_call_decision` → `decide_permission`. `run = "ask"` is what `jp init` writes when the user picks the option the wizard itself labels "(safest option)", so the input is as ordinary as it gets, and the tool's effect outlives the command.

Two corrections to where the fix goes.

The branch predates the flag. Before 469ddb5 the same code read `stdout().is_terminal()`, so `jp query "..." > out.txt` has auto-approved `ask` tools for as long as the branch has existed. Gating only the `--non-interactive` door leaves the piped door open — the door #1068 was actually about — and buys an inconsistency where the flag fails a tool call and a pipe approves it. The fix belongs on the branch, for every non-interactive run, which is RFD 049's `deny` policy (`049-...md:53` states the safe default outright) and wants its own tests and review.

What this PR did get wrong is the advertising, and that I have fixed. The description says permission requests "decline"; they do the opposite. Plugin approval does decline (`plugin/install.rs:51`) and the lock timeout does fail (`cmd/lock.rs:99`), but a user who reads that sentence will reach for `--non-interactive` in CI *because* they think it is the safer setting. The flag's help now says what happens instead of gesturing at "as without a terminal": the pickers, the lock-timeout prompt and third-party plugin approval all fail, and a tool set to ask before running runs unconfirmed.

The PR description itself still needs that line corrected — I can't edit it from here.

crates/jp_cli/src/lib.rs
JeanMertz: Facts all check out: registration is unconditional (`turn_loop.rs:209`, `:339`), both actions default to `Prompt` (`jp_config/src/interrupt.rs:127`, `:151`), and `handle_streaming_interrupt` branches on config alone (`handler.rs:243`). Not changing it here.

The consequence is milder than the comment reads, in both directions. Without a controlling terminal, `PromptFailed` is the designed outcome rather than a stuck run: the notice is dropped unresolved, the press keeps its place on the router's escalation ladder, and the second press begins a graceful shutdown (`handler.rs:147-152`, `signals.rs:8-13`). A script timeout sends SIGTERM, which cancels the shutdown token directly (`signals.rs:333-336`) and never reaches this path. With a terminal, the menu opens — but the user just pressed Ctrl-C, so they are at the keyboard; that is a different class of surprise from blocking a pipeline nobody is watching.

Fixing it means choosing *which* action `--non-interactive` substitutes, and that choice writes data: `tool_call.action = "stop"` records each tool's `cancellation_response` into the stream. That is the detached-prompt-policy half of RFD 049, and picking a default for it in a review round is how you end up with one nobody agreed to. The per-run escape hatch already exists: `--cfg interrupt.streaming.action=abort`, `--cfg interrupt.tool_call.action=stop`.

PR #1074, opened 2026-09-04
feat(cli): Announce runs that leave the current directory's workspace

JeanMertz:
A `jp w use` from days ago keeps redirecting every later command, and a session sticky to its active workspace does it without ever asking. The user stands in one workspace, the command operates on another, and nothing on screen says so; which matters most for `jp query`, where the workspace decides which provider the request goes to.

Runs resolved from the session-active workspace while the current directory belongs to a different one now open with a line on stderr:

    using session workspace /code/alpha (cwd: /code/beta)

The selected root carries the same bold yellow `jp w use` gives the workspace it switches to, and the directory it isn't is grey. Runs from inside the selected workspace, runs outside any workspace, and runs with an explicit `--workspace` stay silent: the first two have nothing to disagree with, and the third was typed this invocation.

The line is chrome, so `-q` suppresses it and under `--format json` it goes to stderr as plain
text, like every other chrome line; RFD 092 phase 1 moves NDJSON wrapping into
the printer's stderr methods and picks this up with the rest..

### DIFF
--- a/crates/jp_cli/src/lib.rs
+++ b/crates/jp_cli/src/lib.rs
@@ -609,6 +609,13 @@ fn run_inner(cli: Cli, format: OutputFormat) -> Result<()> {
         session.as_ref(),
         cli.globals.no_interactive,
     )?;
+    // A run that went somewhere other than the directory the user is standing
+    // in says so, rather than leaving a days-old `jp w use` to redirect the
+    // command silently.
+    if let Some(notice) = exec.foreign_root_notice() {
+        printer.eprintln(notice);
+    }
+
     trace!(
         root = %exec.root,
         source = ?exec.source,
--- a/crates/jp_cli/src/bootstrap.rs
+++ b/crates/jp_cli/src/bootstrap.rs
@@ -99,6 +99,13 @@ pub(crate) struct ExecutionContext {
     /// The selected checkout root.
     pub(crate) root: Utf8PathBuf,
 
+    /// The launch cwd's own workspace root, when it is inside one.
+    ///
+    /// `None` when `jp` was invoked outside every workspace.
+    /// Kept alongside [`Self::root`] so a run can tell that it operates on a
+    /// workspace other than the one the user is standing in.
+    cwd_root: Option<Utf8PathBuf>,
+
     /// The working directory for spawned children, when it differs from the
     /// process cwd.
     ///
@@ -132,6 +139,30 @@ impl ExecutionContext {
         self.child_cwd().unwrap_or(&self.launch_cwd)
     }
 
+    /// The chrome line announcing that this run operates on a workspace other
+    /// than the one the user is standing in.
+    ///
+    /// `None` unless the root came from the session-active workspace while the
+    /// launch cwd sits in a different one: an explicit `--workspace` was typed
+    /// this invocation, a cwd-derived root is the directory the user is in, and
+    /// standing outside every workspace leaves no competing answer to name.
+    pub(crate) fn foreign_root_notice(&self) -> Option<String> {
+        if self.source != RootSource::SessionActive {
+            return None;
+        }
+
+        let cwd_root = self.cwd_root.as_ref()?;
+        if same_dir(cwd_root, &self.root) {
+            return None;
+        }
+
+        Some(format!(
+            "using session workspace {} (cwd: {})",
+            self.root.to_string().bold().yellow(),
+            cwd_root.to_string().bold().grey(),
+        ))
+    }
+
     /// An execution context for a test-constructed workspace: as if `jp` was
     /// launched from the workspace root itself.
     #[cfg(test)]
@@ -139,6 +170,7 @@ impl ExecutionContext {
         Self {
             launch_cwd: workspace.root().to_owned(),
             root: workspace.root().to_owned(),
+            cwd_root: Some(workspace.root().to_owned()),
             child_cwd: None,
             source: RootSource::Cwd,
         }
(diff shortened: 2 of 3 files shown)


crates/jp_cli/src/lib.rs
```diff
@@ -596,6 +596,13 @@ fn run_inner(cli: Cli, format: OutputFormat) -> Result<()> {
         session.as_ref(),
         cli.globals.non_interactive,
     )?;
+    // A run that went somewhere other than the directory the user is standing
+    // in says so, rather than leaving a days-old `jp w use` to redirect the
+    // command silently.
+    if let Some(notice) = exec.foreign_root_notice() {
+        printer.eprintln(notice);
```
JeanMertz: blocking: A user running `jp --format json query ...` from workspace B while a sticky session points at workspace A gets this as plain text on stderr. `Printer::eprintln` only selects the error target and appends a newline; it does not call `wrap_json`, so the stderr stream is not valid NDJSON and a structured consumer can discard the warning or fail on the stream. Please send this through the JSON-aware `println` path with an error-targeted `PrintTask` (for example via `PrintableExt::to_err`) and cover the JSON stderr output in a regression test.

crates/jp_cli/src/lib.rs
```diff
@@ -596,6 +596,13 @@ fn run_inner(cli: Cli, format: OutputFormat) -> Result<()> {
         session.as_ref(),
         cli.globals.non_interactive,
     )?;
+    // A run that went somewhere other than the directory the user is standing
+    // in says so, rather than leaving a days-old `jp w use` to redirect the
+    // command silently.
+    if let Some(notice) = exec.foreign_root_notice() {
```
JeanMertz: nit: A user running `jp -q query ...` from workspace B while a sticky session points at workspace A still reaches this branch and writes the notice. `quiet` is passed to logging, but the `Printer` itself is not a sink and there is no quiet check here, so this contradicts the flag's "Suppress all output" contract and the PR description. Guard this chrome emission with `!cli.globals.quiet` and add a quiet-mode regression test.

crates/jp_cli/src/lib.rs
JeanMertz: The mechanism is as you describe: `Printer::eprintln` (`crates/jp_printer/src/printer.rs:254-259`) only sets the target and appends a newline, and `wrap_json` runs from `print`/`println` alone (lines 214-236). Under `--format json` this line reaches stderr as prose.

Not taking the call-site fix, and not treating it as blocking, for two reasons.

It isn't a property of this site. Every stderr chrome emission in the tree behaves the same way, including the two this line sits next to in a `jp --format json query` run: the non-standard-finish notice (`cmd/query/turn/coordinator.rs:343`) and the retry notice (`cmd/query/stream/retry.rs:257`, the non-TTY branch, which is exactly the piped case). A consumer parsing stderr as NDJSON already fails on those, so wrapping this one line makes one record valid in a stream that stays invalid. `to_err` also has no callers anywhere in the repo — this would be the only site using it, and the printer-level migration would then have to unpick it.

And the gap is already owned. RFD 092's output-discipline section says it outright: "the `Printer`'s stderr methods currently bypass NDJSON wrapping (only stdout prints wrap), which already falls short of RFD 048's 'chrome on stderr is NDJSON under `--format json`' rule; phase 1 extends the wrapping to stderr chrome" (`docs/rfd/092-...md:183-186`). That is one change in `eprintln`/`eprint` that converts every site at once.

On the regression test: `run_inner` builds `Printer::terminal(format)` internally (`lib.rs:542`), so its output isn't capturable in-process — the existing `run_inner` tests (`lib_tests.rs:559`, `:724`) assert on persisted state for that reason. Covering format behaviour here needs printer injection into `run_inner`, which is its own change.

What did change: the PR description claimed the line was an NDJSON record, which was wrong. It now says plain text, like every other chrome line, and points at RFD 092 phase 1.

crates/jp_cli/src/lib.rs
JeanMertz: Right that the line prints under `-q`, and right that the PR description claims otherwise — that clause is going.

Not guarding the emission, though. `--quiet` today means one thing: tracing off. `configure_logging` sets the level filter to `OFF` and drops the stderr log layer (`lib.rs:1344-1346`, `1403-1404`), and nothing else reads the flag. The one apparent exception, `cmd/conversation/grep.rs:221`, is grep's own `grep -q` semantics (short-circuit to an exit status), not a general rule. Even the run's error message ignores it: `lib.rs:444-452` `eprintln!`s it regardless, so "including errors" in the help text is already unmet.

So `!cli.globals.quiet` here would make this the only printer emission in the CLI honouring the flag, while `jp -q query` continues to print the assistant response, tool headers, retry notices and the finish notice. That's an arbitrary line to draw, and it hands the next chrome site the same coin-flip.

The flag/behaviour mismatch is real and worth fixing centrally, so I filed T-0dd9cpr for it: the `Printer` is the only place that sees every emission, `Printer::sink()` already exists, and the actual work is deciding whether `-q` means chrome-only (per RFD 048's channel split), everything (per today's help text), or graduated levels (which D15 lists as future work spanning the `Printer`, `ChatResponseRenderer`, and interactivity systems). The docs currently disagree with each other on this too — RFD 072 asserts the printer respects `--quiet`, D32 says `-q` suppresses chrome and not tracing, D15 says the opposite.

Cost of leaving it: a user running `jp -q` sees this one extra line on stderr, along with all the others `-q` doesn't suppress.

PR #1075, opened 2026-09-04
fix(printer): Wrap stderr chrome in NDJSON under --format json

JeanMertz:
[RFD 048] puts assistant output on stdout and chrome on stderr, and says
both render as NDJSON when `--format json` is set. Only stdout did.
A consumer running `jp -F json ... 2>&1 | jq` hit raw text among the
records and had to special-case the merged stream.

Chrome written with `eprintln` is now wrapped the same way `println`
wraps stdout, ANSI stripped and one record per line.

Scripts parsing JP's stderr as plain text while asking for
`--format json` see `{"message":"..."}` where they saw the bare line.
Text formats are unaffected.

### DIFF
--- a/crates/jp_printer/src/printer.rs
+++ b/crates/jp_printer/src/printer.rs
@@ -245,7 +245,7 @@ impl Printer {
         self.send(Command::Print(task));
     }
 
-    /// Print a fragment of chrome, with no trailing newline.
+    /// Print a fragment of chrome, with no trailing newline in text formats.
     ///
     /// In JSON mode the fragment becomes a whole record, as it does on stdout:
     /// `2>&1 | jq` has to parse everything it is given, so a fragment a
--- a/crates/jp_printer/src/printer.rs
+++ b/crates/jp_printer/src/printer.rs
@@ -245,12 +245,17 @@ impl Printer {
 
     /// Print a fragment of chrome, with no trailing newline.
     ///
-    /// Never JSON-wrapped, even in JSON mode: a fragment is half a line, and an
-    /// NDJSON record has to be a whole one.
-    /// Use [`Self::eprintln`] for anything a consumer should be able to read
-    /// back as a record.
+    /// In JSON mode the fragment becomes a whole NDJSON record, as it does on
+    /// stdout: `2>&1 | jq` has to parse every line, so a fragment a consumer
+    /// does not care about is one they filter out, not one that breaks the
+    /// stream.
     pub fn eprint<P: Printable>(&self, p: P) {
         let mut task = p.into_task();
+        if self.format.is_json() {
+            task = self.wrap_json(task);
+            task.content.push('\n');
+        }
+
         task.target = PrintTarget::Err;
         self.send(Command::Print(task));
     }
--- a/crates/jp_printer/src/printer.rs
+++ b/crates/jp_printer/src/printer.rs
@@ -243,16 +243,29 @@ impl Printer {
         self.send(Command::Print(task));
     }
 
-    /// Print error.
+    /// Print a fragment of chrome, with no trailing newline.
+    ///
+    /// Never JSON-wrapped, even in JSON mode: a fragment is half a line, and an
+    /// NDJSON record has to be a whole one.
+    /// Use [`Self::eprintln`] for anything a consumer should be able to read
+    /// back as a record.
     pub fn eprint<P: Printable>(&self, p: P) {
         let mut task = p.into_task();
         task.target = PrintTarget::Err;
         self.send(Command::Print(task));
     }
 
-    /// Print error followed by a newline.
+    /// Print a line of chrome.
+    ///
+    /// In JSON mode, the content is wrapped in an NDJSON envelope:
+    /// `{"message":"..."}\n`, matching [`Self::println`] so both streams carry
+    /// the same record shape.
     pub fn eprintln<P: Printable>(&self, p: P) {
         let mut task = p.into_task();
+        if self.format.is_json() {
+            task = self.wrap_json(task);
+        }
+
         task.content.push('\n');
         task.target = PrintTarget::Err;
         self.send(Command::Print(task));
--- a/crates/jp_printer/src/printer.rs
+++ b/crates/jp_printer/src/printer.rs
@@ -209,8 +209,9 @@ impl Printer {
     }
     /// Print content.
     ///
-    /// In JSON mode, the content is wrapped in an NDJSON envelope:
-    /// `{"message":"..."}\n`.
+    /// Under [`OutputFormat::Json`] the content becomes one NDJSON record,
+    /// `{"message":"..."}\n`; under [`OutputFormat::JsonPretty`] the same
+    /// record is indented across several lines.
     pub fn print<P: Printable>(&self, p: P) {
         let mut task = p.into_task();
         if self.format.is_json() {
@@ -223,8 +224,9 @@ impl Printer {
 
     /// Print content followed by a newline.
     ///
-    /// In JSON mode, the content is wrapped in an NDJSON envelope:
-    /// `{"message":"..."}\n`.
+    /// Under [`OutputFormat::Json`] the content becomes one NDJSON record,
+    /// `{"message":"..."}\n`; under [`OutputFormat::JsonPretty`] the same
+    /// record is indented across several lines.
     pub fn println<P: Printable>(&self, p: P) {
         let mut task = p.into_task();
         if self.format.is_json() {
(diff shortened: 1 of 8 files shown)


crates/jp_printer/src/printer.rs
```diff
@@ -243,16 +243,34 @@ impl Printer {
...
 
-    /// Print error followed by a newline.
+    /// Print a line of chrome.
+    ///
+    /// In JSON mode, the content is wrapped in an NDJSON envelope:
+    /// `{"message":"..."}\n`, matching [`Self::println`] so both streams carry
+    /// the same record shape.
     pub fn eprintln<P: Printable>(&self, p: P) {
         let mut task = p.into_task();
+        if self.format.is_json() {
```
JeanMertz: blocking: A transient provider failure while running `jp -F json query ... 2> >(jq .)` still corrupts stderr. `Ctx::term.is_tty` tracks stdout, so it remains true here; `StreamRetryState::notify` then writes `\r\x1b[K...` through `printer.err_writer()`, and `PrinterWriter` sends that `Err` task without passing through this wrapper. `jq` aborts on the raw fragment and drops the rest of the stream. Please route JSON-mode chrome away from `err_writer()` (or make that path record-aware), and add a regression test with JSON format plus `is_tty = true`.

crates/jp_printer/src/printer.rs
```diff
@@ -243,16 +243,34 @@ impl Printer {
...
+        }
+
         task.target = PrintTarget::Err;
         self.send(Command::Print(task));
     }
 
-    /// Print error followed by a newline.
+    /// Print a line of chrome.
+    ///
+    /// In JSON mode, the content is wrapped in an NDJSON envelope:
```
JeanMertz: nit: `is_json()` also covers `OutputFormat::JsonPretty`, and `wrap_json` renders that format across multiple lines, so this is not NDJSON for every mode described as “JSON mode.” The new `eprint` docs make the same claim. Please distinguish compact `Json` (one NDJSON record per line) from newline-terminated `JsonPretty` output.

JeanMertz: Correct, and fixed — but with wider scope than the comment asked for.

`print` (printer.rs:212) and `println` (printer.rs:226) carry the identical "In JSON mode, the content is wrapped in an NDJSON envelope" sentence and predate this PR, so fixing only the two new doc comments would have left the file saying two different things about the same `wrap_json` branch. All four now read:

> Under `OutputFormat::Json` the content becomes one NDJSON record, `{"message":"..."}\n`; under `OutputFormat::JsonPretty` the same record is indented across several lines.

Worth noting for anyone reading this later: the practical impact is nil. jq's parser is not line-based, so `JsonPretty` on stderr still parses fine — this was a docs-accuracy fix, not a behavior one.

crates/jp_printer/src/printer.rs
JeanMertz: The gap is real and the mechanism is right. Fixed, but not the way the comment proposes.

**Verified:** `PrinterWriter::write_str` (printer.rs:517-531) builds a `PrintTask` and sends it straight down the channel — no `wrap_json`, on either the `fmt::Write` or `io::Write` impl. `Ctx::term.is_tty` is `io::stdout().is_terminal()` (ctx.rs:115) and `CliFormat::resolve` (lib.rs:337) ignores it for an explicit `Json`, so `is_tty && format == Json` is reachable.

One detail off: the `err` `Sink` is `Strip` under `Json` (printer.rs:767, `Json` is not `is_pretty()`), so the escapes are gone before the consumer sees them. I confirmed the exact bytes by running the new test against the unfixed code — stderr held `⚠ rate_limit, retrying (1/3)…`, no `\r`, no trailing newline. Different bytes than named, same outcome: jq aborts and drops the remainder.

**Two more sites in the same class**, both gated on bare `is_tty`: `spawn_waiting_indicator` (turn_loop.rs:123) and the MCP startup line in `await_mcp_servers` (query.rs:1253). `chat.rs:722` is already safe — its timer is gated on `pretty_printing_enabled()` (chat.rs:333).

**Why not "make `err_writer()` record-aware".** It is a byte-stream `io::Write`, and every JSON-hostile caller uses it precisely to manipulate the cursor. There is no NDJSON encoding of "erase the line I just printed" — line-buffering it would produce a record whose content is a control sequence, which is not an improvement. The tree already picked the other answer twice: `turn_loop.rs:225` hands `ToolRenderer` a `Printer::sink()` under JSON, and `chat.rs:333` gates the reasoning timer on pretty. The invariant is *in-place redraw chrome is suppressed under JSON*, not *transcoded to JSON*.

**What changed:**

- retry.rs — new `can_redraw(is_tty, printer)` helper, `is_tty && !printer.format().is_json()`, used by both `clear_line` and `notify`. `notify` then falls through to `printer.eprintln(msg)`, which this PR already wraps. Keeping the check inside `notify` rather than at the `StreamRetryState::new` call site keeps the invariant in one place.
- turn_loop.rs:128 — `spawn_waiting_indicator` returns `None` under JSON too.
- query.rs:1253 — `config.show && is_tty && !printer.format().is_json()`.
- New `json_retry_notification_is_a_record_even_on_a_tty` in retry_tests.rs: `StreamRetryState::new(config, true)` against `Printer::memory(OutputFormat::Json)`, asserting stderr is exactly `{"message":"⚠ rate_limit, retrying (1/3)…"}\n` and that `line_active` stays false. Every existing helper there passes `is_tty: false`, so this is new coverage. Watched it fail against the unfixed gate before keeping it.

`jp_cli` 1398/1398 and `jp_printer` 38/38 pass.

**Left alone deliberately:** `lock.rs:169` writes an interactive `Select` prompt through `err_writer()`. Whether an interactive prompt should render at all under `--format json` is a different question from record shape, and answering it here would widen the PR past its subject.

On the `blocking:` label — the retry line was raw under JSON before this PR too, so it is not a regression. Taking it here anyway because the commit title claims the whole class, and leaving three redraw paths raw would have made that claim false.

JeanMertz: blocking: The new caller guards still miss other `spawn_line_timer` paths. A user running `jp -F json query ... 2> >(jq .)` with stdout on a TTY and the selected conversation locked reaches `spawn_lock_timer`, which passes `config.show` without checking the printer format; after the delay, `spawn_line_timer` writes the raw pending-lock line through `err_writer()` and `jq` aborts. `drain_background_tasks` has the same unguarded shape. Since `spawn_line_timer` is defined as a `\r`-based stderr renderer, please enforce the JSON exclusion there rather than at selected callers, and add a JSON-format test at that boundary.

crates/jp_printer/src/printer.rs
```diff
@@ -243,16 +245,38 @@ impl Printer {
         self.send(Command::Print(task));
     }
 
-    /// Print error.
+    /// Print a fragment of chrome, with no trailing newline.
```
JeanMertz: nit: This summary promises no trailing newline, but the JSON branch below appends one. Please qualify this as text-format behavior so the first sentence states the contract accurately, for example: “Print a fragment of chrome without a trailing newline in text formats.”

JeanMertz: Taken. printer.rs:248 now reads:

> Print a fragment of chrome, with no trailing newline in text formats.

The body already covers the JSON case, so nothing else moved.

Specific to `eprint`, for the record: `print` (printer.rs:214) appends the same newline inside its `is_json()` branch, but its summary is just "Print content." and promises nothing, so there was no contradiction to fix there.

crates/jp_printer/src/printer.rs
JeanMertz: Right on both paths, and right about the location. Moved the exclusion into `spawn_line_timer`.

**Verified both:**

- `spawn_lock_timer` (lock.rs:212) passes `config.show` with no format check. `acquire_lock` returns `Err(LockTimeout)` at lock.rs:111 when `!is_tty`, so the timer only spawns on a TTY — exactly the combination at issue. `LockWaitConfig::show` defaults to `true` (jp_config/src/style/lock_wait.rs:22), so it fires by default.
- `drain_background_tasks` sets `show_chrome = ctx.term.is_tty` (lib.rs:741) and spawns at lib.rs:747 and again at lib.rs:774 on the shutdown path. A first query registers a title-generation task, so an ordinary `jp -F json query "..."` on a TTY reaches it.

That is five production call sites — lock.rs:220, query.rs:1256, turn_loop.rs:135, lib.rs:747, lib.rs:774 — plus chat.rs:331, which is safe only because it happens to pass `pretty_printing_enabled()`. My previous reply claimed "two more sites in the same class." That count came from grepping `err_writer()` and `ErrChannel`, which misses every caller reaching `err_writer` indirectly through timer.rs:154/170. Grepping `spawn_line_timer` is the search that answers the question, and I did not run it.

**What changed:**

- timer.rs:126 — `if !show || printer.format().is_json()`, with the doc updated to say so. The function already owns the `Arc<Printer>`, its own doc defines it as a `\r`-based renderer that clears with `\r\x1b[K`, and every call site's format string opens with `\r\x1b[K`. The invariant belongs to the renderer, not to each caller.
- turn_loop.rs:128 and query.rs:1253 — removed the caller-side checks I added earlier, so the rule lives in exactly one place. Both keep their own `is_tty` gates, which are a separate concern.
- retry.rs `can_redraw` stays: `notify` writes through `err_writer()` directly, not through the timer, so the boundary guard does not cover it.
- New `test_line_timer_json_format_spawns_nothing` in timer_tests.rs, mirroring `test_line_timer_show_false_spawns_nothing`: `OutputFormat::Json` with `show = true` returns `None` and leaves stderr empty. Watched it fail against the unguarded version (`assertion failed: timer.is_none()`) before keeping it. One test at the boundary covers all five call sites.

`jp_cli` 1399/1399 and `jp_printer` 38/38 pass.

**One thing I left alone.** Grepping `\r\x1b[K` across `crates/` turns up one further family: prompter.rs:261 and prompter.rs:432. Those write through `prompt_writer()`, which targets `PrintTarget::Tty` when `/dev/tty` is available — a channel documented as never JSON-wrapped (printer.rs:876) — so they are a different question. `lock.rs:169` is the odd one out: it prompts through `err_writer()` where every other prompt uses `prompt_writer()`. That looks wrong on its own terms, but it is a channel-choice question rather than a record-shape one, so I would rather file it than widen this PR.

PR #1076, opened 2026-09-04
feat(cli): Add `jp workspace use --always` to pin a session

JeanMertz:
Pinning a session to its active workspace was reachable only by answering `A` at the cwd-vs-active conflict prompt, which meant the only way to stop being asked was to trigger the question first. Worse, the weaker form was the one users could state deliberately: `jp w use A` recorded a selection that the very next command run from another workspace would prompt about again.

`--always` states it up front. The selection outranks the current directory from then on, and JP stops asking:

    jp workspace use ~/code/alpha --always
    Session-active workspace set to /Users/me/code/alpha (always,
    ignoring the current directory)

A later `jp workspace use` without the flag releases the pin, so one invocation describes the whole intent rather than inheriting a pin set days earlier. `jp workspace use cwd` still clears the selection outright, and combining it with `--always` is rejected: there would be no selection left to pin to.

Pairs with the notice a run prints when it resolves away from the current directory, so a pinned session stays visible rather than silently redirecting later commands.

### DIFF
--- a/crates/jp_cli/src/cmd/workspace/use_.rs
+++ b/crates/jp_cli/src/cmd/workspace/use_.rs
@@ -34,7 +34,7 @@ pub(crate) struct Use {
     /// With it, the selection wins and JP stops asking; every run from
     /// elsewhere reports the workspace it used.
     ///
-    /// A later `jp workspace use` without `--always` releases the pin, and `jp
+    /// A later `jp workspace use` without `--always` releases it, and `jp
     /// workspace use cwd` clears the selection outright.
     #[arg(long)]
     pub(super) always: bool,
@@ -70,18 +70,22 @@ impl Use {
             );
         };
 
-        // `cwd` drops the record the pin would live on, so the two together
-        // ask for a selection that is both absent and permanent.
+        // `cwd` drops the record the sticky flag would live on, so the two
+        // together ask for a selection that is both absent and permanent.
         if self.always && matches!(target, WorkspaceTarget::Cwd) {
             return Err(format!(
-                "`{}` clears the session-active workspace, so there is nothing for `{}` to pin to.",
+                "`{}` clears the session-active workspace, so there is nothing for `{}` to keep \
+                 active.",
                 "jp workspace use cwd".bold().yellow(),
                 "--always".bold().yellow(),
             )
             .into());
         }
 
-        let previous = env.store.active(session);
+        let mapping = env.store.load(session);
+        let was_sticky = mapping.as_ref().is_some_and(|mapping| mapping.sticky);
+        let previous = mapping.and_then(|mapping| mapping.history.into_iter().next());
+        let suffix = sticky_suffix(self.always, was_sticky);
 
         match target::resolve(&target, env)? {
             ResolvedTarget::Help => unreachable!("handled before resolution"),
@@ -129,36 +133,37 @@ impl Use {
                 if previous.as_ref().is_some_and(|entry| {
                     entry.id().is_some_and(|prev| prev == id) && entry.root == selected.root
                 }) {
-                    // Re-selecting the same workspace still restates the pin:
+                    // Re-selecting the same workspace still restates the flag:
                     // `--always` on an already-active workspace is how a user
-                    // pins one they are standing in.
+                    // keeps the one they are standing in.
                     env.store.set_sticky(session, self.always)?;
 
                     printer.println(format!(
-                        "Already the session-active workspace: {}{}",
+                        "Already the session-active workspace: {}{suffix}",
                         selected.root.to_string().bold().yellow(),
-                        pin_suffix(self.always),
                     ));
                     return Ok(());
                 }
 
-                env.store
-                    .record_selection(session, &id, &selected.root, Utc::now())?;
-
                 // The invocation states the whole intent for the session: a
-                // selection made without `--always` is one made without a pin,
-                // including when an earlier selection left one behind.
-                env.store.set_sticky(session, self.always)?;
+                // selection made without `--always` is one made without the
+                // flag, including when an earlier selection left one behind.
+                env.store.record_selection_with_sticky(
+                    session,
+                    &id,
+                    &selected.root,
+                    Utc::now(),
+                    self.always,
+                )?;
 
                 let to = selected.root.to_string().bold().yellow();
-                let pinned = pin_suffix(self.always);
                 match previous {
                     Some(entry) => printer.println(format!(
-                        "Switched the session-active workspace from {} to {to}{pinned}",
+                        "Switched the session-active workspace from {} to {to}{suffix}",
                         entry.root.to_string().bold().grey(),
                     )),
                     None => {
-                        printer.println(format!("Session-active workspace set to {to}{pinned}"));
+                        printer.println(format!("Session-active workspace set to {to}{suffix}"));
                     }
                 }
             }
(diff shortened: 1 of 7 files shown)


crates/jp_cli/src/cmd/workspace/use_.rs
```diff
@@ -106,23 +129,37 @@ impl Use {
...
                     return Ok(());
                 }
 
                 env.store
                     .record_selection(session, &id, &selected.root, Utc::now())?;
 
+                // The invocation states the whole intent for the session: a
+                // selection made without `--always` is one made without a pin,
+                // including when an earlier selection left one behind.
+                env.store.set_sticky(session, self.always)?;
```
JeanMertz: blocking: A user who runs `jp w use alpha --always` while the sessions filesystem has room for the first atomic rewrite but not a second hits a partial commit: `record_selection` persists `alpha` with the old `sticky` value, then this call fails because `write_json` needs a full temporary copy. The command returns an error, but later commands observe the changed selection without the requested `--always` state. Persist the history update and `sticky` value through one `WorkspaceSessionStore` operation and one `write_json` call so this invocation is atomic.

crates/jp_cli/src/cmd/workspace/use_.rs
```diff
@@ -106,23 +129,37 @@ impl Use {
...
+                    // Re-selecting the same workspace still restates the pin:
+                    // `--always` on an already-active workspace is how a user
+                    // pins one they are standing in.
+                    env.store.set_sticky(session, self.always)?;
+
                     printer.println(format!(
-                        "Already the session-active workspace: {}",
+                        "Already the session-active workspace: {}{}",
                         selected.root.to_string().bold().yellow(),
+                        pin_suffix(self.always),
```
JeanMertz: nit: A user who runs `jp w use alpha --always` and later `jp w use alpha` to release it takes this branch. `set_sticky` changes the persisted state, but `Already the session-active workspace` reports a no-op and never confirms that `--always` was released. Keep the previous `sticky` value available and report the `true` to `false` transition explicitly, with a regression test for this same-target path.

crates/jp_cli/src/cmd/workspace/use_.rs
```diff
@@ -26,6 +26,18 @@ pub(crate) struct Use {
...
+
+    /// Keep using this workspace even from inside another one.
+    ///
+    /// Without it, standing in a different workspace makes JP ask which of the
+    /// two you meant.
+    /// With it, the selection wins and JP stops asking; every run from
+    /// elsewhere reports the workspace it used.
+    ///
+    /// A later `jp workspace use` without `--always` releases the pin, and `jp
+    /// workspace use cwd` clears the selection outright.
```
JeanMertz: nit: This adds a second way to clear `sticky`, but RFD 087 still says the `A` state is cleared with `jp w use cwd` and persists until no live root remains, while `WorkspaceSessionMapping::sticky` and `set_sticky` carry the same lifetime claim. A maintainer following that accepted contract can undo this behavior by preserving `sticky` on reselection. Update the RFD and the session-store docs with the new rule that any `jp w use` without `--always` releases it.

crates/jp_cli/src/cmd/workspace/use_.rs
```diff
@@ -58,6 +70,17 @@ impl Use {
             );
         };
 
+        // `cwd` drops the record the pin would live on, so the two together
+        // ask for a selection that is both absent and permanent.
+        if self.always && matches!(target, WorkspaceTarget::Cwd) {
+            return Err(format!(
+                "`{}` clears the session-active workspace, so there is nothing for `{}` to pin to.",
+                "jp workspace use cwd".bold().yellow(),
+                "--always".bold().yellow(),
```
JeanMertz: nit: `pin` belongs to the Pinned Conversation vocabulary. RFD 087 names this workspace-session state `sticky` and says the concepts use different words on purpose. A user who runs `jp w use cwd --always` sees this error using the conversation term for workspace behavior. Keep the help, error, helper name, and comments on `--always` or `sticky`, such as “nothing for `--always` to keep active.”

crates/jp_cli/src/cmd/workspace/use_.rs
JeanMertz: Accepted — the documented rule was stale, and `--always` appeared nowhere in `docs/`.

- RFD 087: added a `jp w use <target> --always` bullet to the command surface, and amended the `A` paragraph so the flag's lifetime reads "set by `A` or `--always`, released by any `jp w use` without `--always`, dropped with the record by `jp w use cwd`".
- `session_store.rs`: same rule on the `WorkspaceSessionMapping::sticky` field doc and on `set_sticky`.

One part I did not change: `record_selection`'s "The session-level `sticky` flag is preserved" is accurate for the store — the release happens in the CLI — and `session_store_tests.rs` names like `sticky_flag_survives_reselection` describe store behavior that still holds. The regression you describe would also be caught: `use_tests.rs` asserts the release on both the switch and the same-target paths.

crates/jp_cli/src/cmd/workspace/use_.rs
JeanMertz: Taking the wording change, but not the premise it rests on.

RFD 087 does not reserve "pin" for conversations — line 570 writes "the `sticky` pin" for *this* state, and line 397 calls the conversation one a "listing-pin". What the RFD reserves is the compound term **Pinned Conversation**, and there is no workspace cluster in `docs/architecture/ubiquitous-language/` yet to say otherwise.

The argument that does hold is narrower: the user-facing surface for this exact state already says **sticky** everywhere — `jp w show` renders a `Sticky` row (`format/workspace.rs:154`) and a note reading "this session is sticky to its active workspace" (`show.rs:346`), and the conflict prompt offers "keep the session sticky to it" (`bootstrap.rs:420`). A user who runs `--always` and then `jp w show` should not meet two words for one state.

So: the error now ends "…nothing for `--always` to keep active", `pin_suffix` is `sticky_suffix`, and the flag help says "releases it" rather than "releases the pin". I left "pin" in internal comments where it reads as an ordinary verb next to "sticky" — `bootstrap.rs:240` already does that, as does the RFD.

crates/jp_cli/src/cmd/workspace/use_.rs
JeanMertz: Accepted. The branch really did change the record while printing a no-op, and the existing `selecting_again_without_always_releases_the_flag` test switches between two workspaces, so it never reached this path.

`use_.rs` now loads the mapping once and keeps the previous flag alongside the previous selection:

```rust
let mapping = env.store.load(session);
let was_sticky = mapping.as_ref().is_some_and(|mapping| mapping.sticky);
let previous = mapping.and_then(|mapping| mapping.history.into_iter().next());
let suffix = sticky_suffix(self.always, was_sticky);
```

`sticky_suffix` reports the transition rather than the end state, so a release prints " (no longer always; the current directory applies again)". I applied it to the switch path too, not just the same-target one — both release the flag, and reporting it on one but not the other is the same gap in a different place. `reselecting_the_active_workspace_reports_the_release` covers the path you named; it fails against the old code, where the clause did not exist.

This also drops a redundant store read: `active()` loaded the whole mapping just to take `history[0]`.

crates/jp_cli/src/cmd/workspace/use_.rs
JeanMertz: Made the change, but not as a merge gate — the input as named does not hold up.

`set_sticky` returns early when the value is unchanged (`session_store.rs:200-202`), so the second write only happens when the flag actually flips; an ordinary `jp w use beta` is one write. And "room for the first atomic rewrite but not a second" needs free space to land in a band narrower than one ~400-byte record, since both writes need a full temp copy of near-identical content. That is a construction rather than an input a user produces. A kill between the two renames is the believable trigger, and re-running the command repairs it.

What justifies the change is the shape: `record_selection` documents that it preserves `sticky`, so any caller wanting to *state* the flag has to remember a two-step protocol. `use_.rs` is the only such caller today, which is when consolidating is still cheap.

I did not change `record_selection`'s signature — roughly 40 call sites, nearly all tests. Instead there is a private `write_selection(.., sticky: Option<bool>)` doing one load, one mutate, one `write_json`; `record_selection` passes `None`, and a new `record_selection_with_sticky` passes `Some(sticky)`. `use_.rs:145` is now that single call. Covered by `record_selection_with_sticky_writes_both_halves`.

PR #1079, opened 2026-09-04
fix(llm): Surface error messages from flat error envelopes

JeanMertz:
API errors from providers that return a flat error body now show the provider's own explanation instead of a bare status code. Asking Cerebras for an oversized prompt reported `Context window exceeded: HTTP 400`; it now reports `Current length is 200068 while limit is 131000`.

Error-body extraction only understood the nested
`{"error": {"message": "..."}}` envelope used by OpenAI, Anthropic and Google. Cerebras returns a flat `{"message": ..., "type": ..., "code": ...}` for every status it produces, so the message was dropped and the display fell back to `HTTP <code>`. Classification was unaffected, because the kind heuristics match against the raw body text rather than the extracted message; only the sentence the user reads was lost.

A body carrying both shapes is read as nested, on the grounds that a provider that nests is describing the failure there and using the outer key for something else.

### DIFF
--- a/crates/jp_llm/src/error_tests.rs
+++ b/crates/jp_llm/src/error_tests.rs
@@ -85,6 +85,70 @@ async fn a_429_outranks_a_context_window_phrasing_in_the_body() {
     assert!(err.is_retryable());
 }
 
+/// Cerebras returns a flat error envelope with no `error` wrapper.
+///
+/// Every status it produces uses this shape, so failing to read it renders
+/// every Cerebras failure as a bare `HTTP <code>` with the useful sentence
+/// discarded.
+#[tokio::test]
+async fn a_flat_error_envelope_reaches_the_message() {
+    let err = StreamError::from_eventsource(invalid_status(
+        404,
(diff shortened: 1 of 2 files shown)


crates/jp_llm/src/error_tests.rs
```diff
@@ -85,6 +85,70 @@ async fn a_429_outranks_a_context_window_phrasing_in_the_body() {
     assert!(err.is_retryable());
 }
 
+/// Cerebras returns a flat error envelope with no `error` wrapper.
+///
+/// Every status it produces uses this shape, so failing to read it renders
+/// every Cerebras failure as a bare `HTTP <code>` with the useful sentence
+/// discarded.
```
JeanMertz: nit: A Cerebras `429` with this flat envelope is a plausible exception to this claim. `from_eventsource` routes that status through `Self::rate_limit(retry_after)`, whose message is hardcoded to `Rate limited`, so the provider sentence is still discarded. Either narrow “every status” to the statuses whose `display` is retained, or preserve `display` in the rate-limit branch and add a `429` assertion.

PR #1080, opened 2026-09-04
fix(llm, cli, config): Jitter retry backoff so sessions don't collide

JeanMertz:
Retry waits now carry a random extra of up to a quarter of the delay, so several JP sessions sharing one API key stop resuming in the same instant after being rate limited together. Cerebras enforces its limits per organization and answers a 429 with `retry-after: 60`, so every session that hit the limit slept exactly 60000ms, woke together, and walked into the limit again — burning the whole retry budget without any of them getting through.

The `base_backoff_ms` documentation has described this jitter since the setting was introduced, down to a stated 0-500ms range, but nothing ever implemented it. That text reaches users through the generated `config.toml` and the exported JSON schema, so it was a promise the program did not keep. It now describes what the code does.

Jitter is added after the backoff ceiling rather than folded inside it. Folding it in means it disappears exactly when it matters, since every caller pinned at the ceiling gets the same number. It is also only ever added, never subtracted: a `retry_after` is the minimum the provider asked for, and waiting less walks straight back into the limit. One consequence worth knowing about is that an individual wait can now exceed `max_backoff_secs` by up to a quarter.

Both retry layers pick their delay through the new `jp_llm::retry_delay`, which replaces the identical `retry_after`-or-exponential match they each carried. `exponential_backoff` stays pure and keeps its exact tests.

### DIFF
--- a/crates/jp_config/src/assistant/request.rs
+++ b/crates/jp_config/src/assistant/request.rs
@@ -40,21 +40,30 @@ pub struct RequestConfig {
 
     /// Base delay for exponential backoff (in milliseconds).
     ///
-    /// The actual delay is calculated as:
+    /// Defaults to `1000`.
+    /// The delay doubles with each attempt:
     ///
     /// ```text
-    /// delay = min(base_backoff * 2^attempt + jitter, max_backoff)
+    /// delay = min(base_backoff_ms * 2^attempt, max_backoff_secs) + jitter
     /// ```
     ///
-    /// Where jitter is a random value between 0-500ms to prevent thundering
-    /// herd problems.
+    /// Jitter is a random extra of up to a quarter of the delay.
+    /// It keeps several JP sessions sharing one API key from resuming in the
+    /// same instant after they are rate limited together, which would just
+    /// trigger the limit again.
+    ///
+    /// When the provider says how long to wait, that delay is used instead of
+    /// the doubling one, and still gets the jitter.
     #[setting(default = 1000)]
     pub base_backoff_ms: u32,
 
     /// Maximum backoff delay (in seconds).
     ///
-    /// The backoff delay will never exceed this value, regardless of the number
-    /// of retry attempts.
+    /// Defaults to `60`.
+    /// Caps both the doubling delay and a wait the provider asked for, no
+    /// matter how many attempts have been made.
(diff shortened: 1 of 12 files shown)


crates/jp_config/src/assistant/request.rs
```diff
@@ -40,21 +40,30 @@ pub struct RequestConfig {
 
     /// Base delay for exponential backoff (in milliseconds).
     ///
-    /// The actual delay is calculated as:
+    /// Defaults to `1000`.
+    /// The delay doubles with each attempt:
     ///
     /// ```text
-    /// delay = min(base_backoff * 2^attempt + jitter, max_backoff)
+    /// delay = min(base_backoff_ms * 2^attempt, max_backoff_secs) + jitter
```
JeanMertz: nit: With the documented defaults, the first retry waits about 1000 ms, but this formula gives `1000 * 2^1` because the implementation treats `attempt` as 1-based and subtracts one before exponentiation. It also compares milliseconds with `max_backoff_secs` directly. A user tuning these fields gets the wrong schedule from the generated config documentation. Please use one unit and `2^(attempt - 1)`, or state in prose that the first retry uses `base_backoff_ms` and subsequent retries double it.

crates/jp_llm/src/retry.rs
```diff
@@ -128,10 +130,47 @@ pub async fn collect_with_retry(
...
+/// failed together do not resume together.
+///
+/// A `retry_after` the provider supplied is used as the delay, bounded by
+/// `max_backoff_secs`.
+/// Without one, the delay grows exponentially with `attempt`.
+/// Either way a random extra of up to a quarter of that delay is added on top.
+///
+/// The extra is added after the bound rather than folded inside it, so it still
+/// has an effect once the bound is reached.
+/// It is only ever added: `retry_after` is the minimum the provider asked for,
```
JeanMertz: nit: A plausible `retry_after = 600s` with `max_backoff_secs = 60` reaches this helper as a 60-second base and returns less than 75 seconds, so it does wait less than the provider asked and may hit the same limit again. That contradicts this rationale and the “provider-supplied delay is a minimum” test wording. If the configured cap remains authoritative, please say jitter is only added to the capped delay. If the provider value must be a minimum, `retry_after` cannot be capped below it.

PR #1081, opened 2026-09-04
fix(llm): Stop vague quota wording from making a 429 fatal

JeanMertz:
A rate limit whose body happens to say "quota exceeded" is no longer reported as exhausted billing. Before, such a response ended the turn with a fatal `Insufficient API quota` and discarded the `Retry-After` the provider sent, instead of waiting out a bucket that refills seconds later.

Quota detection runs on the response body, and two of its markers — `quota exceeded` and `resource_exhausted` — are also how providers word an ordinary rate limit. Google reports every 429 as `RESOURCE_EXHAUSTED` whether or not billing is involved, and Cerebras answers an exhausted per-minute token bucket with `token_quota_exceeded`, one character away from matching. A status code is the stronger signal, so on a 429 only a marker that names billing outright now counts: `insufficient_quota`, `billing_error`, `credit balance is too low`, and the credit wording OpenRouter uses. OpenAI reports a drained account as a 429 `insufficient_quota`, so that case stays fatal.

Away from a 429 there is no authoritative status to defer to and all markers still classify, so 402 and 403 responses are unaffected. This brings the shared classifier in line with the Google provider, which already checked its 429 first and used a typed `quotaId` rather than prose to separate a spent daily allowance from a refilling one.

### DIFF
--- a/crates/jp_llm/src/error.rs
+++ b/crates/jp_llm/src/error.rs
@@ -574,8 +574,8 @@ impl PartialEq for ToolError {
 /// - Google: `"RESOURCE_EXHAUSTED"`, `"Quota exceeded"`
 /// - OpenRouter: `"insufficient credits"`, `"out of credits"`
 ///
-/// `"quota exceeded"` and `"resource_exhausted"` are also how providers word an
-/// ordinary rate limit.
+/// `"quota exceeded"`, `"resource_exhausted"` and `"insufficient quota"` are
+/// also how providers word an ordinary rate limit.
 /// Where a status code already says the failure is one, reach for
 /// [`looks_like_billing_exhaustion`] instead.
 pub(crate) fn looks_like_quota_error(text: &str) -> bool {
@@ -584,6 +584,7 @@ pub(crate) fn looks_like_quota_error(text: &str) -> bool {
     looks_like_billing_exhaustion(text)
         || lower.contains("quota exceeded")
         || lower.contains("resource_exhausted")
+        || lower.contains("insufficient quota")
 }
 
 /// The subset of quota markers that name a billing or credit problem outright.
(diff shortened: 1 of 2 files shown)


crates/jp_llm/src/error.rs
```diff
@@ -540,16 +551,34 @@ impl PartialEq for ToolError {
...
+///
+/// A drained account and an exhausted rate-limit bucket both arrive as HTTP 429
+/// from OpenAI, and only the wording separates them.
+/// So every marker here has to be one no provider uses for a bucket that
+/// refills on its own.
+pub(crate) fn looks_like_billing_exhaustion(text: &str) -> bool {
+    let lower = text.to_ascii_lowercase();
+
     lower.contains("insufficient_quota")
         || lower.contains("insufficient quota")
```
JeanMertz: question: Is the space-separated `insufficient quota` marker intentionally part of the strict 429 subset? Unlike the typed `insufficient_quota` code, this is untyped prose and does not name billing or credit, and it is absent from the strict-marker list in the PR description. A provider describing a refillable service quota this way would still produce a fatal `InsufficientQuota` and lose its `Retry-After`. Unless an observed drained-account response requires this phrase on the 429 path, could we move it to `looks_like_quota_error` beside the other ambiguous prose and add a regression case?

JeanMertz: Not intentional, and you've found a real inconsistency. Moved in `3c6743e3`.

All three of your observations hold up. `insufficient quota` with a space appears **exactly once** in the workspace — the line you commented on. Every other in-tree reference is the underscore form: the doc comment above credits `"insufficient_quota"` to OpenAI, `openai.rs:1527-1528` matches it as a typed `code`/`type` field, and `openai_tests.rs` uses it. `git log -S` traces the space variant to #402, the original resilient-streaming commit, where it was added beside the typed code with the documentation crediting only the underscore. It has never had a named producer.

Which means it failed the bar this PR's own doc comment sets:

> every marker here has to be one no provider uses for a bucket that refills on its own

That test is unsatisfiable for a phrase no provider is known to use for anything. And your point about it not naming billing is the sharp version: every other strict marker names money outright (`credits`, `billing_error`, `credit balance is too low`), while `insufficient_quota` earns its place by being a *typed code*. Remove the underscore and you remove the only qualifying property — what's left is the ambiguous word `quota` with a qualifier, which is the shape of a refilling bucket as readily as a drained account. I've written that reasoning into `looks_like_billing_exhaustion`'s docs so the next person adding a marker has the rule spelled out.

Worth being straight about severity: I can't name a provider that emits this, so it wasn't a demonstrated bug. What tipped it is that the fix *deletes* a speculative marker rather than adding defensive code, and the risk is asymmetric in the direction you describe. Kept, a wrong guess ends the turn and discards `Retry-After`. Moved, a wrong guess costs up to five retries and then surfaces correctly — recoverable and visible. The one attested drained-account-on-429 case is OpenAI's typed code, which is untouched.

Regression case added as a third body in `a_429_outranks_vague_quota_phrasing_in_the_body`, which already asserts kind, retryability, and surviving retry timing:

```rust
r#"{"message":"Insufficient quota for this service tier."}"#,
```

I ran it against the unfixed code first and it failed on exactly that body (`left: InsufficientQuota, right: RateLimit`) with the other two still green.

Nothing else moves. The permissive check is a superset, so 402/403 still classify on the phrase, and the in-stream paths in `openai.rs::classify_stream_error` and `google.rs` consult it directly and are unaffected. `a_429_yields_to_an_explicit_billing_marker` still genuinely exercises the strict path — its body reads "exceeded your current quota", which contains neither `quota exceeded` nor a space-separated `insufficient quota`, so it isn't passing by accident through another marker.

One bonus: the PR description already listed the strict set without this phrase. That omission was the correct instinct that hadn't reached the code, and the description is now accurate as written.

PR #1082, opened 2026-09-04
refactor(llm): Move shared chat-completions types to their own module

JeanMertz:
`cerebras` and `llamacpp` both speak the OpenAI `/v1/chat/completions` dialect, and both need the same chunk types and assistant-message merging to do it. Those lived in `llamacpp`, so `cerebras` reached into a sibling provider for them and the shared thing carried the name of one of its two users.

They now live in `provider::openai_compat`, named for the protocol instead: the SSE chunk shapes, their `reasoning_content` extension, and `merge_consecutive_assistant_messages`. Both providers import from it and neither imports from the other. The module doc says what the dialect is and how it differs from the Responses API that the `openai` provider speaks, so the next contributor adding an OpenAI-compatible provider has somewhere obvious to look.

Behavior is unchanged; this is a move plus the import rewiring it forces. `parameters_with_strict_mode` is the remaining shared helper still living in `openai`, used by four providers including the two here. Leaving it alone: it is reached from a module that genuinely is about OpenAI, and moving it would drag its test module along for no gain here.

### DIFF
--- a/crates/jp_llm/src/provider.rs
+++ b/crates/jp_llm/src/provider.rs
@@ -7,7 +7,7 @@ pub mod llamacpp;
 pub mod mock;
 pub mod ollama;
 pub mod openai;
-pub mod openai_compat;
+pub(crate) mod openai_compat;
 pub mod openrouter;
 
 use std::sync::atomic::{AtomicU64, Ordering};
--- a/crates/jp_llm/src/provider.rs
+++ b/crates/jp_llm/src/provider.rs
@@ -7,6 +7,7 @@ pub mod llamacpp;
 pub mod mock;
 pub mod ollama;
 pub mod openai;
+pub mod openai_compat;
 pub mod openrouter;
 
 use std::sync::atomic::{AtomicU64, Ordering};
--- a/crates/jp_llm/src/provider/openai_compat.rs
+++ b/crates/jp_llm/src/provider/openai_compat.rs
@@ -1,0 +1,107 @@
+//! The OpenAI chat-completions wire format, shared by the providers that speak
+//! it.
+//!
+//! This is the `/v1/chat/completions` dialect: SSE chunks carrying
+//! `choices[].delta`, terminated by a `[DONE]` sentinel.
+//! The `llamacpp` and `cerebras` providers both stream it.
+//! It is a different protocol from the OpenAI Responses API that the `openai`
+//! provider uses, which has its own typed event enum.
+//!
+//! The chunk types mirror llama.cpp's `common_chat_msg_diff_to_json_oaicompat`
+//! output.
+//! Their addition over the plain OpenAI shape is `reasoning_content`, which
+//! carries extracted reasoning for llama.cpp's `--reasoning-format deepseek`
+//! (default) and `deepseek-legacy` modes; Cerebras spells the same field
+//! `reasoning`, and a serde alias accepts both.
+//!
+//! Deserialization is deliberately lenient: every field defaults, and unknown
+//! ones are ignored, so a provider adding a field to a chunk does not break the
+//! stream.
+
+use serde::Deserialize;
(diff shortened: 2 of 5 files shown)


crates/jp_llm/src/provider.rs
```diff
@@ -7,6 +7,7 @@ pub mod llamacpp;
 pub mod mock;
 pub mod ollama;
 pub mod openai;
+pub mod openai_compat;
```
JeanMertz: nit: `openai_compat` is an internal protocol helper, and every item it contains is `pub(crate)`. Publishing the module adds an externally nameable but unusable path to `jp_llm`'s API. Can this be `mod openai_compat;` (or `pub(crate) mod` if crate-wide access is intended) so the new boundary stays internal?

crates/jp_llm/src/provider/openai_compat.rs
```diff
@@ -0,0 +1,107 @@
...
+//! The chunk types mirror llama.cpp's `common_chat_msg_diff_to_json_oaicompat`
+//! output.
+//! Their addition over the plain OpenAI shape is `reasoning_content`, which
+//! carries extracted reasoning for llama.cpp's `--reasoning-format deepseek`
+//! (default) and `deepseek-legacy` modes; Cerebras spells the same field
+//! `reasoning`, and a serde alias accepts both.
+//!
+//! Deserialization is deliberately lenient: every field defaults, and unknown
+//! ones are ignored, so a provider adding a field to a chunk does not break the
+//! stream.
```
JeanMertz: nit: The leniency claim is broader than the type definition: `StreamChoice::delta` is required, so a choice without `delta` fails deserialization and both callers discard that chunk after logging a warning. Please either say that optional fields default and unknown fields are ignored, or add `#[serde(default)]` to `delta` if missing deltas are meant to be accepted.

crates/jp_llm/src/provider.rs
JeanMertz: Agreed, changed to `pub(crate) mod openai_compat;`.

Every item in the module is `pub(crate)`, so `pub mod` published a path nothing downstream can use, plus an empty page in `cargo doc`. The `pub mod` neighbours aren't a counter-precedent: `llamacpp`, `openai` and `mock` each export a genuinely `pub` provider type (`mock`'s doc example even shows `use jp_llm::provider::mock::MockProvider`).

Took `pub(crate)` over bare `mod` because it mirrors the items' own visibility and makes the absolute path in `cerebras_tests.rs` (`crate::provider::openai_compat::StreamChunk`) obviously legal, rather than resting on the private-item-visible-to-descendants rule. `cargo check` and the 703 `jp_llm` tests pass.

crates/jp_llm/src/provider/openai_compat.rs
JeanMertz: Correct on the facts, and I fixed the doc rather than the type.

`delta` is indeed required, and the fallout is real: `cerebras.rs:807-815` and `llamacpp.rs:241-249` both `warn!` and return `Ok(vec![])`, so one unparseable choice drops the whole chunk. The module doc now reads:

```rust
//! Deserialization is deliberately lenient: unknown fields are ignored and the
//! optional fields default, so a provider adding a field to a chunk does not
//! break the stream.
//! `StreamChoice::delta` is the one required field; a chunk whose choice omits
//! it fails to parse, and both providers log a warning and skip that chunk.
```

I left the type alone. Adding `#[serde(default)]` to `delta` is a behavior change in a PR whose contract is "behavior is unchanged; this is a move plus the import rewiring it forces", and the change it buys is close to nothing: a defaulted empty `StreamDelta` emits no events, so the only observable difference from skipping the chunk is the lost warning. If you want the leniency to be total, that's its own one-line commit (`StreamDelta` already derives `Default`).

PR #1083, opened 2026-09-04
feat(llm): Log chat-completions chunks that get dropped

JeanMertz:
A stream chunk that JP discards now leaves a record. An error reported inside the stream logs a warning naming the provider and the payload; a chunk with nothing to emit logs at debug. Both cases previously vanished without a trace at any log level a user is likely to have on.

The chunk types ignore fields they do not declare, and `choices` defaults to empty, so a payload like `{"error": {...}}` deserialized into a well-formed chunk with no choices and produced no events. The stream would then end without a terminal event, which the retry layer reads as a dropped connection: it resends the whole request several times and finally tells the user the connection failed, with the provider's actual explanation never printed.

No provider on this dialect is known to send such a chunk. Nine attempts to force one out of Cerebras — oversized prompts, malformed tool schemas, exhausted token buckets mid-stream, flex-tier preemption — produced an HTTP status before the stream opened every time, and llama.cpp is a local server with no such failure path. So this classifies nothing and changes no behavior; it exists to be seen if the assumption ever breaks, which is cheaper than diagnosing it from a blank retry loop.

Parsing moves behind `openai_compat::parse_chunk`, replacing the identical parse-and-warn block each of the two providers carried.

### DIFF
--- a/crates/jp_llm/src/provider/openai_compat.rs
+++ b/crates/jp_llm/src/provider/openai_compat.rs
@@ -22,11 +22,22 @@
 
 use serde::Deserialize;
 use serde_json::{Value, json};
+use tracing::{debug, warn};
 
 #[derive(Debug, Deserialize)]
 pub(crate) struct StreamChunk {
     #[serde(default)]
     pub choices: Vec<StreamChoice>,
+
+    /// An error the provider reported inside the stream, once the response had
+    /// already returned 200.
+    ///
+    /// Held as raw JSON because nothing reads it.
+    /// Every failure these providers are known to produce arrives as an HTTP
+    /// status before the stream opens, so this field exists for [`parse_chunk`]
+    /// to say something out loud if that ever stops being true.
+    #[serde(default)]
+    pub error: Option<Value>,
 }
 
 #[derive(Debug, Deserialize)]
@@ -64,6 +75,41 @@ pub(crate) struct FunctionDelta {
     pub name: Option<String>,
     #[serde(default)]
     pub arguments: Option<String>,
+}
+
+/// Parse one SSE `data:` payload from `provider` into a chunk.
+///
+/// Returns `None` when the payload yields nothing to emit, having logged the
+/// reason.
+pub(crate) fn parse_chunk(data: &str, provider: &str) -> Option<StreamChunk> {
+    let chunk: StreamChunk = match serde_json::from_str(data) {
+        Ok(chunk) => chunk,
+        Err(error) => {
+            warn!(provider, %error, data, "Failed to parse chunk.");
+            return None;
+        }
+    };
+
+    // Nothing downstream can act on this, so at least record it. A stream that
+    // reports a failure this way and then stops has sent no terminal event,
+    // which reads to the retry layer as a dropped connection: it resends the
+    // request blind, several times, and the user is told the connection failed.
+    if let Some(error) = &chunk.error {
+        warn!(
+            provider,
+            error = %error,
+            "Provider reported an error inside the stream. It will surface as a \
+             truncated response instead of this message."
+        );
+        return None;
+    }
+
+    if chunk.choices.is_empty() {
+        debug!(provider, data, "Chunk carried no choices.");
+        return None;
+    }
+
+    Some(chunk)
 }
 
 /// Merge consecutive assistant messages in an OpenAI-compatible
@@ -107,3 +153,7 @@ pub(crate) fn merge_consecutive_assistant_messages(messages: Vec<Value>) -> Vec<
             acc
         })
 }
+
+#[cfg(test)]
+#[path = "openai_compat_tests.rs"]
+mod tests;
--- a/crates/jp_llm/src/provider/openai_compat.rs
+++ b/crates/jp_llm/src/provider/openai_compat.rs
@@ -32,7 +32,8 @@ pub(crate) struct StreamChunk {
     /// An error the provider reported inside the stream, once the response had
     /// already returned 200.
     ///
-    /// Held as raw JSON because nothing reads it.
+    /// Held as raw JSON because nothing interprets the provider's error schema;
+    /// [`parse_chunk`] only checks whether it is present and logs it verbatim.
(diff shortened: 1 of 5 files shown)


crates/jp_llm/src/provider/openai_compat.rs
```diff
@@ -64,6 +75,41 @@ pub(crate) struct FunctionDelta {
...
+        warn!(
+            provider,
+            error = %error,
+            "Provider reported an error inside the stream. It will surface as a \
+             truncated response instead of this message."
+        );
+        return None;
+    }
+
+    if chunk.choices.is_empty() {
```
JeanMertz: blocking: llama.cpp's `to_json_oaicompat_chat` emits an initial `{"choices":[{"delta":{"role":"assistant","content":null}, ...}]}` chunk. `choices` is nonempty, so this returns `Some`; both handlers ignore `role` and null content, produce an empty event vector, and emit no debug record. The OpenAI schema also permits deltas containing only a role or nothing. That leaves a common dropped-chunk path unobserved, contrary to this PR's contract. Please log after each handler has built `events` when it is empty, or have this helper reject choices with no fields the handlers can emit, and add this payload as a regression test.

crates/jp_llm/src/provider/openai_compat.rs
```diff
@@ -64,6 +75,41 @@ pub(crate) struct FunctionDelta {
...
+    // Nothing downstream can act on this, so at least record it. A stream that
+    // reports a failure this way and then stops has sent no terminal event,
+    // which reads to the retry layer as a dropped connection: it resends the
+    // request blind, several times, and the user is told the connection failed.
+    if let Some(error) = &chunk.error {
+        warn!(
+            provider,
+            error = %error,
+            "Provider reported an error inside the stream. It will surface as a \
+             truncated response instead of this message."
```
JeanMertz: nit: For an `{"error": ...}` chunk followed by EOF, the query loop synthesizes `provider stream ended without a terminal event`, retries, and eventually reports that stream error when the budget is exhausted. It does not return a truncated response. If `[DONE]` follows, the response may complete normally. Avoid predicting either outcome here, for example: `Provider reported an error inside the stream; dropping this chunk.`

crates/jp_llm/src/provider/openai_compat.rs
```diff
@@ -20,11 +20,22 @@
...
 
 #[derive(Debug, Deserialize)]
 pub(crate) struct StreamChunk {
     #[serde(default)]
     pub choices: Vec<StreamChoice>,
+
+    /// An error the provider reported inside the stream, once the response had
+    /// already returned 200.
+    ///
+    /// Held as raw JSON because nothing reads it.
```
JeanMertz: nit: `parse_chunk` does read this field to branch and log it, so “nothing reads it” is misleading. The intended contract seems to be that JP does not interpret the provider's error schema. `Stored as raw JSON because only its presence and serialized value are used` would state that directly.

JeanMertz: Agreed — the line contradicts itself three lines later, where the doc says the field exists for `parse_chunk` to log. Reworded to state what is actually meant:

```rust
/// Held as raw JSON because nothing interprets the provider's error schema;
/// [`parse_chunk`] only checks whether it is present and logs it verbatim.
```

crates/jp_llm/src/provider/openai_compat.rs
JeanMertz: Correct, and the message contradicts the block comment right above it. On EOF the synthetic `StreamError::transient("provider stream ended without a terminal event")` (`turn_loop.rs:375-379`, `retry.rs:98`) routes through the retry budget and surfaces as a stream error; if `[DONE]` follows, `Finished(Completed)` is emitted and the turn ends normally. Predicting either was wrong.

Taken as suggested:

```rust
"Provider reported an error inside the stream; dropping this chunk."
```

The retry-loop consequence stays in the `//` comment above, which describes it accurately.

crates/jp_llm/src/provider/openai_compat.rs
JeanMertz: The observation holds. `server_task_result_cmpl_partial::to_json_oaicompat_chat` emits `add_delta({{"role","assistant"},{"content",nullptr}})` whenever `first || is_progress`, so it arrives on the first chunk of every stream and again on each progress update. `choices` is non-empty, both handler loops act only on `reasoning_content` / `content` / `tool_calls` / `finish_reason`, and the chunk left no record. `parse_chunk`'s own doc ("Returns `None` when the payload yields nothing to emit") was false for exactly this input.

Fixed, but not with the first of the two suggestions. Logging when a handler's `events` comes back empty would misreport ordinary traffic: `ReasoningExtractor::handle` (`stream/aggregator/reasoning.rs:53-65`) holds back up to `"<think>\n".len() - 1` bytes while no tag is in the buffer, so a short content chunk in `--reasoning-format none` mode legitimately drains nothing. That fix would file real content as dropped, and it would put the same log block back into both providers.

Took the second suggestion, narrowed to the fields these shared types declare, so the check stays a property of `StreamChoice` rather than a copy of handler logic:

```rust
impl StreamChoice {
    /// Whether this choice carries any field the providers turn into events.
    fn is_actionable(&self) -> bool {
        self.finish_reason.is_some()
            || self.delta.content.is_some()
            || self.delta.reasoning_content.is_some()
            || self.delta.tool_calls.is_some()
    }
}
```

```rust
if !chunk.choices.iter().any(StreamChoice::is_actionable) {
    debug!(provider, data, "Chunk carried no actionable choices.");
    return None;
}
```

Behaviour is unchanged — both handlers already produced zero events for such a chunk. The payload is in as `parse_chunk_drops_a_role_only_chunk`; I confirmed it fails against the unguarded version before landing the guard.

One case left alone: `{"delta":{"content":""}}` passes `is_actionable` and still emits nothing, since both handlers skip empty strings. Narrow enough that chasing it costs more than it returns.

PR #1088, opened 2026-09-04
fix(cli, printer): Close the chrome channel under --quiet

JeanMertz:
`-q` promised to suppress all output and delivered none of it: the only
thing consulting the flag was the tracing setup, so a quiet run still
printed the assistant's response, tool call headers, retry notices, and
status lines.

`-q` now silences chrome — the run's commentary on itself, everything on
stderr. Command output and the assistant's response keep stdout, because
that is the data the caller asked for, and a failing run still reports
its error. `>/dev/null` and `2>/dev/null` remain the way to drop those,
which is why the flag does not try to: the shell already expresses
"silence everything", and cannot express "chrome but not errors".

    jp -q query "summarize this" > answer.md    # no headers, no spinner

The policy lives on the `Printer`, the one place every emission passes,
rather than at each of the call sites that emit. `Chrome::Silenced`
discards `eprint`, `eprintln`, `erase_line`, and anything written
through `err_writer`, and leaves stdout and the prompt channel alone.
Chrome that repaints a line in place asks `chrome_repaints`, which
answers for a silenced channel and a JSON consumer at once, so the
waiting timer does not spawn a task to format a line nobody receives.

Scripts reading JP's stderr under `-q` saw output before and see none
now. `jp c grep -q` is unchanged: reporting a match through the exit
status alone is grep's own convention, and it keeps it.

### DIFF
--- a/crates/jp_cli/src/cmd/plugin/dispatch.rs
+++ b/crates/jp_cli/src/cmd/plugin/dispatch.rs
@@ -1309,6 +1309,7 @@ pub(crate) async fn resolve_plugin_binary(
     name: &str,
     plugins_config: &PluginsConfig,
     interactive: bool,
+    printer: &Printer,
 ) -> Result<Option<Utf8PathBuf>, cmd::Error> {
     let plugin_cfg = plugins_config.command.get(name);
 
@@ -1327,14 +1328,14 @@ pub(crate) async fn resolve_plugin_binary(
     }
 
     // 2. Check registry.
-    if let Some(path) = try_registry_install(name, plugins_config, interactive).await? {
+    if let Some(path) = try_registry_install(name, plugins_config, interactive, printer).await? {
         return Ok(Some(path));
     }
 
     // 3. Check $PATH with run policy.
     let segments: Vec<&str> = name.split('-').collect();
     if let Some(path) = find_plugin_binary(&segments) {
-        check_run_policy(name, &path, plugin_cfg, interactive)?;
+        check_run_policy(name, &path, plugin_cfg, interactive, printer)?;
         return Ok(Some(path));
     }
 
@@ -1369,6 +1370,7 @@ async fn try_registry_install(
     name: &str,
     plugins_config: &PluginsConfig,
     interactive: bool,
+    printer: &Printer,
 ) -> Result<Option<Utf8PathBuf>, cmd::Error> {
     let Some(reg) = registry::load_cached() else {
         return Ok(None);
@@ -1426,17 +1428,13 @@ async fn try_registry_install(
                 )));
             }
 
-            let mut writer = std::io::stderr();
-            drop(writeln!(
-                writer,
-                "  \u{2192} Plugin `{id}` found in registry."
-            ));
+            printer.eprintln(format!("  \u{2192} Plugin `{id}` found in registry."));
             let options = vec![
                 InlineOption::new('y', "install and run"),
                 InlineOption::new('n', "cancel"),
             ];
             let answer = InlineSelect::new("Install and run it?", options)
-                .prompt(&mut writer)
+                .prompt(&mut printer.prompt_writer())
                 .map_err(|e| cmd::Error::from(format!("prompt failed: {e}")))?;
 
             if answer != 'y' {
@@ -1446,18 +1444,12 @@ async fn try_registry_install(
         RunPolicy::Unattended => {}
     }
 
-    drop(writeln!(
-        std::io::stderr(),
-        "  \u{2192} Installing jp-{id} for {target}..."
-    ));
+    printer.eprintln(format!("  \u{2192} Installing jp-{id} for {target}..."));
     let client = reqwest::Client::new();
     let data = registry::download_and_verify(&client, binary_info).await?;
 
     let path = registry::install_binary(id, &data)?;
-    drop(writeln!(
-        std::io::stderr(),
-        "  \u{2192} Installed to {path}",
-    ));
+    printer.eprintln(format!("  \u{2192} Installed to {path}"));
 
     // Verify against pinned checksum if configured.
     verify_checksum(id, &path, plugin_cfg)?;
@@ -1471,6 +1463,7 @@ fn check_run_policy(
     binary_path: &Utf8Path,
     plugin_cfg: Option<&CommandPluginConfig>,
     interactive: bool,
+    printer: &Printer,
 ) -> Result<(), cmd::Error> {
     // Verify pinned checksum first.
     verify_checksum(name, binary_path, plugin_cfg)?;
@@ -1501,10 +1494,8 @@ fn check_run_policy(
                 return Ok(());
             }
 
-            let mut writer = std::io::stderr();
-            drop(writeln!(
-                writer,
-                "  \u{2192} Found jp-{name} on $PATH ({binary_path})",
+            printer.eprintln(format!(
+                "  \u{2192} Found jp-{name} on $PATH ({binary_path})"
             ));
             let options = vec![
                 InlineOption::new('y', "run this time"),
@@ -1512,7 +1503,7 @@ fn check_run_policy(
                 InlineOption::new('n', "deny"),
             ];
             let answer = InlineSelect::new("Run it?", options)
-                .prompt(&mut writer)
+                .prompt(&mut printer.prompt_writer())
                 .map_err(|e| cmd::Error::from(format!("prompt failed: {e}")))?;
 
             match answer {
@@ -1696,8 +1687,13 @@ pub(crate) async fn run_external(args: &[String], ctx: &mut Ctx) -> cmd::Output 
     }
 
     let config = ctx.config();
-    let Some(binary) =
-        resolve_plugin_binary(subcommand, &config.plugins, ctx.term.interactive).await?
+    let Some(binary) = resolve_plugin_binary(
+        subcommand,
+        &config.plugins,
+        ctx.term.interactive,
+        &ctx.printer,
+    )
+    .await?
     else {
         return Err(unknown_subcommand_error(subcommand));
     };
--- a/crates/jp_cli/src/cmd/plugin/dispatch.rs
+++ b/crates/jp_cli/src/cmd/plugin/dispatch.rs
@@ -1428,7 +1428,7 @@ async fn try_registry_install(
                 )));
             }
 
-            printer.eprintln(format!("  \u{2192} Plugin `{id}` found in registry."));
+            printer.prompt_println(format!("  \u{2192} Plugin `{id}` found in registry."));
             let options = vec![
                 InlineOption::new('y', "install and run"),
                 InlineOption::new('n', "cancel"),
@@ -1494,7 +1494,7 @@ fn check_run_policy(
                 return Ok(());
             }
 
-            printer.eprintln(format!(
+            printer.prompt_println(format!(
                 "  \u{2192} Found jp-{name} on $PATH ({binary_path})"
             ));
             let options = vec![
--- a/crates/jp_cli/src/lib.rs
+++ b/crates/jp_cli/src/lib.rs
@@ -50,7 +50,7 @@ use jp_config::{
         load_partials_with_inheritance, log_load_diagnostics,
     },
 };
-use jp_printer::{OutputFormat, OutputWidth, Printer};
+use jp_printer::{Chrome, OutputFormat, OutputWidth, Printer};
 use jp_storage::backend::{
     FsStorageBackend, NullLockBackend, NullPersistBackend, ReadOnlySessionBackend,
 };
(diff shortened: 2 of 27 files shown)


crates/jp_cli/src/lib.rs
```diff
@@ -498,9 +506,26 @@ fn detect_output_width(declared: Option<u16>) -> OutputWidth {
...
+/// data.
+fn build_printer(globals: &Globals, format: OutputFormat) -> Printer {
+    let chrome = if globals.quiet {
+        Chrome::Silenced
+    } else {
+        Chrome::Shown
+    };
+
+    Printer::terminal(format)
+        .with_chrome(chrome)
```
JeanMertz: nit: A plausible `jp -q plugin update` still prints `Refreshing plugin registry...` and its completion status because `cmd/plugin/update.rs` writes those lines directly to `std::io::stderr()` rather than through this `Printer`. The plugin install and auto-install paths contain similar direct status writes. This leaves `--quiet` observably incomplete. Please route user-facing status through the run's printer, while keeping actual questions on the prompt channel, and add a CLI-level regression test that exercises one of these paths under `-q`.

crates/jp_printer/src/printer.rs
```diff
@@ -160,6 +181,35 @@ impl Printer {
...
+    /// Callers use it to skip the work behind a repaint: a status line nobody
+    /// receives is a timer ticking for nothing.
+    #[must_use]
+    pub const fn chrome_repaints(&self) -> bool {
+        matches!(self.chrome, Chrome::Shown) && !self.format.is_json()
+    }
+
+    /// Whether a task aimed at `target` reaches it.
+    const fn accepts(&self, target: PrintTarget) -> bool {
+        !matches!((self.chrome, target), (Chrome::Silenced, PrintTarget::Err))
```
JeanMertz: blocking: A text `jp -q query` whose response reaches a configured `assistant.model.parameters.max_tokens` loses its truncation warning here. `TurnCoordinator::handle_streaming_event` sends `finish_notice(...)` through `eprintln`, then transitions to `Complete`; `turn_loop` returns `Ok(())`. The caller therefore gets a truncated answer (or a silently discarded incomplete tool call) with exit status 0 and no indication that generation stopped early. This is failure information, not progress chrome. Preserve abnormal-finish diagnostics independently of chrome, or propagate these outcomes to the existing top-level error path. Please add a regression test for a quiet turn ending with `FinishReason::MaxTokens`.

crates/jp_cli/src/cmd/plugin/dispatch.rs
```diff
@@ -1501,18 +1494,16 @@ fn check_run_policy(
                 return Ok(());
             }
 
-            let mut writer = std::io::stderr();
-            drop(writeln!(
-                writer,
-                "  \u{2192} Found jp-{name} on $PATH ({binary_path})",
+            printer.eprintln(format!(
+                "  \u{2192} Found jp-{name} on $PATH ({binary_path})"
             ));
```
JeanMertz: blocking: With `jp -q <plugin>` resolving an unapproved binary from `$PATH`, this discards the resolved path and leaves only `Run it? [y,Y,n,?]` on the prompt channel. An ordinary PATH change can select a different build of the same plugin, and `Y` persists approval for that undisclosed path and checksum. The binary identity is part of the authorization question, not optional status. Include it in the prompt or write the preamble through `prompt_writer()`. The same split in `plugin/install.rs` hides the third-party warning under `-q`; keep that warning with its confirmation too. Add quiet-mode prompt coverage that checks these details remain visible.

JeanMertz: Confirmed and fixed in `b13de84f`.

The split was exactly as you describe: the identity went out through `printer.eprintln` (chrome, closed by `-q`) while the question went out through `prompt_writer()` (which `-q` leaves alone), so `Y` reached `registry::save_approval(name, binary_path)` for a binary the run had declined to name.

Three lines had the same shape, not one — the `$PATH` identity here, `Plugin \`{id}\` found in registry.` at the auto-install prompt, and the third-party warning in `plugin/install.rs`. All three exist only to make the question that immediately follows answerable, so they now travel with it via a new `Printer::prompt_println`. It skips the JSON envelope for the same reason the other prompt writers do: a prompt is not data.

I kept them as separate lines rather than folding the identity into the prompt title. A `$PATH` binary path can be long, and `InlineSelect` renders its title inline with the options, so the wrapping would be worse than the disclosure gain.

Genuine status in those paths stays chrome: downloading, installing, and where the binary landed. Those report work whose result is observable elsewhere.

On coverage: a full test of `check_run_policy` under `-q` needs an injectable prompt backend, which `jp_inquire::InlineSelect` doesn't have at these call sites, and I'd rather not build that seam here. What I did add pins the property the fix depends on — `silenced_chrome_leaves_the_prompt_channel_alone` in `printer_tests.rs` asserts the identity line survives `Chrome::Silenced` while a chrome line next to it does not.

crates/jp_cli/src/lib.rs
JeanMertz: This describes the state before the plugin commit in this PR — it's addressed already, except for one clause worth pulling out.

`plugin/update.rs` now takes `let printer = &ctx.printer` and emits both lines through it, with the outdated-plugin reporting moved into `report_updates(printer, &outdated)`. `rg 'io::stderr\(\)' crates/jp_cli/src/cmd/plugin` returns nothing, so the install and auto-install paths you also name are covered. The regression test you asked for is `a_quiet_run_reports_nothing` in `update_tests.rs`, asserting an empty error buffer under `Chrome::Silenced` for both the outdated and up-to-date branches.

I split deciding-what's-outdated from reporting-it to get that test, because everything that prints on these paths sits behind either a network fetch or an interactive prompt. A test driving `Update::run` end-to-end would either hit the network or pass whether or not the suppression worked.

The clause that wasn't satisfied is "while keeping actual questions on the prompt channel". The questions were already there, but the lines telling you *what* you were being asked about were not — which is the blocking comment on `dispatch.rs`. Fixed in `b13de84f`.

crates/jp_printer/src/printer.rs
JeanMertz: The path is real and I traced it: `finish_notice` goes out through `printer.eprintln`, `-q` drops it in `Printer::send` before it is queued, the reason is then discarded at `transition_from_streaming(&mut self, stream, _reason: FinishReason)`, and `FinishReason` appears nowhere in `jp_conversation`. When the notice fires, the information is destroyed rather than merely hidden.

I'm still not changing anything here, for two reasons.

**It's commentary, by the definition the code uses.** `finish_notice` fires, the coordinator transitions to `Complete`, `turn_loop` returns `Ok(())`, the outcome is `RunOutcome::AsExpected`, exit 0. It never reaches `parse_error` or the error report in `run()`. An error message is what JP prints when it did not produce a result; this prints when it did. I had drafted a diagnostic channel exempt from `-q` and withdrew it: that would have been a third output category invented to avoid the classification, while leaving the line classified as chrome everywhere in the code. If the information is load-bearing enough to survive `-q`, then the run isn't succeeding, and the fix belongs to the outcome — at which point it flows through the error report, which `-q` already keeps, and no new channel exists.

**It fires far less often than the comment implies.** `chain_on_max_tokens` defaults to true on Anthropic, and `should_chain` (`provider/anthropic.rs:447-451`) swallows a `MaxTokens` finish to issue a continuation request instead of ending the turn. So on Anthropic with default config the plain truncation notice is close to unreachable. Chaining is Anthropic-only, so `openai`, `google`, `cerebras`, `llamacpp`, and `openrouter` do reach it, as does Anthropic once `max_tokens` is set explicitly — but that's a narrower input than "a configured `max_tokens`" suggests.

Rarity also argues against the channel rather than for it. A line shouted at a user who isn't watching for it is worth less than a durable record they can find afterwards, which is where the real defects are:

- **T-0eyhhgp** — a refusal discards the streamed answer, pops the flushed responses back out of the stream, and exits 0. A failed run reporting success, independent of `-q` or any channel. The question is whether a refusal should be an error.
- **T-0eyhkvs** — a truncated tool call is discarded after its name is harvested for the notice, so the prose that survives reads as if the assistant chose not to act, and replay shows the same. Proposal is to persist the finish reason and the dropped names, which removes the sole-witness property entirely and keeps `-q` free of exemptions.

Declining the second remedy in the comment as well: routing an abnormal finish to the top-level error path changes `jp query` from exit 0 to non-zero for a run that produced a usable partial answer, which every script around JP would feel. That belongs in T-0eyhhgp's decision, scoped to refusals, not in this PR.

crates/jp_cli/src/cmd/plugin/dispatch.rs
JeanMertz: Verified at `b13de84f`: the PATH identity, registry identity, and third-party warning use `prompt_println`, which selects the same target as the following `prompt_writer()` and bypasses chrome suppression. This addresses the blocker. The printer tests cover suppression and JSON wrapping; I have inspected them, not run them.

crates/jp_printer/src/printer.rs
JeanMertz: blocking: I still consider this unresolved at `b13de84f`. The chaining detail limits the default Anthropic case, but the input in this finding explicitly configures `max_tokens`; `anthropic.rs:192-195` disables chaining for that input. A budget-limited request that stops while assembling an edit remains a plausible trigger.

Changing exit codes has a compatibility cost, so I would not require that particular remedy here. Preserving the abnormal-finish notice while suppressing progress is sufficient. Returning `Ok(())` describes the current implementation; it does not establish that the caller can safely mistake an interrupted answer for a completed one.

The missing durable record predates this PR. Suppressing the existing text-mode notice under `-q` is the additional behavior under review. T-0eyhhgp and T-0eyhkvs document follow-up work but do not provide a replacement signal in this head. Keep the notice visible until that replacement exists, without changing ordinary query exit codes.

docs/ticket/0eyhhgp-a-model-refusal-discards-the-answer-and-reports-success.md
```diff
@@ -0,0 +1,54 @@
...
+
+`TurnCoordinator::handle_streaming_event`
+(`crates/jp_cli/src/cmd/query/turn/coordinator.rs:326-332`) clears the event
+builder and pops the already-flushed responses back out of the conversation
+stream, per the `Refused` contract that partial output must not be kept.
+It then transitions to `Complete`, `turn_loop` returns `Ok(())`, and the run
+ends as `RunOutcome::AsExpected`.
+
+So the run produces no answer, records no answer, and tells the caller it
+succeeded.
```
JeanMertz: nit: For a provider that streams partial text before `FinishReason::Refused`, stdout can still contain that text. `Event::Part` renders through `self.view` immediately, and the refusal branch clears the event builder and stored responses, not the view; `self.view.flush()` then renders any remaining buffered text. Please distinguish discarded conversation events from terminal output here and in the severity section. Empty stdout describes a refusal without visible content, not every refusal; another outcome is a partial answer on stdout that is absent from the saved conversation.

crates/jp_printer/src/printer.rs
```diff
@@ -478,6 +551,10 @@ impl Printer {
     /// it needs.
     fn send(&self, command: Command) {
         if let Command::Print(task) = &command {
+            if !self.accepts(task.target) {
```
JeanMertz: This is the line that decides the open question on this PR, so leaving a pointer here: **T-0ez6tx1**, `docs/ticket/0ez6tx1-quiet-hides-the-only-notice-that-a-turn-ended-abnormally.md`.

A turn that ends on a non-standard finish reason reports it through `printer.eprintln` at `coordinator.rs:351-352`, which this check drops when the chrome channel is silenced. Nothing else records the fact — `transition_from_streaming` takes the reason as `_reason` (`coordinator.rs:364-367`), `FinishReason` appears nowhere in `jp_conversation`, and the names of tool calls discarded mid-build come from buffers that are then dropped. So under `-q` a truncated or refused turn is silent and leaves no trace.

We're shipping this PR with that behaviour and revisiting the question separately rather than settling it here. The ticket records both positions in full, including the objection that exit 0 describes the implementation rather than justifying it, and pins the input that reaches the notice: an explicitly configured `max_tokens` sets `chains_remaining = 0` (`anthropic.rs:194-195`), so Anthropic's request chaining does not absorb the finish, and chaining is Anthropic-only in any case.

Two related tickets came out of the same review, both pre-dating this PR:

- **T-0eyhhgp** — a refusal erases the answer from the conversation while leaving whatever streamed on the terminal, and exits 0.
- **T-0eyhkvs** — a tool call truncated by `max_tokens` is discarded after its name is harvested for the notice, so the surviving prose reads as if the assistant chose not to act.

T-0eyhkvs is the cheapest resolution of T-0ez6tx1: persisting the finish reason gives the fact a durable home, at which point the notice stops being its only witness and `-q` needs no exception. The design home for a severity floor on chrome is [D15]'s graduated `-q` levels, not a new output channel.

[D15]: https://github.com/dcdpr/jp/blob/main/docs/rfd/drafts/D15-structured-logging-infrastructure.md

PR #1090, opened 2026-09-04
fix(config, llm): Accept MCP parameters with no declared type

JeanMertz:
An MCP tool whose input schema leaves a parameter's `type` out no longer disappears from the assistant's tool list. A property with no `type` keyword is valid JSON Schema meaning "any value", which is how a server declares a free-form parameter: a JSON-RPC passthrough, an arbitrary value stored under a key, a payload whose shape depends on a sibling argument. The parameter reaches the provider as the server wrote it, and a `type` in the tool's TOML still narrows it.

Only the absent keyword counts as open. A `type` that is present but empty or unrecognised (`type = []`, `type = "widget"`) is malformed and still fails, as does a `$ref` that could not be followed: what that node declares is unknown rather than unconstrained, and forwarding a dangling pointer trades one skipped tool for a rejected request.

An unconstrained parameter also stops argument coercion. Coercion repairs a string the schema cannot accept, and a parameter that accepts anything has nothing to repair, so `"3"` reaches the tool as the text the model sent rather than the number 3.

### DIFF
--- a/crates/jp_llm/src/tool.rs
+++ b/crates/jp_llm/src/tool.rs
@@ -1076,14 +1076,13 @@ fn coerce_object(arguments: &mut Map<String, Value>, node: &Node<'_>) {
 }
 
 fn coerce_value(value: &mut Value, node: &Node<'_>) {
-    // Coercion repairs a string the schema cannot accept. A parameter that
-    // declares no type accepts it as written, so parsing it would hand the
-    // tool a number or an object where the model sent text.
-    if let Value::String(raw) = value
-        && !node.is_unconstrained()
-        && !node.types().iter().any(|type_| type_ == "string")
+    // Coercion repairs an argument the schema cannot take as written. A
+    // parameter that permits the string has nothing to repair, so parsing it
+    // would hand the tool a number or an object where the model sent text.
+    if let Value::String(raw) = &*value
+        && !node.permits(value)
         && let Ok(parsed) = serde_json::from_str::<Value>(raw)
-        && node.accepts(&parsed)
+        && node.permits(&parsed)
     {
         *value = parsed;
     }
--- a/crates/jp_llm/src/tool.rs
+++ b/crates/jp_llm/src/tool.rs
@@ -1076,7 +1076,11 @@ fn coerce_object(arguments: &mut Map<String, Value>, node: &Node<'_>) {
 }
 
 fn coerce_value(value: &mut Value, node: &Node<'_>) {
+    // Coercion repairs a string the schema cannot accept. A parameter that
+    // declares no type accepts it as written, so parsing it would hand the
+    // tool a number or an object where the model sent text.
     if let Value::String(raw) = value
+        && !node.is_unconstrained()
         && !node.types().iter().any(|type_| type_ == "string")
         && let Ok(parsed) = serde_json::from_str::<Value>(raw)
         && node.accepts(&parsed)
--- a/crates/jp_llm/src/tool/json_schema.rs
+++ b/crates/jp_llm/src/tool/json_schema.rs
@@ -302,6 +302,41 @@ fn validate_types(path: &str, types: &[String]) -> Result<(), ToolError> {
     Ok(())
 }
 
+/// Whether any node in the document leaves the JSON type of its value open.
+///
+/// Walks properties and array items, reading through `$ref` the way [`Node`]
+/// does, and stops at a definition already on the path so a recursive schema
+/// terminates.
+#[must_use]
+pub fn has_unconstrained_node(schema: &Value) -> bool {
+    Node::root(schema)
+        .properties()
+        .iter()
+        .any(|(_, property)| is_open(property, &mut vec![]))
+}
+
+fn is_open(node: &Node<'_>, visiting: &mut Vec<String>) -> bool {
+    if let Some(origin) = node.origin() {
+        if visiting.iter().any(|seen| seen == origin) {
+            return false;
+        }
+        visiting.push(origin.to_owned());
+    }
+
+    let open = node.is_unconstrained()
+        || node.items().is_some_and(|items| is_open(&items, visiting))
+        || node
+            .properties()
+            .iter()
+            .any(|(_, property)| is_open(property, visiting));
+
+    if node.origin().is_some() {
+        visiting.pop();
+    }
+
+    open
+}
+
 /// Expand every same-document `$ref` and drop the definitions block.
 ///
 /// For providers that cannot follow references.
--- a/crates/jp_llm/src/tool/json_schema.rs
+++ b/crates/jp_llm/src/tool/json_schema.rs
@@ -413,13 +413,15 @@ impl<'a> Node<'a> {
 
     /// Whether this node leaves the JSON type of its value open.
     ///
-    /// A schema with no `type` keyword accepts any value, which is how a server
-    /// declares a free-form parameter.
-    /// A node still holding a `$ref` that could not be followed is not open:
-    /// what it declares is unknown.
+    /// A schema object with no `type` keyword accepts any value, which is how a
+    /// server declares a free-form parameter.
+    /// Anything else that reads as declaring no type is not open, because what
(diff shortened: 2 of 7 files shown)


crates/jp_llm/src/tool/json_schema.rs
```diff
@@ -141,7 +141,14 @@ fn validate_node_inner(
     visiting: &mut Vec<String>,
 ) -> Result<(), ToolError> {
     let types = node.types();
-    validate_types(path, &types)?;
+    // A schema with no `type` keyword accepts any value, and constrains
+    // nothing that could contradict its `items`, `properties`, `enum` or
+    // `default`. A `type` that is present but empty or unrecognised, and a
+    // `$ref` that could not be followed, remain malformed.
+    let unconstrained = node.is_unconstrained();
+    if !unconstrained {
```
JeanMertz: blocking: A plausible input is the motivating MCP schema, such as `"value": {"description": "Any JSON value."}`, used with the OpenAI provider. This lets the schema through, but `openai::convert_tools` always sends tools with `strict: true`, and OpenAI's strict schema subset rejects an untyped property with `schema must have a 'type' key`. The whole query then fails instead of only this tool being skipped. Please adapt this at the provider boundary, for example by disabling strict mode for a tool the strict subset cannot represent, and add a provider conversion test for this schema.

crates/jp_llm/src/tool/json_schema.rs
```diff
@@ -404,9 +411,26 @@ impl<'a> Node<'a> {
...
     }
 
+    /// Whether this node leaves the JSON type of its value open.
+    ///
+    /// A schema with no `type` keyword accepts any value, which is how a server
+    /// declares a free-form parameter.
+    /// A node still holding a `$ref` that could not be followed is not open:
+    /// what it declares is unknown.
+    #[must_use]
+    pub fn is_unconstrained(&self) -> bool {
```
JeanMertz: blocking: A plausible malformed MCP response is `"properties": {"value": null}` or `"value": "string"`, which can come from a server's schema-generation bug. `Value::get` returns `None` for every non-object value, so this classifies those invalid subschemas as unconstrained and `validate` forwards them to the provider, where they can reject the entire query. Restrict this case to an object schema with no `type` or `$ref`, and handle boolean schemas explicitly (`true` is open, while `false` accepts nothing), or reject boolean schemas as unsupported. Please add malformed-scalar regression cases as well.

crates/jp_llm/src/tool.rs
```diff
@@ -1076,7 +1076,11 @@ fn coerce_object(arguments: &mut Map<String, Value>, node: &Node<'_>) {
 }
 
 fn coerce_value(value: &mut Value, node: &Node<'_>) {
+    // Coercion repairs a string the schema cannot accept. A parameter that
+    // declares no type accepts it as written, so parsing it would hand the
+    // tool a number or an object where the model sent text.
     if let Value::String(raw) = value
+        && !node.is_unconstrained()
```
JeanMertz: nit: A plausible input is an MCP property declared as `{"enum": [3]}` with an argument emitted as the JSON string `"3"`. Omitting `type` leaves the JSON type open, but `enum` still rejects the string and accepts the parsed number. This unconditional skip sends a schema-invalid value to the MCP server, which will usually return a visible argument error. Please base this decision on all supported constraints, or at least coerce when the enum excludes the raw string and contains the parsed value, and cover that case in a test.

crates/jp_llm/src/tool/json_schema.rs
JeanMertz: Accepted, fixed in 94e2440e.

Confirmed the mechanism: `Node::child` calls `resolve`, which hands back the value untouched when there is no `$ref` to follow, so `self.node` can be `Null`, `String` or `Bool`. `Value::get` yields `None` for all of them, and the predicate returned `true`. On `main` those same nodes gave `types() == []` and were rejected, so this was a regression that turned one skipped tool into a schema the provider gets to reject.

`is_unconstrained` now requires an object:

```rust
pub fn is_unconstrained(&self) -> bool {
    self.node.is_object()
        && self.node.get("type").is_none()
        && self.node.get("$ref").is_none()
}
```

One correction to the framing: boolean schemas are not malformed, they are legal JSON Schema 2020-12, and `true` genuinely does mean "any value". I took the "reject as unsupported" branch of your suggestion rather than the "handle explicitly" one, because reading `true` as open would mean teaching `types`, `items`, `properties`, `enumeration` and `accepts_type` about a node that isn't a map — much larger than this PR, and I have no evidence of a server emitting one. Rejecting matches what `main` already did. If a real server turns up using the boolean form, that is worth its own change.

`a_property_that_is_not_a_schema_object_is_rejected` covers `null`, `"string"`, `true` and `false`, asserting the exact message at the config key.

crates/jp_llm/src/tool/json_schema.rs
JeanMertz: Accepted, fixed in 109bb96e.

Verified the whole path before changing it. `convert_tools` hardcodes `strict: true` with no config route around it, and inside `process_strict` an untyped property is inert: `is_object_type(map.get("type"))` is false so it gets no strict treatment, and `make_schema_nullable` returns early precisely because there is no `type` key. It would have gone out verbatim inside a `strict: true` function schema, listed in `required`.

What settles it is the severity rather than the API detail: RFD-era #998 established that a tool JP cannot describe is skipped rather than failing the query, and this PR was making a tool JP *can* describe fail the entire query on the most common provider. Strictly worse than the bug it fixes.

Took your suggestion. `convert_tools` now asks `json_schema::has_unconstrained_node` and drops to `strict: false` (with `parameters_with_strict_mode(_, false)`, which already existed) for that one tool. The walk reads through `$ref` the way `Node` does and closes recursive definitions the same way `validate` does. Three tests in `mod convert_tools`: a free-form property, a free-form `items` schema, and a fully typed schema that stays strict.

I have not reproduced the 400 against the live API, so the specific error string is yours, not mine — but the direction holds regardless of the exact wording.

Two things deliberately left out of scope: Anthropic threads its own `strict_tools` flag through `convert_tools`, and Gemini has its own subset that `transform_schema` passes an untyped node straight through. Neither is verified against this shape.

crates/jp_llm/src/tool.rs
JeanMertz: Taken, but with a different fix than either you proposed, and one correction first.

This is not something the skip introduced. On `main`, `{"enum": [3]}` with `"3"` already passes the string through untouched: `types()` is empty, so `accepts(3)` is false and no coercion happens. Same outcome before and after this PR, so "this unconditional skip sends a schema-invalid value" misplaces the cause.

What is true, and is the useful half of the comment, is that the guard forecloses the fix. Without it, the new `accepts` early-return would have coerced `"3"` → `3` correctly. And the deeper problem is that `is_unconstrained()` means "declares no type" while the coercion site was using it as a proxy for "accepts anything" — a node with an `enum` accepts only its members, so the proxy is wrong there.

Your fallback (special-case the enum) is the shape I did not take: it would have added a fourth condition to a chain already asking the same question three ways. Folding `enum` into `is_unconstrained` is not available either, since `validate_node_inner` reads the same predicate and a typeless `{"enum": [1, "two"]}` has to stay legal there.

So I took your primary suggestion instead. 90e1294c renames `accepts` to `accepts_type` (which is all it ever checked) and adds `Node::permits`, applying the type and the `enum` together. Coercion is now: the string is not permitted as written, and the parsed value is. That drops both the `is_unconstrained` guard and the `types().any("string")` test, and gets every case right — `{"description": …}` + `"3"` stays a string, `{"enum": [3]}` + `"3"` becomes `3`, `{"enum": ["3"]}` + `"3"` stays a string, and the five rows of `coerces_json_strings_to_declared_parameter_types` are unchanged.

One deliberate behaviour change worth flagging rather than leaving to be discovered: `{"type": "integer", "enum": [1, 2]}` with `"5"` used to coerce to `5` and now stays `"5"`. The enum forbids `5`, so there is no repair to make. Both forms fail downstream, so nothing gets worse; it is in the commit message.

PR #1091, opened 2026-09-04
feat(config, tool): Add `access.env` grants for local tools

JeanMertz:
Local tools can now declare which environment variables they may read, alongside the existing `access.fs` path grants. A rule names one variable exactly, or a name prefix when it ends in `*`:

```toml
[[conversation.tools.my_tool.access.env]]
name = "AWS_*"
read = true

[[conversation.tools.my_tool.access.env]]
name = "AWS_SECRET_ACCESS_KEY"
read = false
```

The most specific rule wins, where specificity is the byte size of the literal name without its `*` sentinel. At equal size an exact rule beats a prefix rule, and equal-specificity rules break toward the one declared last, matching how appended config layers resolve for `access.fs`. A name carrying a `*` anywhere but its final character is rejected when the policy is compiled, so a rule that cannot mean what it looks like fails loudly instead of silently never matching.

`AccessPolicy::matching_env_rule` returns the matching rule rather than a verdict, because "no rule mentions this variable" and "a rule denies it" are different facts and consuming tools want to treat them differently — one can reasonably prompt the user, the other cannot. Enforcement stays cooperative: the grants travel in the tool's `Context`, and the tool decides what to pass on.

Declaring only `env` rules leaves filesystem access untouched. The deny-all sentinel that `compile_policy` synthesizes when every `fs` rule compiles away is an `access.fs` construct, and a config that never mentioned `fs` does not acquire one.

Implements a phase of RFD 076.

### DIFF
--- a/crates/jp_cli/src/access/compile.rs
+++ b/crates/jp_cli/src/access/compile.rs
@@ -41,11 +41,12 @@ pub enum CompileError {
     )]
     ExternalInsideWorkspace(String),
 
-    /// An `env` rule name is empty, or contains a `*` somewhere other than the
-    /// final character.
+    /// An `env` rule name is empty, carries a `*` somewhere other than the
+    /// final character, or contains a character no environment variable name
+    /// can hold.
     #[error(
-        "access.env rule name '{0}' is invalid; a name must be non-empty and may only use '*' as \
-         its final character"
+        "access.env rule name '{0}' is invalid; a name must be non-empty, may only use '*' as its \
+         final character, and cannot contain '=' or a null byte"
     )]
     InvalidEnvName(String),
 }
@@ -198,16 +199,23 @@ pub fn compile_policy(
 
 /// Validate and convert `access.env` rules into [`EnvRule`]s.
 ///
+/// A name that cannot match any real variable is rejected rather than compiled
+/// into a rule that silently never fires: a denying rule that never fires
+/// leaves the variable to whatever broader rule matches instead.
+/// `=` and the null byte are the two characters no environment variable name
+/// can hold — the rest of the character set is left alone, because real names
+/// range from POSIX `SHOUTY_SNAKE` to Windows' `ProgramFiles(x86)`.
+///
 /// # Errors
 ///
-/// Returns [`CompileError::InvalidEnvName`] for an empty name, or one carrying
-/// a `*` anywhere but the final character.
+/// Returns [`CompileError::InvalidEnvName`] for an empty name, one carrying a
+/// `*` anywhere but the final character, or one containing `=` or a null byte.
 fn compile_env(config: &[EnvRuleConfig]) -> Result<Vec<EnvRule>, CompileError> {
     config
         .iter()
         .map(|rule| {
             let literal = rule.name.strip_suffix('*').unwrap_or(&rule.name);
-            if rule.name.is_empty() || literal.contains('*') {
+            if rule.name.is_empty() || literal.contains('*') || rule.name.contains(['=', '\0']) {
                 return Err(CompileError::InvalidEnvName(rule.name.clone()));
             }
             Ok(EnvRule {
--- a/crates/jp_cli/src/access/compile.rs
+++ b/crates/jp_cli/src/access/compile.rs
@@ -1,15 +1,18 @@
-//! Host-side compilation of `access.fs` config rules into a
+//! Host-side compilation of `access` config rules into a
 //! [`jp_tool::AccessPolicy`].
 //!
-//! Compilation canonicalizes each rule path, runs the approval lifecycle for
-//! `external` rules, and bakes the approved canonical target into the compiled
-//! [`FsRule`].
+//! Compilation canonicalizes each `fs` rule path, runs the approval lifecycle
+//! for `external` rules, and bakes the approved canonical target into the
+//! compiled [`FsRule`].
+//! `env` rules are validated for pattern syntax and carried through as-is.
 //! The cooperative checker and the OS sandbox both consume the resulting
 //! policy.
 
 use camino::Utf8Path;
-use jp_config::conversation::tool::access::{AccessConfig, FsRuleConfig};
-use jp_tool::{AccessPolicy, FsRule, canonicalize_workspace_target, lexical_workspace_relative};
+use jp_config::conversation::tool::access::{AccessConfig, EnvRuleConfig, FsRuleConfig};
+use jp_tool::{
+    AccessPolicy, EnvRule, FsRule, canonicalize_workspace_target, lexical_workspace_relative,
+};
 use tracing::warn;
 
 use crate::access::approvals::{ApprovalLookup, ApprovalStore};
@@ -37,6 +40,14 @@ pub enum CompileError {
          pointing at the external symlink instead"
     )]
     ExternalInsideWorkspace(String),
+
+    /// An `env` rule name is empty, or contains a `*` somewhere other than the
+    /// final character.
+    #[error(
+        "access.env rule name '{0}' is invalid; a name must be non-empty and may only use '*' as \
+         its final character"
+    )]
+    InvalidEnvName(String),
 }
 
 /// The outcome of compiling a set of `access.fs` rules.
@@ -164,6 +175,7 @@ pub fn compile_policy(
     root: &Utf8Path,
     approve: impl FnMut(&str, &Utf8Path) -> ApprovalDecision,
 ) -> Result<(AccessPolicy, Vec<String>), CompileError> {
+    let env = compile_env(&config.env)?;
     let compiled = compile_fs(config, root, approve)?;
     let mut rules = compiled.rules;
 
@@ -178,9 +190,32 @@ pub fn compile_policy(
 
     let policy = AccessPolicy {
         fs: rules,
+        env,
         ..AccessPolicy::default()
     };
     Ok((policy, compiled.warnings))
+}
+
+/// Validate and convert `access.env` rules into [`EnvRule`]s.
+///
+/// # Errors
+///
+/// Returns [`CompileError::InvalidEnvName`] for an empty name, or one carrying
+/// a `*` anywhere but the final character.
+fn compile_env(config: &[EnvRuleConfig]) -> Result<Vec<EnvRule>, CompileError> {
+    config
+        .iter()
+        .map(|rule| {
+            let literal = rule.name.strip_suffix('*').unwrap_or(&rule.name);
+            if rule.name.is_empty() || literal.contains('*') {
+                return Err(CompileError::InvalidEnvName(rule.name.clone()));
+            }
+            Ok(EnvRule {
+                name: rule.name.clone(),
+                read: rule.read(),
+            })
+        })
+        .collect()
 }
 
 fn build_rule(
--- a/crates/jp_cli/src/access/compile_tests.rs
+++ b/crates/jp_cli/src/access/compile_tests.rs
@@ -102,6 +102,48 @@ fn an_interior_star_in_an_env_name_is_rejected() {
     assert_eq!(error, CompileError::InvalidEnvName("AWS_*_KEY".to_owned()));
 }
 
+/// `KEY=VALUE` is how environment variables are usually written down, so a rule
+/// name carrying an `=` is a plausible mistake.
+/// It can never match a real variable, and a denying rule that never matches
+/// leaves the variable to whatever broader rule does.
+#[test]
+fn an_assignment_shaped_env_name_is_rejected() {
+    let dir = camino_tempfile::tempdir().unwrap();
+    let config = env_config(vec![env_rule("AWS_PROFILE=prod", false)]);
+
+    let error = compile_policy(&config, dir.path(), |_, _| unreachable!()).unwrap_err();
+
+    assert_eq!(
+        error,
+        CompileError::InvalidEnvName("AWS_PROFILE=prod".to_owned())
+    );
+}
+
+#[test]
+fn a_null_byte_in_an_env_name_is_rejected() {
+    let dir = camino_tempfile::tempdir().unwrap();
+    let config = env_config(vec![env_rule("AWS\0KEY", true)]);
+
+    let error = compile_policy(&config, dir.path(), |_, _| unreachable!()).unwrap_err();
+
+    assert_eq!(error, CompileError::InvalidEnvName("AWS\0KEY".to_owned()));
+}
+
+/// Windows ships names like `ProgramFiles(x86)`, so only the two characters a
+/// name genuinely cannot hold are rejected.
+#[test]
+fn an_unusual_but_legal_env_name_is_accepted() {
+    let dir = camino_tempfile::tempdir().unwrap();
+    let config = env_config(vec![env_rule("ProgramFiles(x86)", true)]);
+
+    let (policy, _) = compile_policy(&config, dir.path(), |_, _| unreachable!()).unwrap();
+
+    assert_eq!(policy.env, vec![EnvRule {
+        name: "ProgramFiles(x86)".to_owned(),
+        read: true,
+    }]);
+}
+
 #[test]
 fn an_empty_env_name_is_rejected() {
     let dir = camino_tempfile::tempdir().unwrap();
--- a/crates/jp_cli/src/access/compile_tests.rs
+++ b/crates/jp_cli/src/access/compile_tests.rs
@@ -1,7 +1,24 @@
 use super::*;
 
 fn fs_config(rules: Vec<FsRuleConfig>) -> AccessConfig {
-    AccessConfig { fs: rules }
+    AccessConfig {
+        fs: rules,
+        env: vec![],
+    }
+}
+
+fn env_config(rules: Vec<EnvRuleConfig>) -> AccessConfig {
+    AccessConfig {
+        fs: vec![],
+        env: rules,
+    }
+}
+
+fn env_rule(name: &str, read: bool) -> EnvRuleConfig {
+    EnvRuleConfig {
+        name: name.to_owned(),
+        read: Some(read),
+    }
 }
 
 fn rule(path: &str) -> FsRuleConfig {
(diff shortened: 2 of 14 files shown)


crates/jp_config/src/conversation/tool/access.rs
```diff
@@ -74,6 +101,11 @@ impl PartialConfigDelta for PartialAccessConfig {
                 .into_iter()
                 .filter(|v| !self.fs.contains(v))
                 .collect(),
+            env: next
+                .env
+                .into_iter()
+                .filter(|v| !self.env.contains(v))
+                .collect(),
```
JeanMertz: blocking: A continuing conversation that already grants `AWS_*` can hit this when a later `--cfg` layer replaces `access.env` with only a `CI` rule. `get_config_delta_from_cli` diffs the two resolved partials, but this converts the replacement wrapper into a plain append containing only novel entries. The persisted delta therefore adds `CI` without removing `AWS_*`, so the old credential grant silently returns on the next invocation. Preserve replacement semantics whenever rules disappear, and add a regression test that folds the delta over the previous config and verifies the removed env rule stays removed.

crates/jp_cli/src/access/compile.rs
```diff
@@ -178,11 +190,34 @@ pub fn compile_policy(
 
     let policy = AccessPolicy {
         fs: rules,
+        env,
```
JeanMertz: blocking: A local tool with a custom `style.parameters` formatter and an `access.env` deny rule bypasses this policy during formatting. `format_args_custom` builds the `Action::FormatArguments` context without `access` (`crates/jp_cli/src/render/tool.rs:823-833`), so it deserializes as `None` (unrestricted) while the formatter subprocess still inherits JP's environment. A cooperating formatter can therefore read a denied variable before the `Run` action receives this compiled policy. Compile and pass the same policy into the formatter path, with a test covering both context actions, as RFD 076 Phase 3 specifies.

crates/jp_config/src/conversation/tool/access.rs
```diff
@@ -52,13 +66,26 @@ pub struct AccessConfig {
         merge = vec_with_strategy,
     )]
     pub fs: Vec<FsRuleConfig>,
+
+    /// Environment variable access rules.
+    ///
+    /// Each rule names one variable, or a name prefix when `name` ends in `*`.
+    /// Rules from later config layers append by default; the most specific
+    /// (longest literal name) rule wins for a given variable.
```
JeanMertz: nit: A user who adds only `AWS_*` can reasonably read this as an additive grant, but a non-empty `access.env` list makes every unmatched variable default-deny. Because these comments become generated TOML and schema documentation, please state that absent or empty `access.env` is unrestricted, any rule switches it to default-deny, and omitting `read` denies the matched name.

crates/jp_tool/src/access.rs
```diff
@@ -780,6 +830,91 @@ mod tests {
...
+                .matching_env_rule("AWS_SECRET_ACCESS_KEY")
+                .unwrap()
+                .read
+        );
+    }
+
+    /// A shorter but more specific rule must not lose to a longer prefix rule
+    /// declared after it.
+    #[test]
+    fn a_longer_prefix_beats_a_shorter_exact_rule() {
```
JeanMertz: nit: This test describes the opposite of its data. `AWS_SECRET_ACCESS_KEY` is the 21-byte exact rule and `AWS_*` has a 4-byte literal prefix, so the body proves that a longer exact rule beats a shorter prefix even when the prefix is declared last. Please rename the test and comment accordingly; the current wording makes the specificity contract harder to verify.

crates/jp_cli/src/access/compile_tests.rs
```diff
@@ -17,6 +34,84 @@ fn rule(path: &str) -> FsRuleConfig {
...
+/// for a config that never mentioned `fs`.
+#[test]
+fn env_only_config_leaves_fs_unrestricted() {
+    let dir = camino_tempfile::tempdir().unwrap();
+    let config = env_config(vec![env_rule("GITHUB_TOKEN", true)]);
+
+    let (policy, _) = compile_policy(&config, dir.path(), |_, _| unreachable!()).unwrap();
+
+    assert!(policy.fs.is_empty());
+    assert!(!policy.is_restricted());
```
JeanMertz: theoretical: I could not find an in-tree caller that treats `is_restricted()` as policy-wide, so I cannot name a failing operation today. This assertion nevertheless codifies that an `AccessPolicy` carrying env restrictions reports itself as unrestricted. Now that the policy has multiple active resource axes, the generic name is hazardous for future sandbox gating. Rename it to `is_fs_restricted()` (and assert that here), or add explicit per-resource predicates before consumers depend on it.

crates/jp_tool/src/access.rs
```diff
@@ -369,6 +395,30 @@ pub struct EnvRule {
...
+        !self.name.ends_with('*')
+    }
+
+    /// Whether the rule applies to the variable `name`.
+    #[must_use]
+    pub fn matches(&self, name: &str) -> bool {
+        if self.is_exact() {
+            name == self.name
+        } else {
+            name.starts_with(self.literal())
```
JeanMertz: theoretical: There is no in-tree env consumer yet, so I cannot name a JP operation that fires this today. On Windows, environment-variable names are case-insensitive, but both branches here are case-sensitive. A policy that allows `*` and explicitly denies `GITHUB_TOKEN` would select the wildcard allow for a request named `github_token`, even though Windows resolves that to the denied variable. Define matching in terms of the target platform's env-name identity and add a Windows case-variant test before a consumer relies on this matcher.

crates/jp_cli/src/access/compile.rs
JeanMertz: Correct on the facts, and the RFD citation is exact — Phase 3 does say "`execute_local()` **and** the `FormatArguments` path", and only the first half shipped. Filed as **T-0dftakj** rather than fixed here.

Two reasons it isn't this PR's:

This PR is Phase 2 (config types in `jp_config`). Phase 3 is a separate phase that the RFD lists as depending on it, and it touches `jp_llm`/`jp_cli` render code this branch doesn't go near. The gap is also not new — `access.fs` has had it since `ab94ec0a` (#727) in June, so nothing here widened it.

I'd also push back on "bypasses this policy". `run_tool_command` never clears the environment (`crates/jp_llm/src/tool.rs:461`), for either action — so a local tool binary on the **`Run`** path can read a denied variable with `std::env::var` just as easily, compiled policy in hand. Enforcement is cooperative throughout; the policy in `Context` is a contract a tool opts into, not a sandbox. The real defect is narrower and still worth fixing: a formatter author who *wants* to honour the grants can't read them, so a preview can't say "this variable is denied". That's what the ticket describes, along with compiling the policy once for both context constructions rather than twice.

crates/jp_cli/src/access/compile_tests.rs
JeanMertz: Agreed on the hazard, deferred on the timing — filed as **T-0dftcc5**.

Confirming your read: `is_restricted` has one production caller, `AccessPolicy::permits` (`crates/jp_tool/src/access.rs:112`), which is fs-only, plus two test sites. The doc comment already says "Whether **filesystem** access is restricted". So there's nothing to correct today.

`jp_tool` is the wire crate external tool binaries build against, so the rename is a public API break. Doing it now buys a better name and no behaviour change; doing it alongside the second predicate (RFD 076's `net`/`env` consumers, or RFD 075's sandbox gating) makes the pair obvious and costs one break instead of two. I'd rather spend the break once, when there's a sibling to name it against.

On your alternative — adding per-resource predicates up front — I'd skip that specifically. A predicate with no caller is a guess at what the caller will want, and `env` has already shown it wants three states rather than a boolean (see `matching_env_rule` returning the rule, not a verdict).

The assertion this thread is anchored to stays as is: the test's claim is that no fs deny-all sentinel gets synthesized for an env-only config, and `is_restricted()` is exactly the fs predicate for that.

crates/jp_config/src/conversation/tool/access.rs
JeanMertz: Taking the third clause, declining the first two.

The `read` default was genuinely missing from where a user reads it — the fact lived on the struct doc, while the field a user actually sees in generated TOML said only "Grant reading the variable's value." Fixed:

```rust
/// Grant reading the variable's value.
///
/// Defaults to `false`: a rule that omits `read` denies the names it
/// matches.
pub read: Option<bool>,
```

"Absent or empty `access.env` is unrestricted, any rule switches it to default-deny" I've left out, because it describes `access.fs`, not this field. Nothing in this PR defines what an unmatched variable means — `matching_env_rule` returns `Option<&EnvRule>` precisely so the consumer decides, and its doc says so. RFD 076's worked example goes the other way from your wording: "`HOME` → no match → **denied**."

Concretely, the consumer I'm writing against this treats an unmentioned variable as *ask the user*, which is neither "unrestricted" nor "default-deny". Writing either into the schema now would be documenting a guess, and it'd be a guess users configure against. That sentence belongs in the field doc once Phase 3+ lands a consumer and the answer is real.

crates/jp_config/src/conversation/tool/access.rs
JeanMertz: Mechanism confirmed end to end — I traced every link:

- `PartialAccessConfig.env` is `MergeableVec<PartialEnvRuleConfig>` (`types/vec.rs:25-30`).
- `ToPartial` deliberately emits `Merged { strategy: Some(Replace) }` (`types/vec.rs:140-162`).
- `MergeableVec::into_iter` calls `into_vec()`, dropping the wrapper; `FromIterator` rebuilds a bare `Self::Vec` (`types/vec.rs:179-193`).
- `vec_with_strategy` reads a bare `Vec` as `None` strategy → the `None | Some(Append)` arm (`internal/merge/vec.rs:45, 64-68`).
- Live path: `get_config_delta_from_cli` → `PartialAppConfig::delta` → `delta_opt_partial(self.access…)` → here.

So the narrowing delta appends and the dropped grant returns. You're right.

Not fixing it here, though, because **RFD 070 already owns it** and says so explicitly (lines 705-716):

> the wrapper's `strategy` is dropped and a key present in the previous state but absent from the next produces no entry rather than a removal… a rule the replacement dropped comes back. `attachments` (a `MergeableVec`) and `conversation.tools` (a plain `IndexMap`) behave the same way. This RFD's `unsets` mechanism is what closes it; **until then the gap is uniform across every container-backed field.**

`access.fs`, two fields up, has byte-identical code from `ab94ec0a`. Fixing `env` alone leaves two sibling fields in one struct with different removal semantics — worse for anyone reasoning about `access`, and undone when `unsets` lands. That uniformity is what makes the gap tractable to sweep in one pass.

What I have done is make `access` findable when that sweep happens: RFD 070's list of affected fields now names `access.fs` and `access.env` with the specific `PartialAccessConfig::delta` mechanism, so it can't be missed. RFD 070 is Accepted and sits at priority 10.

crates/jp_tool/src/access.rs
JeanMertz: The premise holds further than you claimed — Windows isn't hypothetical here, `.github/workflows/rust.yml:102` runs the suite on `windows-latest`. And RFD 076 specifies case-insensitivity for **net** host matching (line 368) while saying nothing about case for env, so the spec has a real hole, not just the code.

I've closed the hole by deciding the other way, and written the decision down in both places.

Matching stays case-sensitive on every platform. Platform-dependent identity would mean a `.jp/config.toml` checked into a repo denies `GITHUB_TOKEN` for a Linux teammate and, for the `github_token` spelling, grants it for a Windows one. A security-relevant rule whose meaning depends on who ran it is worse than one that is uniformly strict — and strict is the safe direction when they disagree.

`EnvRule::matches` now says so, and RFD 076's env section gained a paragraph stating it and contrasting it with net host matching. The paragraph also names where the Windows concern does get handled: a consumer that needs to accommodate Windows callers normalizes the requested name before matching, which keeps the rule set's meaning fixed. That's a Phase 3 question and belongs with the consumer — a Windows case-variant test written now would pin behaviour for a caller that doesn't exist and whose normalization choice would immediately change it.

crates/jp_tool/src/access.rs
JeanMertz: Right, and not really a nit — this test's entire job is documenting the specificity contract, and both its name and its comment stated the reverse of what it proves. `AWS_SECRET_ACCESS_KEY` is 21 bytes exact, `AWS_*` has a 4-byte literal, and the longer exact rule wins despite the prefix being declared last. Nothing here is a "longer prefix".

Fixed:

```rust
/// Specificity decides, not declaration order: the 21-byte exact rule wins
/// over the 4-byte `AWS_` prefix even though the prefix is declared last.
#[test]
fn a_longer_exact_rule_beats_a_shorter_prefix_declared_after_it() {
```

The assertion is unchanged — the data was always right, only the label was wrong.

crates/jp_cli/src/access/compile.rs
JeanMertz: nit: The phase boundary and the pre-existing `access.fs` gap change the triage. This PR adds the Phase 2 config and compilation surface; Phase 3 formatter plumbing is separate, and process isolation is not provided on either action. T-0dftakj records the actual defect and the shared-policy regression test it needs. Treating this thread as non-blocking for this PR.

crates/jp_config/src/conversation/tool/access.rs
JeanMertz: blocking: Recording the limitation in RFD 070 does not change the runtime severity. `strategy = "replace"` is documented as supported for `access.env`, and the ordinary continuing-conversation path silently persists the opposite policy after a grant is removed. The fact that other container fields share the defect does not make credential-grant resurrection safe. Preserve replacement semantics for both `access.fs` and `access.env` when rules disappear, or defer this resource axis until RFD 070 lands.

crates/jp_config/src/conversation/tool/access.rs
JeanMertz: nit: The added `read` documentation resolves that part. The resource default still belongs here: RFD 076 lines 90-97 defines non-empty `access.env` as default-deny, and its worked examples mark every no-match case denied. A consumer may escalate an unmentioned name to the user, but that is not unrestricted access. The generated field documentation should say that absent or empty is unrestricted, while a non-empty list does not grant unmatched names.

crates/jp_tool/src/access.rs
JeanMertz: theoretical: The added rationale does not make the policy uniformly strict. With `*` granting reads and exact `GITHUB_TOKEN` denying them, a Windows request for `github_token` misses the exact deny, selects the wildcard allow, and then resolves to the same OS variable. On Unix that lookup can be absent, so this still creates the platform-dependent grant the new text says it avoids. Normalizing only the request is also insufficient unless configured rules use the same canonical form. This stays non-blocking because there is no in-tree env consumer, but the Windows identity rule needs a deny-preserving definition before one lands.

crates/jp_cli/src/access/compile.rs
```diff
@@ -178,11 +190,34 @@ pub fn compile_policy(
...
+/// # Errors
+///
+/// Returns [`CompileError::InvalidEnvName`] for an empty name, or one carrying
+/// a `*` anywhere but the final character.
+fn compile_env(config: &[EnvRuleConfig]) -> Result<Vec<EnvRule>, CompileError> {
+    config
+        .iter()
+        .map(|rule| {
+            let literal = rule.name.strip_suffix('*').unwrap_or(&rule.name);
+            if rule.name.is_empty() || literal.contains('*') {
```
JeanMertz: nit: A user who pastes `name = "AWS_PROFILE=prod"` can plausibly hit this. Environment-variable keys cannot contain `=` or NUL, but this validator accepts both, so the policy compiles with a rule that cannot match a real variable. Reject those characters here as invalid names and add a regression case for `=`; it is a common assignment-shaped typo.

JeanMertz: Taking this. `AWS_PROFILE=prod` is exactly the shape of mistake `KEY=VALUE` habits produce, and the consequence is worse than a dead rule: if the malformed one was the *deny*, the variable falls through to whatever broader rule matches, so a rule written to close something silently leaves it open.

`compile_env` now rejects `=` and the null byte alongside the existing checks, and `InvalidEnvName`'s message names them.

I stopped at those two rather than enforcing the POSIX charset. They are the only characters a name genuinely cannot hold — `=` by POSIX, NUL because it terminates the string the OS hands back — whereas tightening to `[A-Za-z_][A-Za-z0-9_]*` would reject real names like Windows' `ProgramFiles(x86)`. Three regression cases: `=`, NUL, and `ProgramFiles(x86)` compiling successfully so the permissive half is pinned too.

crates/jp_config/src/conversation/tool/access.rs
JeanMertz: You're right and I was wrong on the citation. Lines 90-97 say it plainly:

> Default-deny applies **per resource type**. Declaring at least one rule in `access.fs` shifts filesystem access to default-deny; `access.net` and `access.env` remain unrestricted until they, too, have at least one rule… absent or empty lists mean "unrestricted" for that type.

The line I quoted back at you last time — "`HOME` → no match → **denied**" — sits inside a policy that *has* env rules, so it is the default-deny case. It supports your reading rather than contradicting it. I read a worked example as a general statement.

The rest of my argument dissolves once the distinction is drawn properly: `matching_env_rule` returning `Option` lets a consumer choose *how* to refuse an unmentioned variable (prompt, or refuse outright), not *whether* it is granted. Both are non-grants, so the Option return and RFD-defined default-deny were never in tension. I conflated the two.

Added to the field doc, above the matching rules so it lands first in generated TOML:

```rust
/// When absent or empty, the tool may read any environment variable.
/// Declaring at least one rule switches to default-deny: a variable is
/// readable only when a rule matching it grants `read`.
```

Worth noting `access.fs` has the same omission at field level — its default-deny sentence lives on the module doc and on `ToolConfig::access`, not on `fs` itself. Left alone here, but the same edit would be right.

crates/jp_config/src/conversation/tool/access.rs
JeanMertz: Fixed, for both fields. You changed my mind on two counts.

My load-bearing objection was divergence — that fixing `env` alone would leave siblings in one struct with different removal semantics. Widening the ask to both answers it directly, and I should have got there myself instead of treating "uniform" as a reason to stay uniformly broken.

The severity argument I also under-weighted. The input is ordinary (`--cfg` narrowing on a continuing conversation) and the damage *hides*: the user believes a credential grant is gone, the fold re-grants it, and nothing surfaces that. Hidden damage from a plausible input is the case that gates a merge, and "RFD 070 will redo this" is not an answer when the local fix is fifteen lines.

There is also in-tree precedent I'd missed. `conversation.labels` was already fixed this exact way — `PartialConversationConfig::delta` detects a dropped key and emits the whole wrapper with `replace` — so the codebase already had fixed and unfixed container fields side by side. My "uniform gap" premise was false before this PR.

`PartialAccessConfig::delta` now routes both fields through:

```rust
fn rule_delta<T: Clone + PartialEq>(
    prev: &MergeableVec<T>,
    next: MergeableVec<T>,
) -> MergeableVec<T> {
    if prev.iter().all(|rule| next.contains(rule)) {
        return next.into_iter().filter(|rule| !prev.contains(rule)).collect();
    }

    MergeableVec::Merged(MergedVec {
        value: next.into_vec(),
        strategy: Some(MergedVecStrategy::Replace),
        dedup: None,
        discard_when_merged: false,
    })
}
```

Five tests, built from `AccessConfig::to_partial()` so they run on the production shape rather than a file-loaded one: removal-survives-the-fold for `env` and for `fs`, additive-still-appends for both, and one asserting the axes diff independently so narrowing `env` doesn't drag `fs` into replace semantics.

I checked the removal tests fail without the fix rather than trusting the reasoning — reverting `rule_delta` to the old body gives `["AWS_*", "CI"]` where the config says `["CI"]`, which is the resurrection in one line of output.

RFD 070's paragraph is updated: it no longer lists `access` among the broken fields, and it now records both point fixes as things `unsets` should delete rather than patterns to copy.

crates/jp_tool/src/access.rs
JeanMertz: Your counterexample holds, and it falsifies the rationale I wrote. Tracing it: `"github_token" == "GITHUB_TOKEN"` is false so the exact deny misses; `"github_token".starts_with("")` is true so the `*` grant matches and wins uncontested; the consumer then calls `std::env::var("github_token")`, which resolves to the denied variable on Windows and usually to nothing on Unix. Same config, same request, different outcome per platform — precisely what the paragraph claimed to avoid. Your second point lands too: canonicalizing only the request leaves an exact rule written in another case unmatched.

The behaviour is unchanged, but the text justifying it was wrong, and a false justification is worse than none because the next reader trusts it. Both places now state the limitation instead of claiming a property the code doesn't have — `EnvRule::matches` carries your `GITHUB_TOKEN` / `github_token` example directly, and RFD 076's paragraph keeps the decision (case-sensitive, so a rule set means one thing wherever it is read) while naming what that does not buy.

RFD 076 also gained an explicit open question, and **T-0e274mn** tracks settling it. The two candidates recorded there are canonicalizing requests *and* rule names, or matching denies case-insensitively while keeping grants exact so the strict direction always wins. I've deliberately not picked one: with no consumer, either choice is a guess, and the second in particular wants a caller to reason against. Agreed it needs answering before the first `env` consumer ships, and the case-variant test belongs with that decision.

crates/jp_config/src/conversation/tool/access.rs
```diff
@@ -69,19 +100,46 @@ impl AssignKeyValue for PartialAccessConfig {
...
+/// and replacing with it loses nothing.
+fn rule_delta<T: Clone + PartialEq>(
+    prev: &MergeableVec<T>,
+    next: MergeableVec<T>,
+) -> MergeableVec<T> {
+    if prev.iter().all(|rule| next.contains(rule)) {
+        return next
+            .into_iter()
+            .filter(|rule| !prev.contains(rule))
+            .collect();
```
JeanMertz: blocking: A continuing conversation can still turn a deny into a grant when the resolved order changes without removing a rule. For example, let `prev` contain `GITHUB_TOKEN` with `read = false`, then apply a `prepend` layer adding the same name with `read = true`. The effective `next` is `[allow, deny]`, so last-wins correctly denies. This branch emits only the new allow as a bare append; folding it over `prev` produces `[deny, allow]`, silently granting the token on the next invocation. Rule order is policy data. Only emit an append delta when `prev` is an ordered prefix of `next`; otherwise emit the full replacement, and add a fold regression with equal-name allow/deny rules.

docs/rfd/070-negative-config-deltas.md
```diff
@@ -712,8 +712,19 @@ records a delta that deep-merges on the next fold, and a rule the replacement
...
 behave the same way.
-This RFD's `unsets` mechanism is what closes it; until then the gap is uniform
-across every container-backed field.
+
+Two fields have since been fixed in place, each by detecting the removal and
+emitting the whole container with `strategy = "replace"` instead of a minimal
+append: `conversation.labels`, and `access.fs` / `access.env` on a tool (via
+`rule_delta` in `jp_config::conversation::tool::access`).
+Both carry security-relevant grants, which is why they were worth closing ahead
+of the general mechanism.
```
JeanMertz: nit: `conversation.labels` does not carry security-relevant grants, so “Both” gives the first point fix the wrong rationale. Please say that `access.fs` and `access.env` were worth fixing early because they carry grants; `conversation.labels` is the separate point fix that exposed the forward-delta problem.

PR #1095, opened 2026-09-04
feat(cli): Search conversation labels in `c grep` and `c use`

JeanMertz:
Labels are now part of the text both search commands read. `jp c grep jp_config` finds a conversation labelled `crate=jp_config` even when nothing in its transcript mentions the crate, and `jp c use --grep jp_config` narrows the picker the same way.

A label is searched one key-value pair per line, written the way it is written on the command line: `crate=jp_config`, or the bare key when it holds no value. A pattern can therefore match a key, a value, or the whole pair. Hits are conversation-scoped, so they carry `..` as their turn field, the same coordinate a title hit gets.

The new `--scope label` restricts a search to labels, and is included in the default `all`. It sits alongside `--label`, which does something different: `--label` chooses *which* conversations are searched, while `--scope label` searches the label text itself. Neither `title` nor `label` reads the event stream, so `jp c grep --scope title,label` stays fast across a large workspace.

### DIFF
--- a/crates/jp_cli/src/cmd/conversation/grep.rs
+++ b/crates/jp_cli/src/cmd/conversation/grep.rs
@@ -312,8 +312,7 @@ impl Grep {
         wanted: &HashSet<ConcreteScope>,
         ctx: &Ctx,
     ) -> Vec<ConversationHits> {
-        // Any scope other than `Title` is sourced from the event stream.
-        // Skipping the event pass entirely when it can't contribute avoids a
+        // Skipping the event pass when no wanted scope can contribute avoids a
         // sequential disk read per conversation.
         let needs_events = needs_events_for(wanted);
 
--- a/crates/jp_cli/src/cmd/conversation/grep.rs
+++ b/crates/jp_cli/src/cmd/conversation/grep.rs
@@ -23,7 +23,8 @@ use crate::{
     ctx::Ctx,
     output::print_json,
     shared::search::{
-        ConcreteScope, Matcher, event_lines, event_scope, resolve_ignore_case, title_for,
+        ConcreteScope, Matcher, event_lines, event_scope, label_lines, resolve_ignore_case,
+        title_for,
     },
 };
 
@@ -382,6 +383,21 @@ impl Grep {
             );
         }
 
+        if wanted.contains(&ConcreteScope::Label) {
+            let labels = label_lines(ctx, &handle);
+            let lines: Vec<&str> = labels.iter().map(String::as_str).collect();
+            collect_scope_hits(
+                &mut group.hits,
+                None,
+                ConcreteScope::Label,
+                None,
+                &lines,
+                matcher,
+                self.context,
+                &mut budget,
+            );
+        }
+
         if !needs_events {
             return group;
         }
@@ -662,6 +678,9 @@ enum Scope {
     /// The conversation title.
     Title,
 
+    /// Conversation labels, each read as `key=value` or as a bare key.
+    Label,
+
     /// User chat requests.
     User,
 
@@ -684,10 +703,15 @@ enum Scope {
     Inquiry,
 }
 
-/// Whether the wanted scope set contains anything sourced from the event stream
-/// (i.e. something beyond `Title`).
+/// Whether the wanted scope set contains anything sourced from the event
+/// stream.
+///
+/// `Title` and `Label` come from the conversation's metadata, so a search
+/// restricted to them never touches the stream.
 fn needs_events_for(wanted: &HashSet<ConcreteScope>) -> bool {
-    wanted.iter().any(|s| *s != ConcreteScope::Title)
+    wanted
+        .iter()
+        .any(|s| !matches!(s, ConcreteScope::Title | ConcreteScope::Label))
 }
 
 /// Expand a user-facing list of scopes to the concrete set the search uses.
@@ -714,6 +738,7 @@ fn expand_scopes(scopes: &[Scope]) -> HashSet<ConcreteScope> {
                 out.extend([ConcreteScope::ToolCall, ConcreteScope::ToolResult]);
             }
             Scope::Title => _ = out.insert(ConcreteScope::Title),
+            Scope::Label => _ = out.insert(ConcreteScope::Label),
             Scope::User => _ = out.insert(ConcreteScope::User),
             Scope::Assistant => _ = out.insert(ConcreteScope::Assistant),
             Scope::Reasoning => _ = out.insert(ConcreteScope::Reasoning),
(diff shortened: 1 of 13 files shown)


crates/jp_cli/src/shared/search.rs
```diff
@@ -141,6 +144,24 @@ pub(crate) fn title_for(ctx: &Ctx, handle: &ConversationHandle) -> Option<String
...
+/// Keys are sorted, and a key's values keep the order they were added in.
+pub(crate) fn label_lines(ctx: &Ctx, handle: &ConversationHandle) -> Vec<String> {
+    let Ok(metadata) = ctx.workspace.metadata(handle) else {
+        return vec![];
+    };
+
+    metadata
+        .labels
+        .iter()
+        .flat_map(|(key, values)| values.iter().map(|value| label_text(key, value)))
```
JeanMertz: blocking: A user can set a literal label value containing a newline, for example `note = "first\nsecond"`; label values explicitly have no character restriction. `label_text` preserves that newline, so this vector contains one logical entry with two physical lines. Searching `second` then emits `ID:..:label:m:note=first\nsecond`, and the second output line has no coordinate even though line mode promises every line is a record. That silently breaks field-splitting and `fzf` consumers. Please split the rendered pair into physical lines before matching and collecting, as `event_lines` does, or define an escaped single-line representation, and add a multiline-label regression test.

crates/jp_cli/src/shared/search.rs
```diff
@@ -367,6 +389,13 @@ fn id_matches(ctx: &Ctx, id: ConversationId, matcher: &Matcher) -> bool {
         return true;
     }
 
+    if label_lines(ctx, &handle)
```
JeanMertz: blocking: An archived conversation carrying `crate=jp_config` is never found by `jp c use ?a --grep jp_config`, despite `Use` advertising `?a` as composable with `--grep`. `source_ids` returns the archived ID, but `id_matches` returns at `acquire_conversation` because archived conversations are deliberately absent from the live workspace index, so this label branch is never reached and the command reports that nothing matched. That is a hidden false negative. Please either make this search path read archived metadata and events, or explicitly reject archived targets rather than claiming the filter found no conversations.

crates/jp_cli/src/cmd/conversation/grep.rs
```diff
@@ -382,6 +383,21 @@ impl Grep {
             );
         }
 
+        if wanted.contains(&ConcreteScope::Label) {
+            let labels = label_lines(ctx, &handle);
```
JeanMertz: nit: The comment above `needs_events` still says every scope other than `Title` comes from the event stream. This `Label` branch makes that statement false and it contradicts `needs_events_for` below. Please update it to name both metadata scopes.

crates/jp_cli/src/shared/search.rs
JeanMertz: The fix is to make labels single line, it was never the intention to allow multi-line labels, let's fix that in this PR as well, even though it's not exactly in-scope. Update the RFD as well to add that restriction. Any newlines in a label value are converted to a space.

crates/jp_cli/src/cmd/conversation/grep.rs
JeanMertz: Agreed, that comment was already wrong the moment `Label` landed.

Rather than extend the enumeration to `Title | Label`, I dropped it. `needs_events_for`'s doc comment names both metadata scopes, so keeping a second copy here is what went stale in the first place and would go stale again the next time a metadata scope is added. What's left is the part only this site can say — why the flag exists:

```rust
// Skipping the event pass when no wanted scope can contribute avoids a
// sequential disk read per conversation.
let needs_events = needs_events_for(wanted);
```

crates/jp_cli/src/shared/search.rs
JeanMertz: Confirmed and fixed, taking the follow-up's approach rather than the one in the comment body.

I reproduced it first by stubbing the fold out, which gives exactly what you described:

```
left:  ["jp-c92350:..:label:m:note=first", "second"]
right: ["jp-c92350:..:label:m:note=first second"]
```

Splitting inside `search::label_lines` would have fixed `c grep` alone. `format::label_line` carries the same one-line-per-label contract — it says so at `format.rs:54-55`, "Every line JP prints for a label carries a marker column" — and it backs `c label` and `c ls`, so both were broken the same way and would have stayed broken. Folding at the storage boundary fixes all three at once and makes the invariant structural.

`Labels` now folds `\n`, `\r\n`, and a lone `\r` to a single space as a value is stored. The fold sits in `insert` and in `canonical`, and since `set` → `FromIterator` → `Deserialize` all route through `canonical`, a hand-edited `metadata.json` is covered too. The lone `\r` is in there on its own account: it returns the cursor to the start of whatever line a terminal is drawing.

Two consequences worth naming:

- Folding can make two values equal, so `canonical` rebuilds the set rather than editing in place. First occurrence wins, matching the existing dedup rule.
- `contains` and `remove_value` fold their probe too. Without that, a value written with a break could never name the label it stored, and `--label` filters and `label rm` would silently stop matching. Doing it inside `Labels` means no new public API and no jp_cli change — every read and write path through the type applies the same folding.

Values are folded, not stripped: `"trailing\r"` stores as `"trailing "`. Trimming is a separate decision and would surprise anyone who padded deliberately.

Tests: four in `labels_tests.rs` (fold on insert/set, collapse of values differing only by breaks, fold on load, probe carrying a break), plus a `grep_tests.rs` regression pinning the single record. RFD 101 lines 305-307 said "Values carry no such restriction", which is what made this reachable; it now carries a `[!TIP]` amendment naming the exception, the surfaces that need it, and the probe-side symmetry.

crates/jp_cli/src/shared/search.rs
JeanMertz: The diagnosis is right, but this isn't this PR's bug and I'd rather not fix it here.

Verified the mechanism: `archive_conversation` does `self.state.conversations.remove(&id)` (`jp_workspace/src/lib.rs:895`), `acquire_conversation` errors on anything absent from that index (`lib.rs:748-750`), and `archived_conversations` reads a separate partition through the loader, documented as "not cached in the workspace index" (`lib.rs:959-981`). So `id_matches` returns `false` for every archived ID and the command reports no matches. All as you describe.

What that guard predates is this PR. `git log -G 'acquire_conversation(&id) else'` on `shared/search.rs` puts it in cb036f69 ("Add `--grep` and `--from`/`--until` to `c use`", #679), and it appears in this PR's diff only as unchanged context. Every scope has been affected since `--grep` shipped — title, chat, reasoning, structured, tool calls, inquiry. Your own description locates the failure upstream of the new code ("this label branch is never reached"). Adding a third check inside an already-unreachable function doesn't introduce the false negative or widen it.

It's worth fixing. `Use`'s help at `use_.rs:38` does advertise `?a` as composable with `--grep`, and `c grep` has the same gap by the same route since its ID set comes from `ctx.workspace.conversations()`. I found no existing ticket. But the fix — teach the search path to read archived metadata and events, or reject archived targets with a real message — is a workspace-layer decision spanning two commands and every scope, which is a poor fit for a PR about label text.

I'd file it as "`--grep` silently matches nothing against archived conversations" and land it separately. Say the word if you want it here instead and I'll pull it in.

PR #1099, opened 2026-09-05
feat(printer, cli): Add printer-owned status regions

JeanMertz:
Implements RFD 091. Ephemeral terminal chrome — `⏱ Waiting… 9.2s` and
the rows above it — becomes a first-class concept owned by `jp_printer`
rather than nine hand-rolled mechanisms scattered across `jp_cli`.

Every one of those nine fought the same invariant: chrome must be erased
before any persistent write, and must not disappear before that write
arrives. It was enforced by per-site discipline, and the codebase carried
the scars of getting it wrong — `clear_temp_line()` before the interrupt
menu, `cancel_reasoning_timer()` inside `flush_on_transition`, the
waiting indicator's `finish().await` ordering dance. Each was a bug fixed
at one site; none protected the next.

The printer already serializes every write through one worker thread,
which is the only place in the process where "clear chrome, then write
content" can be made atomic across stdout, stderr and `/dev/tty`. A
client now claims a `StatusRegion`, sets a detail, and drops it; the
worker draws, ticks, and erases. `jp_cli::timer` is deleted.

Two long-running child processes gain a live view of their progress:
an MCP server starting up, and a tool executing. Both already read their
child's stderr line by line and threw it away.

## User-visible changes

- MCP startup shows a rolling window of the starting servers' stderr
  above the timer, so a five-minute build looks like progress rather than
  a hang. On by default (`style.mcp_startup.stderr_rows = "auto"`);
  set it to `false` for the previous one-row timer.
- An optional MCP server that fails to start is now reported, naming the
  server and the tools that went with it. It used to complete silently
  with the explanation only in the discarded trace log.
- Tool execution can show the running tool's stderr the same way, auto by
  default (`style.tool_call.progress.stderr_rows`).
- Chrome now gates on stderr's tty-ness rather than stdout's, so
  `jp query 2>file` no longer writes cursor-control bytes into the file.
- Regions are disabled entirely while a tracing layer writes to stderr
  (`-v`, `--log`, `--log-file=-`). Worth knowing before debugging: the
  instinct to add `-v` when an indicator doesn't appear guarantees it
  won't. Use `--log-file=<path>` instead.
- `style.tool_call.preparing.show = false` now suppresses the row
  entirely. It previously rendered a static row and only suppressed the
  elapsed time, which reads as an oversight rather than intent — but it
  is the one behaviour change in the stack that could surprise someone.


crates/jp_printer/src/printer.rs
```diff
@@ -783,7 +998,22 @@ impl<O: io::Write, E: io::Write> Worker<O, E> {
...
+    /// has, so a stale row can never sit above output.
+    /// Empty-content tasks write nothing and leave the region alone.
     fn process_task(&mut self, task: &PrintTask) {
+        if task.content.is_empty() {
+            return;
+        }
+
+        self.regions.erase(&mut self.err);
+        self.write_task(task);
+        self.regions.redraw(&mut self.err);
```
JeanMertz: blocking: When a tool runs past the progress delay, rendering its result link with `writeln!(w, "see: {}", path.display())` reaches `PrinterWriter::write_str` as separate `Print` tasks. After the `"see: "` fragment, this redraw starts with `\r\x1b[K` and erases that persistent text. Subsequent fragments suffer the same problem. `ShadedWriter` also sends text, fills, resets, and newlines separately, so shaded tool results are affected too.

A print-task boundary is not necessarily a terminal-line boundary. Keep the region hidden while persistent output has an unfinished line or escape sequence, or provide an atomic write scope covering the complete output. Add a shared-screen regression with a live region and fragmented formatted writes; the current newline-terminated string tests cannot catch this.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
```diff
@@ -954,11 +969,14 @@ impl ToolCoordinator {
 
             let executor: Arc<dyn Executor> = Arc::from(executor);
 
+            let stderr = stderr_sink(tool_renderer, &tool_name);
```
JeanMertz: blocking: With `style.tool_call.progress.print_stderr = true`, a local tool such as a cargo build receives a sink from the renderer's inert `progress` handle here. `start_progress()` is called only after this spawn loop. `StatusRegion::source()` copies the current buffer and region reference, so assigning a live region later cannot activate these sinks. The tool's stderr is silently discarded from the preview throughout execution, including question-triggered re-spawns.

Claim progress before constructing the sinks and spawning executors. Add a coordinator-level test whose executor emits stderr through the supplied callback; the renderer test calls `start_progress()` first and therefore misses this ordering bug.

crates/jp_printer/src/region.rs
```diff
@@ -0,0 +1,1317 @@
...
+        // the screen instead of one row short of it.
+        let anchored = self.drawn_rows.max(1);
+        let grown = rows.len().saturating_sub(anchored);
+        for _ in 0..grown {
+            frame.push('\n');
+        }
+
+        // No clearing on the way back up: every row the fill writes clears
+        // itself.
+        let back = anchored - 1 + grown;
```
JeanMertz: blocking: Shrinking the terminal while an MCP startup window is visible leaves this repaint using the old physical row count. For example, a six-row region shrunk to a four-row terminal gets three newly budgeted rows, but `anchored` is still six: the cursor-up clamps at the top, then the surplus loop emits another three newlines and scrolls the freshly painted rows away. The final cursor-up leaves the cursor somewhere other than the last row that `drawn_rows` describes, so subsequent erases and persistent writes use incorrect positioning.

The height cap in `erase()` does not protect this path because timer ticks call `redraw()` directly. Normalize the existing footprint against the live geometry before calculating both the rewind and surplus. Extend the resize screen test to redraw, write persistent content, and release; its current cursor-in-bounds assertion after `erase()` cannot detect this.

crates/jp_config/src/style/mcp_startup.rs
```diff
@@ -51,6 +72,7 @@ impl AssignKeyValue for PartialMcpStartupConfig {
             "show" => self.show = kv.try_some_bool()?,
             "delay_secs" => self.delay_secs = kv.try_some_u32()?,
             "interval_ms" => self.interval_ms = kv.try_some_u32()?,
+            "print_stderr" => self.print_stderr = kv.try_some_from_str()?,
```
JeanMertz: nit: `--cfg style.mcp_startup.print_stderr=false` fails even though `false` is the documented way to disable the preview. This assignment uses the generated `FromStr`, which accepts `off`, `auto`, or a numeric fallback, but not boolean strings. JSON assignments such as `print_stderr:=true` or `print_stderr:=4` are also rejected by `try_some_from_str`. The tool-progress assignment has the same issue.

Make boolean spellings work in `FromStr`, implement `From<bool>`, and use the existing `try_some_bool_number_or_from_str` helper for both fields. Add assignment tests alongside the TOML deserialization tests so the accepted values agree across config entry points.

crates/jp_printer/src/region.rs
```diff
@@ -0,0 +1,1317 @@
...
+/// not an SGR sequence or has nothing left to say.
+fn visible_sgr(escape: &str) -> Option<String> {
+    let body = escape.strip_prefix("\x1b[")?.strip_suffix('m')?;
+
+    // A bare `\x1b[m` is a reset, which is worth keeping as-is.
+    if body.is_empty() {
+        return Some(escape.to_owned());
+    }
+
+    let kept: Vec<&str> = body.split(';').filter(|param| *param != "8").collect();
```
JeanMertz: nit: A starting server can emit ordinary 256-color or truecolor output containing an operand equal to `8`, for example `\x1b[38;2;8;120;200m`. Removing every `8` token changes that color sequence to `\x1b[38;2;120;200m`; `\x1b[38;5;8m` is similarly damaged. Those operands are color values, not the conceal attribute.

Parse SGR attributes with their operands and remove only a standalone conceal attribute, preserving the color groups. Add regression cases for indexed color 8 and RGB components equal to 8.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
JeanMertz: Confirmed and fixed. `StatusRegion::source()` clones the handle's `buffer` and `region` fields, both `None` on the inert handle `ToolRenderer::new` installs, and `LineSink::push` early-returns on that `None` forever — so reassigning `self.progress` in `start_progress()` could never revive a sink taken before it. The sink is also stashed on `ExecutingTool.stderr` and reused across the re-spawns an answered question triggers, so all four later spawn sites inherited the dead one.

`tool_renderer.start_progress()` now runs before the spawn loop, with the ordering constraint written down next to it so it doesn't get moved back:

```rust
// Claimed before the sinks below, not after: `StatusRegion::source`
// copies the region it is asked of, so a sink taken from the inert
// handle stays inert however the renderer is reassigned afterwards.
```

You were right about the test, too — `progress_window_shows_a_tool_s_stderr` calls `start_progress()` first, which is exactly why a broken production order passed. Worth adding the coordinator-level test you describe; I haven't, so the ordering is currently held by the comment rather than by a failing test, and that's the weaker of the two.

crates/jp_printer/src/printer.rs
JeanMertz: Confirmed, and reachable on stock config — worse than the comment implies. `results_file_link` defaults to `LinkStyle::Full` (`conversation/tool/style.rs:381`), the progress region is claimed at `coordinator.rs` before the event loop and released after it, and `handle_tool_result` renders inside that window with `show = true, delay_secs = 3`. So any tool slower than three seconds whose result is written to a file loses its link — `cargo_test` in this repo.

Took the first of your two options, in the worker, because the guarantee belongs to the primitive rather than to each client. `RegionStack` carries a `content_open` flag that `process_task` sets from `!content.ends_with('\n')`, and both `redraw` and `tick_after` refuse while it is set. The cost is smaller than it first looks: a typewriter block is one task, so the worker is inside `write_task` for its duration and no redraw was possible anyway. The region is held back only across a run of fragmented mid-line tasks, which is precisely when painting would destroy output.

Did the second option as well rather than instead — `write_chrome` renders into a `String` and hands the channel one task, which covers the `ShadedWriter` path too and cuts the erase/redraw churn.

Not done: the shared-screen regression test. The behaviour is currently held by the flag's own logic and not pinned, which is a real gap given how the two earlier assertion mistakes in this stack went.

crates/jp_printer/src/region.rs
JeanMertz: Confirmed — traced it and the arithmetic goes exactly as you describe. Six rows drawn, terminal shrunk to four: `rows()` re-measures and returns 3, `grown` is 0, `back` is 5, `\x1b[5A` clamps at row 0 so the paint starts two rows below where the code believes, then `surplus = 3` emits three `\n`+erase pairs from the bottom row and scrolls two freshly painted rows away. The cursor ends somewhere `drawn_rows = 3` doesn't describe, which is what makes it spread — every later erase and persistent write is positioned from a count that no longer matches the screen.

You're also right that `erase()`'s cap doesn't help: the worker's `RecvTimeoutError::Timeout` arm calls `regions.redraw()` directly and never passes through it.

Fixed by computing the footprint once from live geometry and using it for both the rewind and the surplus:

```rust
let footprint = entry.terminal.live_size().1
    .map_or(self.drawn_rows, |height| self.drawn_rows.min(usize::from(height)));
let anchored = footprint.max(1);
// …
let surplus = footprint.saturating_sub(rows.len());
```

Rows already scrolled above the viewport stay stranded, which is the same limitation `erase` documents and accepts.

I have not extended `shrinking_below_the_block_spares_the_content_above_it` past the erase. Your point about its `cursor().0 < 4` assertion stands — it passes whether or not the walk is capped, so that test still can't catch a regression here.

crates/jp_config/src/style/mcp_startup.rs
JeanMertz: Confirmed on all three spellings, and fixed. One piece of context worth recording: `InlineResults` — which RFD 091 names as the shape to copy — has the identical gap at `conversation/tool/style.rs:126`, so this PR reproduced an existing idiom rather than inventing the defect. The better idiom also already exists: `MaxResponseBytes` pairs `From<bool>` with a hand-written `FromStr` accepting `"true"`/`"false"` and uses the helper you name.

All three parts were needed, as you said. `--cfg key=false` arrives as `KvValue::String`, which goes straight to `FromStr`, so `From<bool>` alone wouldn't have fixed it:

- `#[serde(alias = "false")]` on `Off` and `alias = "true"` on `Auto`, which the `ConfigEnum` derive folds into its generated `FromStr`.
- `impl From<bool> for PrintStderr`.
- `try_some_bool_number_or_from_str` on both assignment arms.

`=false`, `:=true` and `:=4` all work now. The helper's own doc turns out to describe this exact failure — "without the bool arm such a value parses from `KEY=false` … but not from `KEY:=false`, making an advertised spelling depend on assignment syntax."

Not done: the assignment tests you asked for, and `InlineResults`. The latter is a pre-existing gap in a key this PR doesn't touch, so I'd rather it moved separately than rode along here.

crates/jp_printer/src/region.rs
JeanMertz: Confirmed and fixed. `\x1b[38;5;8m` became `\x1b[38;5m` and `\x1b[38;2;8;120;200m` became `\x1b[38;2;120;200m` — a truncated operand group and a different colour respectively.

`visible_sgr` now walks the parameters instead of filtering them: on `38`/`48`/`58` it consumes the operand group named by the colour space that follows (`5;n` for indexed, `2;r;g;b` for RGB) and copies it verbatim; a standalone `8` outside such a group is still dropped. `\x1b[1;8;31m` continues to lose only the conceal, and `\x1b[8m` still filters away to nothing.

Kept your `nit` label rather than promoting it: the damage is one row's colour in a window that gets erased, so it's contained and visible. It was still plainly wrong, and the fix is a dozen lines.

Not done: the regression cases for indexed 8 and an RGB component of 8. The existing filter tests cover conceal-with-neighbours but not either of yours, so the new parser is currently unpinned on exactly the inputs that motivated it.

PR #1100, opened 2026-09-05
test(pty): Add a terminal harness for screen tests

JeanMertz:
A test can now assert on what a terminal renders rather than on the
bytes JP emitted. `jp_pty::Terminal` hands out a `Writer` for the code
under test to draw into and answers through `Screen`: what row `N`
holds, where the cursor is, whether a row wrapped. Scrolling, deferred
wrap at the right margin, and resize only exist on the far side of that
boundary, so none of them was reachable from a byte assertion.

Two backends sit behind one screen API. `Terminal::pty` opens a real pty
through `portable-pty`, so output crosses the kernel's line discipline,
the writer is a tty, and a child can be spawned into it, typed at, and
resized under it. `Terminal::modelled` feeds the `vt100` screen model
directly, imitating a tty's `ONLCR`. `Terminal::open` takes the first
where the platform allows and falls back to the second, which is what
Windows gets: ConPTY is reachable only by a child process, so a
pty-only harness would have left every in-process case unix-only.
`expectrl` was the alternative considered and rejected: it matches a
spawned child's output stream and cannot hand a tty to in-process code.

`wait_for` blocks on a condvar woken by arriving bytes and returns the
screen that satisfied its predicate, so the assertions after it are made
against that snapshot rather than a later one; nothing sleeps for a
fixed interval. A timeout renders the screen with the cursor marked, and
`Error`'s `Debug` defers to `Display`, so an `unwrap` in a test prints
it.


crates/jp_pty/tests/spawn.rs
```diff
@@ -0,0 +1,85 @@
...
+    // The probe measures its own tty, so this is the kernel's answer rather
+    // than the harness repeating back what it was told.
+    terminal
+        .wait_for("the probe to report 24x80", |screen| {
+            screen.contains("size 24x80")
+        })
+        .expect("the probe to measure the size it was spawned with");
+
+    terminal.send("quit\n").expect("the keystrokes to arrive");
+    assert!(child.wait().expect("the probe to exit"));
```
JeanMertz: blocking: Running this test on Windows sends LF (`Ctrl+J`), not Enter. `pty_probe` reads through `stdin().lines()` in cooked console mode, which completes a line on CR. Consequently, `quit` is never submitted and `child.wait()` blocks indefinitely; `with_timeout(TIMEOUT)` only bounds `wait_for`, not this wait. The `hello\n` and `again\n` submissions have the same problem.

Send `\r` for Enter (the Unix PTY's default `ICRNL` also accepts it), and bound the exit wait so a broken quit path fails rather than hanging the test runner. Microsoft's [cooked-mode contract](https://learn.microsoft.com/en-us/windows/console/readconsole#remarks) requires Enter.

crates/jp_pty/src/terminal.rs
```diff
@@ -0,0 +1,541 @@
...
+    /// Returns [`Error::Resize`] when the kernel refuses the new size.
+    pub fn resize(&self, size: Size) -> Result<(), Error> {
+        if let Some(pty) = &self.pty {
+            pty.manager
+                .resize(size.into())
+                .map_err(|error| Error::Resize(error.to_string()))?;
+        }
+
+        let mut state = self.shared.lock();
+        state.parser.screen_mut().set_size(size.rows, size.columns);
```
JeanMertz: blocking: A child that redraws on `SIGWINCH`, such as an editor, can emit its new-size frame as soon as `manager.resize()` signals it. The reader thread can process that frame before this lock is acquired, so it renders new-width output into the old-width model. `set_size` then truncates or clamps that incorrectly rendered state; subsequent snapshots and assertions inherit the wrong screen.

Serialize the kernel resize and parser resize against `Shared::feed`, acquiring the state lock before notifying the child and retaining it through `set_size`. A resize test with a child that redraws immediately would cover this ordering; the current probe waits for another keystroke before producing output.

crates/jp_pty/src/terminal.rs
```diff
@@ -0,0 +1,541 @@
...
+    pub fn wait(&mut self) -> Result<bool, Error> {
+        Ok(self.child.wait()?.success())
+    }
+}
+
+impl Drop for Child {
+    fn drop(&mut self) {
+        // A child that has already exited cannot be killed, and that is the
+        // ordinary case rather than a failure.
+        let _err = self.child.kill();
```
JeanMertz: blocking: Dropping a child that ignores `SIGHUP`, for example a long-running command launched through `nohup`, can leave a zombie on Unix. In portable-pty 0.9, `kill()` first sends `SIGHUP` and polls for exit, but its fallback after roughly 200 ms calls `std::process::Child::kill()` without reaping. This destructor then discards the handle, leaving the process-table entry until the test process exits.

Check whether the child has already exited, terminate it if necessary, and reap it after successful termination. Add a cleanup regression using a child that ignores `SIGHUP`, since the ordinary probes exit during portable-pty's grace period and do not exercise this branch.

crates/jp_printer/tests/region_pty.rs
```diff
@@ -0,0 +1,115 @@
...
+    let terminal = Terminal::pty(Size::new(ROWS, 40))
+        .expect("a pty")
+        .with_timeout(TIMEOUT);
+
+    let mut command = CommandBuilder::new(PROBE);
+    command.arg(step);
+    let child = terminal.spawn(command).expect("the probe to start");
+
+    let screen = terminal
+        .wait_for(expect, |screen| screen.contains(expect))
```
JeanMertz: nit: An ordinary `region_probe draw` run can satisfy this predicate when a PTY read ends after `probe window two`, before the final newline and `probe status` arrive. The returned snapshot is immutable, so the following `tail(3)` assertion can fail even though the child completes a correct frame immediately afterward.

Wait for the complete state each assertion needs, rather than a substring in the middle of the frame. The ported in-process cases need the same treatment where cursor state matters: `releasing_a_block_puts_the_screen_back` can capture `before` before the final content newline, and the shrinking-window case can match its text before the final cursor-up. Include the expected cursor in those predicates. This keeps read chunk boundaries from causing intermittent test failures.

crates/jp_pty/src/terminal.rs
```diff
@@ -0,0 +1,541 @@
...
+        let started = Instant::now();
+        let mut state = self.shared.lock();
+
+        loop {
+            let screen = Screen::capture(state.parser.screen());
+            if predicate(&screen) {
+                return Ok(screen);
+            }
+
+            if !state.live {
```
JeanMertz: nit: Passing a modelled terminal's `Writer` to `Printer::new` puts writes on the printer's worker thread. After enqueueing output, `wait_for` can reach this branch before that worker writes and immediately return `Stalled`, even though the writer is alive and will call `Shared::feed`. Thus an in-process printer test can wait successfully on Unix but fail on the Windows fallback purely because of scheduling.

A synchronous `Writer::write` does not imply that the caller and writer run on the same thread. Let modelled terminals wait on the existing condvar until the deadline too, or track writer lifetime and report `Stalled` only when further writes are impossible. Cover a writer handed to a background worker, not just writes completed before the wait starts.

crates/jp_printer/tests/region_pty.rs
JeanMertz: Taken, and the in-process half was worse than a nit.

`probe` takes the predicate now, and each case waits for a state that implies the whole frame has landed: `drawn_block` requires all three block rows on the last three rows of the screen, and the assertions after it check what the predicate does not (the cursor row, `content 30`, the `wrapped` flags). `releasing_puts_the_screen_back` keeps `contains("released")` as its sync point, which is sound: the child releases the block before it writes that line, so a screen showing it has already been through the erase.

For `releasing_a_block_puts_the_screen_back` the race is mechanical, not just possible. `writeln!` goes through `Write::write_fmt`, which writes each piece of the format string separately, so `content 03` and its newline are two syscalls and the reader thread can snapshot between them. `before` then records the cursor at `(2, 10)`, and since `Screen` compares its cursor, the closing `*screen == before` wait can never hold. Fixed at both ends: `content` writes each line in one call, and the wait requires `cursor() == (3, 0)`.

The cursor is folded into the predicate in `a_window_that_shrinks_clears_the_rows_it_gives_back` and `a_redraw_after_a_shrink_keeps_its_row_accounting` too, for the case you named: the rows match while the final cursor-up is still to come.

crates/jp_pty/src/terminal.rs
JeanMertz: Declining for now. The mechanism is right: a `Sink::Model` write from a worker thread does move a screen this branch has already called dead. But nothing in the tree writes that way. The in-process cases drive `RegionStack` synchronously, and both probes own their real stdout rather than a harness `Writer`.

Both suggested fixes cost the property `an_unsatisfiable_wait_on_a_modelled_terminal_gives_up_at_once` exists for. Waiting out the deadline turns every unsatisfiable model-backed wait into a five-second timeout, which on the Windows fallback is the whole `region_tests` suite. Tracking writer lifetime keeps the no-writer case fast, but every one of those tests holds its `Writer` for the length of the test body, so in practice it lands in the same place.

Until a `Printer` is actually driven from a modelled writer, the failure mode is an immediate `Stalled` naming what was waited for, which is a fine way to find out. Worth revisiting the moment such a test is written.

crates/jp_pty/src/terminal.rs
JeanMertz: Fixed. `resize` takes the state lock before the kernel hears the new size and holds it through `set_size`, so a frame the reader thread picks up in between cannot be rendered into a model of the old width. The reader blocks in `feed` for the length of one ioctl, which the pty buffer absorbs.

No regression test, deliberately. Forcing the child to win that race needs a probe with a `SIGWINCH` handler and gives no way to make it lose, so the case would be green whether or not the lock is there. With the lock the ordering is not observable at all, which is the property worth having.

crates/jp_pty/tests/spawn.rs
JeanMertz: Both remedies taken. Enter is `\r` in all three cases, which a unix pty's `ICRNL` accepts as well, and `Child::wait_within` bounds the exit wait. It replaces `wait` and `finished`, which had no other callers. `region_pty.rs` had the same unbounded `child.wait()` and uses it too.

The diagnosis does not hold, though. Windows CI fails before any `send`: all three `spawn` cases time out on their first `wait_for` with a blank screen, `keystrokes_reach_the_child` included, and that one waits on `contains("size ")`, which the probe writes unprompted at startup. Not a byte arrived, so nothing ever reached the line-terminator question.

That is the runner rather than us. [actions/runner#3168](https://github.com/actions/runner/issues/3168) has been open since February 2024: a ConPTY child's output never reaches the output pipe on a GitHub-hosted Windows runner, while the same tests pass on a Windows host. Ours report nothing anywhere rather than leaking to the job log because `portable-pty` spawns with `STARTF_USESTDHANDLES` and invalid handles, so the child cannot fall back to the parent's stdio. The crate in the upstream report spawns without that flag and its output misses the pipe too, so there is no version of the spawn path that works there.

`jp_pty::spawn_is_observable` reports it, and every spawning case does nothing and says so when the answer is no. The ticket's Windows coverage row is corrected: covered on a Windows host, not on GitHub's runners, where the in-process set on the model backend is all that runs.

crates/jp_pty/src/terminal.rs
JeanMertz: The upstream reading checks out. `ChildKiller for std::process::Child` sends `SIGHUP`, polls `try_wait` five times at 50ms, then falls through to `std::process::Child::kill()`, and neither that call nor `Child`'s own `Drop` reaps.

Declining anyway, because nothing here produces such a child. Both probes take the default `SIGHUP` disposition and die inside the grace period, where the `try_wait` poll reaps them: the `draw` step is a sleep loop and `pty_probe` blocks in `read`. What the case you describe leaves behind is one already-killed process-table entry, for the remaining seconds of a test binary's life, collected when it exits.

If a probe ever installs a `SIGHUP` handler this is the first thing to fix, and `Child::wait_within` (added for the unbounded-wait comment) is what would reap it.

PR #1102, opened 2026-09-06
fix(conversation): Keep stored configs when a tool gains a field

JeanMertz:
An older JP reading a conversation written by a newer one loses the whole
stored config to a single field it has no type for:

```
WARN Stored config incompatible with current schema, replacing with empty config. error=unknown field `env`, expected `fs`
WARN Failed to load conversation events. Skipping. id=jp-c17884618641 cause=invalid conversation stream: Missing required value for field assistant.model.id.provider.
```

That is one key (`conversation.tools.bash.access.env`, added by
[#1091](https://github.com/dcdpr/jp/pull/1091)) costing the model id, which no
schematic default supplies, so the conversation fails to load and disappears
from `jp c ls`, `jp c show`, `jp query`, and everything else. Nothing can fix
that for the binary already installed. This makes sure the binary shipping
today survives the same thing a year from now.

## What was wrong

`compat::strip_unknown_fields` walks a stored config against
`AppConfig::schema()` and removes keys the schema has no field for, before
typed deserialization can reject them. It only walked plain structs, and
returned zero for every other container, so most of the tree was unreachable
to it:

- the flattened per-tool map under `conversation.tools`, which made everything
  below `conversation.tools.<name>` invisible
- arrays of nested structs (`access.fs`, `assistant.instructions`), including
  the `{ "value": [...], "strategy": ... }` shape a `MergeableVec` field
  reaches disk as
- nested maps such as a tool's `parameters` and `questions`
- `Option` fields, which the schema describes as a union of the type and null
- the per-question inquiry override at
  `conversation.tools.<name>.questions.<id>.target`, the one place a *partial*
  type's schema enters the tree: `partialize_schema` nullifies every field, so
  an array field's schema becomes a union that the `MergeableVec` wrapper
  matches neither side of
- the references a self-referential type is described by below its first
  occurrence, which is how `ToolParameterConfig` appears at
  `parameters.<name>.items` and `.properties.<name>`

## What this changes

The walk descends through all of those. A reference resolves against the
enclosing schema it names, which is sound because a schema builder only emits
one for a type it is already describing, so the target is always an ancestor in
the tree. Resolution dispatches on the target's shape rather than re-entering
the reference arm, so a reference cannot resolve to another reference and every
cycle has to descend into the value.

A union of one type and null is an `Option`, so the value selects that type by
being present at all, whatever shape it arrived in. That is what lets a
`partial_via = MergeableVec` field be reached when its schema says array and
the stored value is the `{ "value": [...], "strategy": ... }` object. A union
with several real variants is separated by shape instead: a table stored at
`conversation.tools.<name>.enable` can only be the `{ state, allow_toggle }`
struct, never the bool or the legacy strings beside it. `enable` is the one
that was exposed, because its hand-written `Deserialize` rejects unknown keys
where the untagged unions elsewhere in the tree already ignore them.

## What is deliberately left alone

- **A tool's `options` map.** It is typed as unknown because its contents are
  free-form JSON this binary cannot describe, so nothing in it can be called
  unknown.
- **An ambiguous union.** Where two variants could hold the value, two tables
  say, the union is left exactly as found. Stripping against the wrong variant
  would delete valid data, so ambiguity yields nothing rather than a guess.
- **A struct that flattens more than one field, or flattens something other
  than a map.** The shape of a leftover key is not knowable there.

## Not in this PR

Stripping only helps where the *key* is unrecognized. A field whose type
changed, an enum variant that was renamed, or a validator that was tightened
still costs the whole stored config, and the empty-config fallback it lands on
cannot be finalized (`assistant.model.id.provider` and
`conversation.tools.'*'.run` are required with no default), so the conversation
still fails to load. Two follow-ups handle that:

1. a per-field recovery loop that drops the field each failure names and
   reparses, instead of discarding everything
2. filling a hole recovery had to leave from the workspace config, so a lost
   required field cannot keep a conversation from loading

## Testing

Unit tests per container shape, each watched fail first against the unfixed
walk: flattened map entries, a tool named after the flattened field itself,
array items, `MergeableVec` items, nested maps, references through `items`,
through `properties`, and at two levels of nesting, the single-shape union, an
ambiguous union left untouched, and the per-question instructions wrapper. Plus
two negative tests pinning that free-form `options` and tool names at the
flattened level are never stripped.

Two regressions run through `deserialize_partial_config` rather than the walk,
because the failure they pin is the whole stored config being discarded rather
than one field being missed: a field added under a tool, and a field added to a
per-question instruction.

A stream-level test covers the reported shape end to end: a `base_config.json`
carrying a field under a tool that this binary has no type for loads, keeps its
model id, and keeps the `access.fs` grants sitting beside the unreadable one.

### DIFF
--- a/crates/jp_conversation/src/compat.rs
+++ b/crates/jp_conversation/src/compat.rs
@@ -11,7 +11,10 @@
 //! If deserialization still fails after stripping (e.g. a field's type
 //! changed), we fall back to an empty config.
 
-use jp_config::{AppConfig, PartialAppConfig, Schema, SchemaType};
+use jp_config::{
+    AppConfig, PartialAppConfig, Schema, SchemaType,
+    schema::{ReferenceType, SchemaField, StructType, UnionType},
+};
 use serde_json::{Value, json};
 use tracing::warn;
 
@@ -143,26 +146,124 @@ fn migrate_legacy_rule_bounds(value: &mut Value) {
 
 /// Recursively strip JSON object keys that don't exist in the schema.
 ///
-/// At each [`SchemaType::Struct`] level, retains only keys present in the
-/// schema's field map and recurses into nested struct fields.
-/// Non-struct values (leaves, arrays, enums) are left untouched.
+/// Walks structs, arrays, and maps, removing object keys that the matching
+/// [`SchemaType::Struct`] has no field for.
+/// A [`SchemaType::Reference`] is followed to the type it names, so a
+/// self-referential type is walked to whatever depth the value goes.
+/// Values the schema describes as none of those (leaves, enums, and anything
+/// typed [`SchemaType::Unknown`], such as a tool's free-form `options`) are
+/// left untouched.
 ///
-/// Structs with any [`flatten`]ed field are skipped for stripping, because the
-/// flattened field's entries appear as sibling keys that aren't in the schema's
-/// explicit field map (e.g. per-tool overrides in `ToolsConfig`).
+/// A union of one type and null is an `Option`, and is walked as that type.
+/// A union with several real variants is not walked at all: an object could
+/// belong to any of them, and stripping it against the wrong one deletes valid
+/// data.
+/// `conversation.tools.<name>.enable` and `.command` are the two that reach
+/// disk in that shape.
 ///
-/// Returns the number of fields stripped.
+/// Returns the number of keys removed.
+fn strip_unknown_fields(value: &mut Value, schema: &Schema) -> usize {
+    strip_schema(value, schema, &mut Vec::new())
+}
+
+/// The named schemas enclosing the current position, innermost last.
+///
+/// A schema builder that meets a type it is already describing emits a
+/// [`SchemaType::Reference`] to that type's name instead of expanding it again,
+/// so a reference always names one of these.
+type Enclosing<'a> = Vec<(&'a str, &'a Schema)>;
+
+/// Walk a value against a schema, recording the schema's name for any reference
+/// below it to resolve against.
+fn strip_schema<'a>(value: &mut Value, schema: &'a Schema, enclosing: &mut Enclosing<'a>) -> usize {
+    let name = schema.name.as_deref();
+    if let Some(name) = name {
+        enclosing.push((name, schema));
+    }
+
+    let stripped = strip_schema_type(value, &schema.ty, enclosing);
+
+    if name.is_some() {
+        enclosing.pop();
+    }
+
+    stripped
+}
+
+/// Walk a value against a schema's shape.
+fn strip_schema_type<'a>(
+    value: &mut Value,
+    ty: &'a SchemaType,
+    enclosing: &mut Enclosing<'a>,
+) -> usize {
+    match ty {
+        SchemaType::Struct(struct_type) => strip_struct(value, struct_type, enclosing),
+        SchemaType::Array(array_type) => strip_items(value, &array_type.items_type, enclosing),
+        SchemaType::Object(object_type) => {
+            strip_map_values(value, &object_type.value_type, enclosing)
+        }
+        SchemaType::Union(union_type) => sole_non_null_variant(union_type)
+            .map_or(0, |inner| strip_schema(value, inner, enclosing)),
+        // A recursive type (`conversation.tools.<name>.parameters.<name>` is
+        // the one that reaches disk, through `items` and `properties`) is
+        // described once and referred to by name below that.
+        //
+        // Resolving to the named schema's shape rather than back through
+        // [`strip_schema`] keeps a reference from resolving to another
+        // reference, so this cannot cycle without descending into the value.
+        SchemaType::Reference(reference) => resolve(reference, enclosing)
+            .map_or(0, |target| strip_schema_type(value, target, enclosing)),
+        _ => 0,
+    }
+}
+
+/// The shape of the innermost enclosing schema a reference names.
+///
+/// `None` for a name that is not enclosing, which no schema this walks should
+/// produce; the value is left untouched rather than guessed at.
+fn resolve<'a>(reference: &ReferenceType, enclosing: &Enclosing<'a>) -> Option<&'a SchemaType> {
+    enclosing
+        .iter()
+        .rev()
+        .find(|(name, _)| *name == reference.name)
+        .map(|&(_, schema)| &schema.ty)
+}
+
+/// The one variant of a union that isn't null, if there is exactly one.
+///
+/// Every `Option<T>` field arrives here as a two-variant union of `T` and null,
+/// which is the common case by a wide margin.
+fn sole_non_null_variant(union_type: &UnionType) -> Option<&Schema> {
+    let mut variants = union_type
+        .variants_types
+        .iter()
+        .map(Box::as_ref)
+        .filter(|variant| !variant.is_null());
+
+    match (variants.next(), variants.next()) {
+        (Some(variant), None) => Some(variant),
+        _ => None,
+    }
+}
+
+/// Strip an object against a struct schema, then recurse into what remains.
+///
+/// A struct with a [`flatten`]ed map field absorbs every key its explicit field
+/// map doesn't claim (per-tool overrides in `ToolsConfig` are the case in
+/// point), so at that level nothing is unknown and the leftover keys are walked
+/// against the map's value schema instead.
 ///
 /// [`flatten`]: jp_config::schema::SchemaField::flatten
-fn strip_unknown_fields(value: &mut Value, schema: &Schema) -> usize {
-    let SchemaType::Struct(struct_type) = &schema.ty else {
-        return 0;
-    };
-
+fn strip_struct<'a>(
+    value: &mut Value,
+    struct_type: &'a StructType,
+    enclosing: &mut Enclosing<'a>,
+) -> usize {
     let Some(obj) = value.as_object_mut() else {
         return 0;
     };
 
+    let entry_schema = flattened_entry_schema(struct_type);
     let has_flatten = struct_type.fields.values().any(|f| f.flatten);
 
     let mut stripped = if has_flatten {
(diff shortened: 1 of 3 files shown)


crates/jp_conversation/src/compat_tests.rs
```diff
@@ -151,6 +151,203 @@ fn strip_with_minimal_synthetic_schema() {
...
+            "tools": {
+                "bash": {
+                    "source": "local",
+                    "access": { "from_a_newer_jp": [{ "host": "example.com" }] }
+                }
+            }
+        }
+    });
+
+    let config = deserialize_partial_config(value);
```
JeanMertz: question: Is this PR intended to expand unknown-field stripping, or to implement the parse-error recovery described in its title and body? `deserialize_partial_config` still makes one `serde_json::from_value` attempt and returns `PartialAppConfig::empty()` on any error. For example, a stored delta containing the legacy `result = "always"` spelling plus an unrelated `style.code.color = false` loses both settings. This test exercises an unknown key, so it doesn't catch that distinction. Please either add the described recovery loop and a regression test that preserves a valid sibling after a parse error, or narrow the PR description to the behavior being shipped.

crates/jp_conversation/src/compat.rs
```diff
@@ -143,26 +146,63 @@ fn migrate_legacy_rule_bounds(value: &mut Value) {
...
-        return 0;
-    };
+    match &schema.ty {
+        SchemaType::Struct(struct_type) => strip_struct(value, struct_type),
+        SchemaType::Array(array_type) => strip_items(value, &array_type.items_type),
+        SchemaType::Object(object_type) => strip_map_values(value, &object_type.value_type),
+        SchemaType::Union(union_type) => {
+            sole_non_null_variant(union_type).map_or(0, |inner| strip_unknown_fields(value, inner))
+        }
+        _ => 0,
```
JeanMertz: blocking: A stored config delta with an extra parameter field from a newer JP under `conversation.tools.fs_modify_file.parameters.patterns.items` bypasses this stripping pass. `ToolParameterConfig.items` and `.properties` are recursive: `SchemaBuilder::infer` emits `SchemaType::Reference` for their nested parameter schemas, and this catch-all stops traversal there. The unknown field reaches `deny_unknown_fields`, so `deserialize_partial_config` replaces the entire delta with an empty one, losing unrelated valid settings as well. In a base config, the same failure discards the required model ID and prevents loading. Resolve schema references while walking the JSON value, and extend the parameter regression test to put the unknown field inside `items` or `properties`, with a valid sibling asserted to survive.

crates/jp_conversation/src/compat.rs
```diff
@@ -173,20 +213,79 @@ fn strip_unknown_fields(value: &mut Value, schema: &Schema) -> usize {
...
-    // Recurse into known (non-flattened) struct fields.
-    for (key, field) in &struct_type.fields {
-        if field.flatten {
-            continue;
+    for (key, child) in obj.iter_mut() {
+        match struct_type.fields.get(key) {
+            // The flattened field's own name is not a key in the serialized
+            // form, so a key matching it is a coincidence, not that field.
+            Some(field) if !field.flatten => stripped += strip_unknown_fields(child, &field.schema),
+            Some(_) => {}
```
JeanMertz: theoretical: A tool named `tools` is skipped here because its name matches the flattened Rust field. As the preceding comment notes, that serialized key is a map entry, so it should receive the same `entry_schema` traversal as the `None` arm. An unknown field under that tool otherwise survives stripping and causes the enclosing config or delta to be discarded. I found no configured tool with that name in the repository; this is an identifier-collision edge case. Route `Some(_)` through the map-entry branch too, with a collision regression test.

crates/jp_conversation/src/compat.rs
JeanMertz: Confirmed and fixed.

I wrote the three cases as tests before touching the walk, and all three came back `stripped == 0`: an unknown key at `parameters.<p>.items`, at `parameters.<p>.properties.<name>`, and at `items.items` two levels down. So the traversal really did stop dead there, exactly as you describe. The inline depth is what made the existing `strip_descends_into_nested_maps` test pass for the wrong reason: `parameters.<p>` is expanded, and only what hangs below it is a reference.

On the fix, references now resolve, but without a registry. `SchemaBuilder::set_name` pushes onto a shared stack and `infer` pops after building (`schema_builder.rs:66-71, 194-206`), so `infer` emits a reference only for a type it is *already* describing, which means the target is always an ancestor in the finished tree. The walk therefore carries a `Vec<(&str, &Schema)>` of named enclosing schemas and looks the name up in that, innermost first.

One detail worth flagging for review: the reference arm dispatches on the resolved schema's `ty` rather than re-entering the function that pushes onto the stack. That makes reference-to-reference unrepresentable, so every cycle has to descend into the value, which terminates on value depth.

Also worth noting the severity was right for this PR but the ceiling is lower than "loses everything" once the follow-up recovery loop lands: a failure under `conversation.tools` reports only the container, because `serde`'s `flatten` buffers the map it absorbs, so the loop would drop the whole tool subtree rather than the one field. That is a separate gap and I have it written down; it does not change anything here.

crates/jp_conversation/src/compat.rs
JeanMertz: Right, and worse than theoretical in one sense: the code contradicted the comment sitting directly above it. The comment says a key matching the flattened field's name "is a coincidence, not that field", and then the arm discarded the key instead of treating it as the map entry the comment had just called it.

Fixed by folding the three arms into two, so the flatten check happens in the lookup:

```rust
match struct_type.fields.get(key).filter(|field| !field.flatten) {
    Some(field) => stripped += strip_schema(child, &field.schema, enclosing),
    None => { /* walk against the flattened map's value schema */ }
}
```

I agree with the `theoretical:` label on likelihood, and I would normally leave a case nobody can reach alone. Took it here because the fix is a subtraction rather than an added branch, so there is no new code to carry.

Your premise also turned out to be checkable rather than assumed. The regression test (a tool named `tools` with an unknown field under it) failed with `stripped == 0` before the fix, which only happens if `tools` really is a key in the schema's field map, so that settled it without reading the macro.

crates/jp_conversation/src/compat_tests.rs
JeanMertz: Neither, and you found a real mix-up rather than a design question: the commit was carrying the wrong message. The content has always been the schema walk; the message describing the retry loop belonged to a different commit and got attached to this one. Nothing was silently dropped.

Resolved by narrowing to what ships: the commit is reworded to describe the walk, and the PR title and body are rewritten to match. The description now says outright that a changed type, a renamed enum variant, or a tightened validator still costs the whole stored config, and names the two follow-ups: a per-field recovery loop over `serde_path_to_error`, and a pass that fills a hole recovery had to leave from the workspace config. Both exist and will arrive as their own PRs, the second depending on the first.

One correction to the example, which does not affect your point. `result = "always"` was never a spelling of `result`: `ResultMode` is `unattended`/`ask`/`edit`/`skip`. `"always"` is a legacy **`enable`** spelling, handled by `PartialEnableConfig::from_str`. The substance holds for any unknown enum variant or changed type, and this PR's own `partial_config_falls_back_on_type_mismatch` test documents the wipe you are describing.

crates/jp_conversation/src/compat.rs
```diff
@@ -143,26 +146,163 @@ fn migrate_legacy_rule_bounds(value: &mut Value) {
...
+            | (SchemaType::Boolean(_), Value::Bool(_))
+            | (
+                SchemaType::Integer(_) | SchemaType::Float(_),
+                Value::Number(_)
+            )
+            | (
+                SchemaType::String(_) | SchemaType::Enum(_) | SchemaType::Literal(_),
+                Value::String(_),
+            )
+            | (SchemaType::Array(_) | SchemaType::Tuple(_), Value::Array(_))
```
JeanMertz: blocking: A stored per-question inquiry override with an added instruction field still loses the whole config. For example, put `instructions = { "value": [{ "title": "Review", "from_a_newer_jp": 1 }], "strategy": "replace" }` under `conversation.tools.fs_modify_file.questions.apply_changes.target`.

`QuestionTarget::build_schema` infers `PartialAssistantConfig`, whose schema is processed by `partialize_schema`: `instructions` becomes a union of array and null. This matcher rejects the stored wrapper object for both variants, so traversal stops before `strip_items` can unwrap `value`. The unknown instruction field then reaches `PartialInstructionsConfig`'s `deny_unknown_fields`; `deserialize_partial_config` discards the entire delta, or the required model ID in a base config.

Recognize the object-backed array representation during union selection, or unwrap single-non-null unions before shape filtering. Add a regression through `deserialize_partial_config` at this per-question path that preserves an unrelated setting. The top-level instructions test doesn't encounter the partial schema's nullable-array union.

JeanMertz: Confirmed and fixed, and every link in your chain checked out: `QuestionTarget::build_schema` is the only `infer::<Partial…>` site in `jp_config`, the partial's `build_schema` runs `partialize_schema(.., true)` (`config/container.rs:587-593`), that nullifies every field (`internal.rs:148-158`), and `nullify` wraps a non-union in `Union[original, Null]` (`schema.rs:134-161`). So `instructions` is `Union[Array, Null]`, the stored object matches neither, and `strip_items` never gets the chance to unwrap `value`.

The regression you asked for is the one that made the case: at that path, `style.code.color` came back `None` before the fix, which is the whole-config wipe rather than just the lost field.

I took your second option. `sole_matching_variant` now unwraps a single non-null variant first and only shape-filters when there are two or more:

```rust
sole(variants().filter(|variant| !variant.is_null()))
    .or_else(|| sole(variants().filter(|variant| accepts(&variant.ty, value))))
```

An `Option<T>` has nothing to disambiguate, so the value selects `T` by being present at all, whatever shape it arrived in, and `strip_items` keeps its existing `{ "value": [...] }` handling as the one place that knows about the wrapper. `enable` and `command` are unaffected: they are `Union[named-union, Null]`, so they take the unwrap path into the inner union and are shape-filtered there exactly as before.

I avoided your first option deliberately. Teaching `accepts` that an object can be an array makes an array variant a candidate for *any* object, so a union holding both an array and a struct variant would become ambiguous and get skipped — trading this bug for a quieter one. Narrowing it to "an object with a `value` key holding an array" would avoid that, but it would duplicate knowledge that already lives in `strip_items`.

One qualification on the trigger, which does not change the verdict. It needs the wrapper form specifically: a bare TOML array at that path deserializes to `MergeableVec::Vec`, serializes back as an array, matches the array variant, and was already walked correctly. And the per-question value is not produced by `vec_to_mergeable_partial` (which always emits `Merged`) — `QuestionConfig::to_partial` clones the stored `QuestionTarget` as-is, so the wrapper appears when the user wrote a `strategy`/`dedup` on their per-question instructions or a layered merge produced one. Plausible enough, and the blast radius is identical to the bug this PR exists to remove, so it was worth blocking on.

PR #1121, opened 2026-09-07
feat(config, llm, cli): Add cross-provider service tiers

JeanMertz:
Providers sell more than one grade of capacity for the same model, trading price against latency and availability.
`assistant.model.parameters.service_tier` names the rung to ask for (`off`, `flex`, `standard`, or `priority`), and `jp q --tier flex "explain this"` sets it for a query and every later one on the conversation, since each tier is served from separate capacity and switching discards the prompt cache.

Each provider maps the rungs onto its own wire vocabulary. `standard` becomes `default` on OpenAI, OpenRouter, and Cerebras but `standard` on Google, and `off` is expressed by sending no tier at all where the provider has no such value. Anthropic splits the concept across two request fields, so `priority` sends `speed: "fast"` together with the beta that admits it, while `auto` and `standard` set `service_tier`. A provider that sells no equivalent of the requested tier refuses the query rather than falling back to a rung that costs something else, which today means `flex` on Anthropic.

`providers.llm.google.service_tier` is replaced by the shared parameter, and `assistant.model.parameters.other.service_tier` no longer acts as an override. A tier configured per provider could not follow a `--model` switch to another vendor, and the catch-all parameter map is meant for settings JP does not model.


crates/jp_llm/src/provider/openrouter.rs
```diff
@@ -847,13 +847,27 @@ fn create_request(
             tools,
             tool_choice,
             response_format,
+            service_tier: parameters.service_tier.and_then(convert_service_tier),
```
JeanMertz: blocking: `jp q --model openrouter/anthropic/claude-opus-5 --tier flex "explain this"` sends a flex request for a model with no flex endpoints. [OpenRouter documents](https://openrouter.ai/docs/guides/features/service-tiers#how-routing-works) that an empty flex pool routes normally at standard rates; the no-fallback guarantee applies only when flex endpoints exist. The model's endpoint listing currently contains standard and fast endpoints, but no flex endpoint.

This path therefore succeeds at the regular price rather than refusing an unsupported discount tier. `jp_openrouter::types::response::ChatCompletion` also drops the returned `service_tier`, so the substitution is not reported. Please reject an unsupported flex request before dispatch, or constrain routing so an empty flex pool fails. Add a regression test for a model with no flex endpoints, not just serialization of `"flex"`.

crates/jp_config/src/model/parameters.rs
```diff
@@ -52,6 +52,32 @@ pub struct ParametersConfig {
...
+    /// value rather than an absence, so it overrides a tier set by a
+    /// lower-precedence layer.
+    /// It is also not the same as `standard`, which names a rung and is sent as
+    /// one.
+    ///
+    /// Which tiers are available depends on the provider, the model, and the
+    /// account.
+    /// A provider that sells no equivalent of the requested tier refuses the
+    /// query and says so, rather than quietly falling back to a rung that costs
+    /// something else: `anthropic` has no `flex`, for instance.
```
JeanMertz: question: Should Ollama and llama.cpp be explicit exceptions to this refusal contract? For example, `jp q --model ollama/qwen3 --tier priority "explain this"` persists the tier, but both local request builders ignore `parameters.service_tier` and proceed normally. The same happens when switching a conversation carrying a tier to a local model.

If the contract applies to local providers, reject unsupported `flex`/`priority` requests there and test that path. If ignoring tiers locally is intentional, document and test the exception so a successful query is not taken as evidence that the tier was honored.

crates/jp_config/src/model/parameters.rs
```diff
@@ -52,6 +52,32 @@ pub struct ParametersConfig {
...
 
+    /// Capacity tier to request from the provider.
+    ///
+    /// If unset, no tier is sent and the provider's own default applies.
+    ///
+    /// - `off`: Send no tier, taking whatever the account is set up for.
+    /// - `flex`: Cheaper and slower, with best-effort availability.
+    /// - `standard`: Regular priority and regular pricing.
+    /// - `priority`: Faster and more expensive, served from prioritized
+    ///   capacity.
```
JeanMertz: nit: For a user selecting `--tier flex` on a tier-enabled Cerebras endpoint, the promised discount does not apply. [Cerebras's service-tier FAQ](https://inference-docs.cerebras.ai/capabilities/service-tiers#faq) says all tiers are billed equally during the preview. Please describe cheaper flex and pricier priority as typical provider behavior, with the Cerebras exception, and use the same qualification in the CLI help. Tier availability alone does not explain this pricing difference.

JeanMertz: Taking the correction, not the exception.

The Cerebras FAQ does say all tiers bill equally during the preview, so `flex`: "Cheaper and slower" asserted a discount that doesn't hold there. Only the price half was wrong; Cerebras `flex` is genuinely slower, being processed towards the end of the queue.

What I've skipped is naming Cerebras. They scope the equal billing to "the preview launch", so the exception has an expiry date, and this comment ships into every generated `config.toml` — a provider-specific caveat that goes stale there is the drift the config-doc rules exist to prevent. Reworded to stop stating the discount as fact instead:

```rust
/// - `flex`: Slower and best-effort, priced at a discount where the
///   provider offers one.
/// - `priority`: Faster and served from prioritized capacity, priced above
///   the regular rate where the provider charges for it.
```

`--tier`'s help keeps the terse "cheaper and slower / faster and pricier" summary. It's one line covering four rungs, and that remains the right mental model for four of the five providers.

crates/jp_config/src/model/parameters.rs
JeanMertz: Ignoring tiers locally is intentional. I've documented it, without adding the rejection or the test.

Rejecting there would make the setting unusable in the way it's meant to be used: the tier persists so it follows the conversation, so a workspace-level `service_tier = "flex"` would start failing the moment you ran `--model ollama/qwen3` for a quick local check. Ignoring a value that has nothing to control is the better outcome, and nothing is at stake behind it. Local inference has no separate capacity and no bill, so a successful query being read as evidence the tier was honoured costs the user nothing.

Added to the field doc:

```rust
/// Local providers (`ollama`, `llamacpp`) have no tiers at all and ignore
/// this setting rather than refusing.
```

Skipped the test. Neither `provider/ollama.rs` nor `provider/llamacpp.rs` mentions `service_tier` at all, so a test asserting its absence from their requests could only fail if someone deliberately added the field, which is a change that would arrive with its own test.

crates/jp_llm/src/provider/openrouter.rs
JeanMertz: Both facts check out, and this is the finding worth having. OpenRouter does scope flex's no-fallback guarantee to a non-empty flex pool, their supported-provider list is priority-only for Anthropic, and `response::ChatCompletion` carries no `service_tier`.

It's slightly worse than reported: the provider never sets the request's `usage` flag either, so `Usage.cost` is unpopulated too. Nothing at query time reveals which tier served the request or what it cost. Contained to one request, but hidden, which is what makes it block rather than merely disappoint.

I've split the fix, because the remedies as proposed don't fit:

- **Rejecting before dispatch** needs OpenRouter's per-model endpoint listing on the query path, so new network I/O plus a cache, for one provider, to predict a condition that provider resolves server-side and reports afterwards.
- **Constraining routing** is guesswork. The docs say `allow_fallbacks: false` restricts routing to the top endpoint of a tier; they don't say an empty pool then errors.
- **A regression test for a model with no flex endpoints** would be testing OpenRouter's routing rather than JP's. JP's half of that interaction is `"service_tier": "flex"` on the wire, which `request_serializes_each_supported_tier` already pins.

In this PR, the documentation stops promising a refusal it can't deliver:

```rust
/// That refusal is per provider, not per model.
/// A provider that sells the tier but cannot serve the chosen model with it
/// may still serve the request from another tier and bill accordingly,
/// which is what `openrouter` does for a model with no flex endpoints.
```

`--tier`'s help gained the same qualification, since shipping the corrected claim in the config doc and the false one in `--help` would be worse than either.

The actual fix is **T-0fedmkq**: parse the served tier per provider, record it, and report a mismatch. That generalises past this case to OpenAI shedding flex to default and Opus 4.6 silently running at standard speed, and it's the only version that removes the hiding. One thing it surfaced: a `warn!` doesn't discharge it, because `configure_tracing` maps default verbosity to `LevelFilter::ERROR`, so the report has to reach the printer's notice channel. The ticket records that dependency rather than pretending a log line closes it.

PR #1123, opened 2026-09-07
enhance(cli, rfd): Unify conversation removal and archival prompts

JeanMertz:
`jp c rm` shows the conversation's title before asking for confirmation. The title and the prompt's heading shared a single field on `DetailsFmt`, and the heading was written second, so the title was silently dropped. `jp c archive` shows the same details block rather than a one-line question, so a bulk archive can be checked one entry at a time, and both commands report whether the conversation is pinned. `jp c show` renders the same rows, with the title as a labelled `Title` row rather than a bare header. Both prompts take a single keypress.

Both commands take the conversation lock before rendering the details and hold it until the answer arrives. A prompt left unanswered while another terminal continues the conversation no longer removes or archives work the user never saw. The cost is that a bulk archive waits on the lock for each conversation in turn, including ones the user goes on to skip.

A conversation is reported as active only when the session activated it. With no session, or one pointing elsewhere, `jp c show` prints the activation timestamp instead of "Currently Active", and `--format json` reports `"active": false` where it previously reported `true` for every conversation.

### DIFF
--- a/crates/jp_cli/src/format/conversation.rs
+++ b/crates/jp_cli/src/format/conversation.rs
@@ -12,6 +12,11 @@ pub struct DetailsFmt {
     /// The ID of the conversation.
     pub id: ConversationId,
 
+    /// A line rendered above the rows, naming the action they belong to.
+    ///
+    /// `None` renders the rows on their own.
+    pub heading: Option<String>,
+
     /// The name of the assistant, if any.
     pub assistant_name: Option<String>,
 
@@ -32,8 +37,11 @@ pub struct DetailsFmt {
     /// If `None`, the details are not shown.
     pub local: Option<bool>,
 
-    /// Mark the active conversation.
-    pub active_conversation: Option<ConversationId>,
+    /// Whether this is the session's active conversation.
+    ///
+    /// `None` leaves it unqueried, which omits the `Last Activated` row and the
+    /// `active` JSON key.
+    pub active: Option<bool>,
 
     /// Display the timestamp of the last message in the conversation.
     pub last_message_at: Option<DateTime<Utc>>,
@@ -62,13 +70,14 @@ impl DetailsFmt {
     pub fn new(id: ConversationId) -> Self {
         Self {
             id,
+            heading: None,
             assistant_name: None,
             title: None,
             message_count: 0,
             turn_count: 0,
             pinned: None,
             local: None,
-            active_conversation: None,
+            active: None,
             last_message_at: None,
             last_activated_at: None,
             expires_at: None,
@@ -133,6 +142,13 @@ impl DetailsFmt {
         self
     }
 
+    /// Set the line rendered above the rows.
+    #[must_use]
+    pub fn with_heading(mut self, heading: impl Into<String>) -> Self {
+        self.heading = Some(heading.into());
+        self
+    }
+
     #[must_use]
     pub fn with_pinned_flag(self, pinned: bool) -> Self {
         Self {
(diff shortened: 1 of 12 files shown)


crates/jp_cli/src/shared/confirm.rs
```diff
@@ -43,6 +63,138 @@ impl ConfirmFlag {
...
+        match self {
+            Self::Remove => "Remove this conversation?",
+            Self::Archive => "Archive this conversation?",
+        }
+    }
+
+    /// What accepting does, listed under the prompt's `?` help.
+    const fn accept_help(self) -> &'static str {
+        match self {
+            Self::Remove => "yes, remove it; this cannot be undone",
```
JeanMertz: nit: In an ordinary `jp c rm <id>` run, the irreversible-removal warning is visible only after pressing `?`. `InlineSelect` does not display option descriptions in its initial prompt, so the user sees `Remove this conversation?` without the warning that deletion cannot be undone. Keep that warning in the always-visible question or preamble; the help text can repeat it.

crates/jp_cli/src/format/conversation.rs
```diff
@@ -32,8 +37,11 @@ pub struct DetailsFmt {
     /// If `None`, the details are not shown.
     pub local: Option<bool>,
 
-    /// Mark the active conversation.
-    pub active_conversation: Option<ConversationId>,
+    /// Whether this is the session's active conversation.
+    ///
+    /// `None` leaves it unqueried, which omits the `Last Activated` row and the
+    /// `active` JSON key.
```
JeanMertz: nit: `DetailsFmt::new(id).json()` emits `"active": null`; it does not omit the key when `active` is `None`. The fixed-key contract is also documented on `json()`. Describe `None` as omitting the display row but serializing to `null`, so callers do not infer a different JSON shape from this field's documentation.

JeanMertz: Correct, and the doc contradicts the contract two screens down: `json()` documents "unqueried flags as `null` — so consumers can rely on key presence", and `"active": self.active` serializes `None` to `null` with the key intact. Only the row half was right.

Fixed in 35a42b6f, matching the phrasing the sibling `pinned` / `local` fields use:

```rust
/// Whether this is the session's active conversation.
/// If `None`, the `Last Activated` row is not shown and the `active` JSON
/// key is `null`.
pub active: Option<bool>,
```

crates/jp_cli/src/shared/confirm.rs
JeanMertz: Confirmed, and worse than a nit: this is a regression, on the one command where it costs something. `InlineSelect` passes `help_message: None` to `CustomType` and only writes `help_text` on `'?'` or an invalid key, so an ordinary `jp c rm <id>` showed `Remove this conversation? [y,n,?]` and nothing else. Before this PR `rm` used `Confirm::with_help_message("this action cannot be undone")`, which `inquire` documents as "Message displayed at the line below the prompt" — always rendered. And the warning is accurate: `remove_conversation_with_lock` reaches `fs::remove_dir_all` over both the workspace and user-local roots (`jp_storage/src/lib.rs:406-416, 841-852`), with no trash involved.

Rather than smuggle the line into the preamble, I took your suggestion further and gave `InlineSelect` the capability, since it was one field away from having it. Fixed in 35a42b6f:

```rust
pub fn with_help_message(mut self, help_message: impl Into<String>) -> Self
```

It fills the help line `inquire` renders below the prompt, which is a separate channel from the `?` listing: the listing describes each option to a user who asks, the help line reaches the user who answers straight away. `ConversationAction::caution()` returns `Some("this action cannot be undone")` for `Remove` and `None` for `Archive`, so the reversible action stays uncluttered. `accept_help` drops back to a plain `"yes, remove it"` now that the warning is not hiding there.

Two tests: one in `jp_inquire` pinning that the help message is independent of the `?` listing, one in `shared::confirm` that only the irreversible action carries a caution.

PR #1126, opened 2026-09-08
fix(config, tool): Keep tool access rule order across a config delta

JeanMertz:
Prepending a rule to `access.fs` or `access.env`, inserting one before
the last, or reordering the list, produced a delta that reproduced the
set of rules and not their sequence. Rules of equal specificity break
toward the one declared last, so a `read = false` rule that should have
won lost, and a reorder was recorded as no change at all.

`rule_delta` decided between appending and replacing by asking whether
every previous rule still appeared somewhere in the new list. That is
true of any permutation. Appending reaches the new list only when that
list starts with the old one, which is what it now asks; every other
difference carries the whole list with `replace`.

The delta law is what a delta has to satisfy, so it is now a test:
folding `delta(prev, next)` onto `prev` must produce `next`, checked
across the nine ways one resolved snapshot can differ from another.
Three of them failed before this change. The suite is written against
the shape both sides of a delta arrive in, a resolved list stamped
`replace`, so it exercises the path the producer actually takes.

### DIFF
--- a/crates/jp_config/src/conversation/tool/access.rs
+++ b/crates/jp_config/src/conversation/tool/access.rs
@@ -127,11 +127,14 @@ impl PartialConfigDelta for PartialAccessConfig {
 
 /// Diff two rule lists into a delta that replays to `next`.
 ///
-/// An append-shaped delta can only add, so a rule that disappeared between
-/// `prev` and `next` would come back when the delta is folded over `prev`
-/// again.
-/// When anything is missing from `next`, the delta therefore carries the whole
-/// list with `replace`; otherwise it carries just the new rules and appends.
+/// An append-shaped delta can only add to the end, so it reaches `next` exactly
+/// when `next` starts with `prev`, and the delta is then the tail.
+/// Every other difference (a rule removed, reordered, or inserted before the
+/// last one) has the delta carry the whole list with `replace`.
+///
+/// Order is part of the answer, not a detail: rules of equal specificity break
+/// toward the one declared last, so a delta that reproduced the set of rules
+/// while appending them in a different order would invert which one wins.
 ///
 /// `next` comes from a fully resolved config, so it is the complete rule set
 /// and replacing with it loses nothing.
@@ -139,11 +142,8 @@ fn rule_delta<T: Clone + PartialEq>(
     prev: &MergeableVec<T>,
     next: MergeableVec<T>,
 ) -> MergeableVec<T> {
-    if prev.iter().all(|rule| next.contains(rule)) {
-        return next
-            .into_iter()
-            .filter(|rule| !prev.contains(rule))
-            .collect();
+    if next.starts_with(prev) {
+        return next.iter().skip(prev.len()).cloned().collect();
     }
 
     MergeableVec::Merged(MergedVec {
--- a/crates/jp_config/src/conversation/tool/access.rs
+++ b/crates/jp_config/src/conversation/tool/access.rs
@@ -127,10 +127,11 @@ impl PartialConfigDelta for PartialAccessConfig {
 
 /// Diff two rule lists into a delta that replays to `next`.
 ///
-/// An append-shaped delta can only add to the end, so it reaches `next` exactly
-/// when `next` starts with `prev`, and the delta is then the tail.
-/// Every other difference (a rule removed, reordered, or inserted before the
-/// last one) has the delta carry the whole list with `replace`.
+/// An append-shaped delta can only add to the end, and appending deduplicates,
+/// so it reaches `next` exactly when `next` starts with `prev` and repeats no
+/// rule; the delta is then the tail.
+/// Every other difference (a rule removed, reordered, inserted before the last
+/// one, or repeated) has the delta carry the whole list with `replace`.
 ///
 /// Order is part of the answer, not a detail: rules of equal specificity break
 /// toward the one declared last, so a delta that reproduced the set of rules
@@ -142,7 +143,7 @@ fn rule_delta<T: Clone + PartialEq>(
     prev: &MergeableVec<T>,
     next: MergeableVec<T>,
 ) -> MergeableVec<T> {
-    if next.starts_with(prev) {
+    if next.starts_with(prev) && !repeats_a_rule(&next) {
         return next.iter().skip(prev.len()).cloned().collect();
     }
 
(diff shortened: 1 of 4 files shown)


crates/jp_config/src/delta_law_tests.rs
```diff
@@ -0,0 +1,123 @@
...
+/// silently inverts precedence.
+fn assert_law(before: &[&str], after: &[&str]) {
+    let prev = snapshot(before);
+    let next = snapshot(after);
+
+    let delta = prev.delta(next.clone());
+
+    let mut folded = prev;
+    folded
+        .merge(&(), delta)
```
JeanMertz: blocking: On a continuing conversation with `access.fs = [{path = "src", read = true}]`, using `--cfg` to reorder only `access.env` also clears the filesystem policy when the change is persisted.

`get_config_delta_from_cli` computes `delta_with_unsets`, so the unchanged `fs` becomes an empty append delta. It then passes that delta to `ConversationStream::add_config_delta`, which calls `config.to_partial().delta(*delta)` again when `unsets` is empty. On that second pass, `rule_delta` interprets the empty `fs` delta as a complete empty snapshot and emits `Replace([])`. The stored access block still contains env rules, but an empty filesystem policy means unrestricted workspace access.

Please distinguish already-computed deltas from snapshots at the persistence boundary, and add a regression through the query's delta calculation and `add_config_delta`. The single delta-and-merge cycle here misses this persistent widening of access.

crates/jp_config/src/conversation/tool/access.rs
```diff
@@ -127,23 +127,23 @@ impl PartialConfigDelta for PartialAccessConfig {
...
     prev: &MergeableVec<T>,
     next: MergeableVec<T>,
 ) -> MergeableVec<T> {
-    if prev.iter().all(|rule| next.contains(rule)) {
-        return next
-            .into_iter()
-            .filter(|rule| !prev.contains(rule))
-            .collect();
+    if next.starts_with(prev) {
+        return next.iter().skip(prev.len()).cloned().collect();
```
JeanMertz: blocking: A `--cfg` replacement list can repeat a rule to override an intervening rule. For example, for `GITHUB_TOKEN`, let `prev = [deny, allow]` and `next = [deny, allow, deny]`, with the final deny identical to the first. Replacement lists preserve duplicates, so this is a valid resolved snapshot and its final rule must deny access.

`next.starts_with(prev)` is true, but collecting the tail produces a plain `MergeableVec::Vec`. `vec_with_strategy` deduplicates append merges by default, keeping the first occurrence. Folding `[deny]` onto `[deny, allow]` therefore yields `[deny, allow]`, silently retaining permission to read the token.

Use replacement when append deduplication would change the target sequence, and add a repeated-rule case to the law tests. `assert_law(&["A", "B"], &["A", "B", "A"])` exposes the sequence loss; a same-name allow/deny case would also verify the permission outcome.

JeanMertz: Taking the guard, with two corrections to the framing and a wider condition than proposed.

The mechanism holds. The tail collects into `MergeableVec::Vec`, which gives `strategy = None` in `vec_with_strategy`, so `dedup_active = dedup.unwrap_or(true)` is `true` and `dedup_in_place` keeps the first occurrence. `[deny, allow] + [deny]` folds to `[deny, allow]`.

Not a regression, though: the previous body lost the same case. `prev.iter().all(|r| next.contains(r))` holds for `[deny, allow] -> [deny, allow, deny]`, and the filter drops the third rule because it is already in `prev`.

And the input is narrower than "a `--cfg` replacement list". A plain `--cfg` list or a `[[...access.env]]` sequence merges with `strategy = None`, where dedup is on and the repeat is stripped during resolution — the duplicate only reaches a resolved snapshot through an explicit `strategy = "replace"` or `dedup = false`, and only changes the outcome when a rule of identical specificity and opposite verdict sits between the two copies. So it is not the token-leak-on-a-normal-Tuesday this reads as. What earns the fix is that the law is the claim this PR makes, and it has a counterexample.

The proposed condition is too narrow in one direction. `dedup_in_place` scans the whole merged vector, so it also collapses a repeat that lives entirely inside `prev`: `[A, B, A] -> [A, B, A, C]` starts with `prev`, appends cleanly, and folds to `[A, B, C]`. The guard is therefore "`next` repeats any rule", not "the tail repeats something in `prev`".

Applied:

```rust
if next.starts_with(prev) && !repeats_a_rule(&next) {
    return next.iter().skip(prev.len()).cloned().collect();
}
```

with `repeats_a_rule` checking each element against its prefix, plus the doc comment naming dedup as the second condition. Two law cases: `assert_law(&["A", "B"], &["A", "B", "A"])` (yours) and `assert_law(&["A", "B", "A"], &["A", "B", "A", "C"])` (the `prev`-side repeat). Both were red without the guard, with `["A", "B"]` and `["A", "B", "C"]` respectively, and green with it.

Skipped the same-name allow/deny variant: the sequence assertion already pins what the fold loses, and which rule wins a tie is `jp_tool`'s test to own (`access.rs:992` covers it).

Worth noting for the rest of the stack: #1129 lifts this function into `delta::delta_mergeable_vec` verbatim, so the guard needs to come along in that rebase rather than being reintroduced there.

crates/jp_config/src/delta_law_tests.rs
JeanMertz: Correct in every step, and already fixed in #1135, which is ahead of this one.

Verified the chain: `PartialAccessConfig` implements only `delta`, so it inherits the default `delta_with_unsets` and reports no paths; with `unsets` empty, `add_config_delta` re-diffs, `[].starts_with(&[rule])` is false, and `fs` resolves to `Replace([])`.

Not something this PR can fix, and not something it introduced. An unchanged list returned `Vec([])` under the previous body too (`prev.iter().all(|r| next.contains(r))` holds when the lists are equal), and the second pass read it the same way. The common trigger is a plain `--cfg` append of one env rule, which predates the ordering fix; a reorder is one more way in. An empty `MergeableVec` cannot distinguish "nothing to add" from "cleared", so no change to `rule_delta` settles it — the two passes are entitled to opposite readings of the same value.

#1135 takes the root fix you describe: the delta is recorded as computed, and the second diff is gone. It carries the regression coverage through `config_pipeline` and `add_config_delta`, and files `0ffsw7e` alongside it.

PR #1128, opened 2026-09-08
docs(rfd): Keep conversations pinned across the 070 reshaping

JeanMertz:
A conversation resolves the configuration it was created with, and
neither a binary upgrade nor a later workspace edit changes it. 070
reshaped `base_config.json` so a `-C` can match what the creating
invocation claimed, without accounting for the pinning that same field
was carrying.

`base` stays a resolved `AppConfig`. What narrows is its input: only the
workspace files, with user-global and user-workspace config joining the
environment, `-c` args and CLI flags in `init`, where each carries a
claim and is revertible. A conversation stream is committed to the
repository, so nothing machine-local belongs in the half a colleague
reads.

Phase 2b requires a workspace to resolve from its own files, which is
what makes a cloned repository work without the machine that made it,
and it is the prerequisite Phase 3 builds on.

### DIFF
--- a/docs/rfd/070-negative-config-deltas.md
+++ b/docs/rfd/070-negative-config-deltas.md
@@ -835,8 +835,8 @@ It is to keep the resolved type and narrow *what* is resolved into it.
   User-global config, user-workspace config, environment variables, `-c` args
   and CLI flags are all excluded.
 - **`init`** is an ordered list of `ConfigDelta` entries, one per source the
-  creating invocation layered on top of the workspace files: user-global
-  config, user-workspace config, the environment, each `-c` directive (file,
+  creating invocation layered on top of the workspace files: user-global config,
+  user-workspace config, the environment, each `-c` directive (file,
   `key=value`, or JSON object), each `-C` directive, and the trailing
   shortcut-flags batch.
   Each entry carries its diff, claims map, and any `unsets`.
@@ -847,8 +847,8 @@ The stored type is `AppConfig`; the stored *encoding* is what it is today.
 `base` appears on disk as a partial-shaped object and is finalized on load.
 That path is retained as-is, including the schema-aware stripping of unknown
 fields and the fallback for a stored config that has lost a required field.
-Nothing about this RFD changes how `base` is encoded, only what is resolved
-into it.
+Nothing about this RFD changes how `base` is encoded, only what is resolved into
+it.
 
 ##### What an `init` entry holds
 
@@ -863,8 +863,8 @@ A workspace naming `assistant.model.id = "coder"` with
 `providers.llm.aliases.coder = "openai/gpt-4o"`, plus a user-workspace override
 setting only that alias to `anthropic/claude-opus-4-6`, resolves to the
 Anthropic model — but `base` already holds the flattened `openai/gpt-4o`, and
-`ModelIdOrAliasConfig::resolve_in_place` is a no-op once the field is a
-concrete `Id`.
+`ModelIdOrAliasConfig::resolve_in_place` is a no-op once the field is a concrete
+`Id`.
 Replaying the alias map alone would leave the conversation on the wrong model.
 A difference between resolutions carries the resolved `assistant.model.id` too,
 so the replay lands where ordinary resolution lands.
@@ -960,12 +960,12 @@ RFD 038 persists `--cfg=WORKSPACE` as a `Reset` followed by an `Apply` of the
 workspace state, and `fold_config_delta` implements `Reset` by returning the
 accumulated state to the empty partial.
 A `Reset` inside `init` would therefore discard `base` during replay, and every
-field the reset's own `Apply` does not restore — `assistant.request.max_retries`
-and every other `#[setting(default)]` — would come from whichever binary reads
-the conversation next.
+field the reset's own `Apply` does not restore —
+`assistant.request.max_retries` and every other `#[setting(default)]` — would
+come from whichever binary reads the conversation next.
 That is precisely the pinning break this section exists to prevent.
-Creation keeps absorbing reset directives into `base`, and `init` holds only
-the sources that follow the last reset in the list.
+Creation keeps absorbing reset directives into `base`, and `init` holds only the
+sources that follow the last reset in the list.
 
 The cost in both cases is provenance, not pinning: `-C` cannot reach a
 creation-time reset, or anything the reset discarded, or any contribution to a
@@ -1033,8 +1033,8 @@ transparently:
 2. If the root is a JSON object with a `base` key → new format; read `base` as
    the stored config and `init` (defaulting to `[]` if absent) as the
    creation-time delta list.
-3. Otherwise the root object *is* the stored config; wrap it as `{ base:
-   <that>, init: [] }`.
+3. Otherwise the root object *is* the stored config; wrap it as `{ base: <that>,
+   init: [] }`.
    Legacy conversations therefore have an empty `init` list, matching their
    historical "no claims" behavior.
    `-C` is a no-op on their fields.
@@ -2004,8 +2004,8 @@ Phase 2b already rejects a workspace whose files do not resolve on their own, so
 this phase can rely on `base` being valid.
 
 An invocation carrying a reset keyword, or `NONE`, keeps today's behavior:
-creation absorbs those contributions into `base` and writes an empty `init`
-(see [Creation-time resets, and `NONE`](#creation-time-resets-and-none)).
+creation absorbs those contributions into `base` and writes an empty `init` (see
+[Creation-time resets, and `NONE`](#creation-time-resets-and-none)).
 
 **Storage layer** (`jp_storage`):
 
--- a/docs/rfd/070-negative-config-deltas.md
+++ b/docs/rfd/070-negative-config-deltas.md
@@ -819,26 +819,70 @@ It is to keep the resolved type and narrow *what* is resolved into it.
 
 ```json
 {
-  "base": { /* AppConfig — the resolved workspace configuration */ },
+  "base": { /* the workspace configuration, partial-encoded */ },
   "init": [
-    { /* ConfigDelta — first creation-time directive */ },
-    { /* ConfigDelta — next directive */ }
+    { /* ConfigDelta — first creation-time source */ },
+    { /* ConfigDelta — next source */ }
   ]
 }
 ```
 
-- **`base`** is a resolved `AppConfig`, as today.
-  What changes is its input: the workspace's own config files merged via
-  inheritance — `.jp/config.toml` and the `.jp.toml` chain — resolved on their
+- **`base`** is an `AppConfig`, as today: total, valid, and with program
+  defaults baked into concrete values.
+  What changes is its input — the workspace's own config files merged via
+  inheritance (`.jp/config.toml` and the `.jp.toml` chain), resolved on their
   own.
   User-global config, user-workspace config, environment variables, `-c` args
   and CLI flags are all excluded.
-  It stays the total, valid, defaults-baked type the conversation has always
-  stored; it simply stops absorbing anything that is not workspace state.
-- **`init`** is an ordered list of `ConfigDelta` entries, one per directive
-  (`-c` file, `-c key=value`, `-c '{...}'`, `-C ...`, or the trailing
-  shortcut-flags batch) from the invocation that created the conversation.
+- **`init`** is an ordered list of `ConfigDelta` entries, one per source the
+  creating invocation layered on top of the workspace files: user-global
+  config, user-workspace config, the environment, each `-c` directive (file,
+  `key=value`, or JSON object), each `-C` directive, and the trailing
+  shortcut-flags batch.
   Each entry carries its diff, claims map, and any `unsets`.
+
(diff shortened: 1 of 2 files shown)


docs/rfd/070-negative-config-deltas.md
```diff
@@ -1743,24 +1828,78 @@ Tests:
...
+  deltas.
 - Update `Iter`, `IterMut`, and `IntoIter` to fold `base` and each `init` delta
   before walking events, so per-event views match `config()`.
 - `from_parts()` / `to_parts()` keep their two-component signatures (base_config
   JSON value, events JSON vec); the JSON shape of the base_config component
   changes to `{ base, init }`.
 
+**Conversation creation** resolves `base` from the workspace's own config files
+and records user-global config, user-workspace config, environment variables,
+`-c` args and flags as `init` deltas, rather than resolving all of them together
```
JeanMertz: blocking: With `conversation.tools.'*'.run = "unattended"` in user-global config and `run = "ask"` in the workspace, the existing load order makes the workspace's `ask` win (`load_partial_configs_from_files` loads user-global first). Folding a workspace-only `base` followed by the user-global contribution in `init` reverses that precedence and silently changes tool execution policy.

Please specify how the user-global delta and its claims preserve its lower precedence, including when a workspace field shadows it. If changing precedence is intentional, that needs an explicit compatibility decision rather than being implicit in the storage reshaping. Add a conflicting user-global/workspace case to Phase 3's tests.

docs/rfd/070-negative-config-deltas.md
```diff
@@ -788,23 +788,53 @@ the first invocation's per-directive deltas, preserving the `events.json`
...
 
-- **`base`** is the workspace `PartialAppConfig` — files merged via
-  inheritance, without environment variables, `-c` args, or CLI flags.
-  Same content as today's `base_config.json`; unchanged shape for hand-editing.
+- **`base`** is a resolved `AppConfig`, as today.
+  What changes is its input: the workspace's own config files merged via
+  inheritance — `.jp/config.toml` and the `.jp.toml` chain — resolved on their
+  own.
+  User-global config, user-workspace config, environment variables, `-c` args
+  and CLI flags are all excluded.
```
JeanMertz: blocking: Consider a valid workspace with `assistant.model.id = "coder"` and `providers.llm.aliases.coder = "openai/gpt-4o"`, plus a user-workspace override setting only that alias to `anthropic/claude-opus-4-6`. Resolving the workspace first bakes `openai/gpt-4o` into `base.assistant.model.id`. Replaying the alias-map delta cannot update that concrete ID: `ModelIdOrAliasConfig::resolve_in_place` is a no-op for `Id`. The conversation silently uses a different model from the ordinary merged invocation.

Please define how `init` captures changes to resolved values caused by later layers, not just the fields those layers explicitly set. Add a test comparing the initial replay result with resolution of the full invocation, including an alias-only user-workspace override.

docs/rfd/070-negative-config-deltas.md
```diff
@@ -1743,24 +1828,78 @@ Tests:
...
+
 ### Phase 3: Conversation creation and `base_config.json` shape change
 
+This phase is the one that can break the pinning rule (see [Pinning is a
+constraint on the reshaping](#pinning-is-a-constraint-on-the-reshaping)), so its
+tests carry a case that fails loudly if it does: resolve a conversation, change
+a `#[setting(default)]` value, resolve it again, and assert the result is
+unchanged.
+That test is why `base` stays a resolved `AppConfig`, and it belongs in the same
+commit as the shape change rather than after it.
```
JeanMertz: blocking: A new conversation created with `-c WORKSPACE` in an otherwise valid workspace can still lose default pinning. RFD 038 persists this directive as `Reset` plus `Apply(workspace partial)`. `fold_config_delta` resets to an empty partial, and that workspace partial need not contain `assistant.request.max_retries`. If this sequence goes into `init`, replay discards the defaults baked into `base` and picks the retry default from the reading binary. Creation currently absorbs the directive into the resolved snapshot.

Please define how creation-time reset directives preserve invocation-time defaults, and include a serialize/reload regression case with a reset in `init`. A test without reset directives would not exercise this loss of the resolved base.

docs/rfd/070-negative-config-deltas.md
```diff
@@ -1743,24 +1828,78 @@ Tests:
...
+A prerequisite for Phase 3, shipped separately because it rejects an invocation
+that succeeds today (see [The workspace has to be
+self-contained](#the-workspace-has-to-be-self-contained)).
+
+Conversation creation resolves the workspace's own config files —
+`.jp/config.toml` and the `.jp.toml` chain, merged via inheritance — and fails
+when they do not produce a valid `AppConfig`.
+User-global config, user-workspace config, environment variables, `-c` arguments
+and CLI flags are all excluded from that resolution, since none of them travel
+with the repository.
```
JeanMertz: question: What should `jp query --new -c NONE -c ./mre.toml hello` do when `.jp/config.toml` is broken or lacks a model, but `mre.toml` is complete? RFD 038 explicitly supports this escape hatch, and `ConfigPipeline::new` skips implicit file loading under `NONE`. Requiring a valid workspace-only `base` would force that invocation to load the config it opted out of and then fail.

Please specify the `NONE` case here, including what supplies its stored base. If the escape hatch is deliberately being restricted, document that compatibility change alongside Phase 2b rather than leaving the two RFDs with conflicting contracts.

docs/rfd/070-negative-config-deltas.md
```diff
@@ -867,14 +942,20 @@ transparently:
...
 1. Parse `base_config.json` as JSON.
 2. If the root is a JSON object with a `base` key → new format; use `base` as
-   the workspace partial and `init` (defaulting to `[]` if absent) as the
-   creation-time delta list.
-3. Otherwise treat the root object as a legacy `PartialAppConfig` and wrap it as
-   `{ base: <that>, init: [] }`.
+   the resolved workspace config and `init` (defaulting to `[]` if absent) as
+   the creation-time delta list.
+3. Otherwise treat the root object as a legacy `AppConfig` and wrap it as `{
+   base: <that>, init: [] }`.
```
JeanMertz: question: Does `AppConfig` here mean the in-memory invariant, or a new JSON encoding? `ConversationStream::to_parts()` currently serializes `base_config.to_partial()`, and `from_parts()` deserializes through `compat::deserialize_partial_config` before finalizing. The legacy file therefore contains the partial encoding of a resolved snapshot, not a directly serialized `AppConfig`.

Please distinguish the in-memory type from the wire format and say whether that existing compatibility path is retained. This also needs to agree with the remaining `PartialAppConfig` descriptions in User-facing commands and Phase 3's Storage layer; otherwise an implementer cannot tell whether wrapping the legacy JSON really is sufficient.

docs/rfd/070-negative-config-deltas.md
JeanMertz: Correct, and the fix generalises past aliases.

Verified the mechanism: `AppConfig::resolve_aliases` flattens the map and rewrites `assistant.model.id` to a concrete `Id`, and `resolve_in_place` documents itself as a no-op from there (`model/id.rs:196-204`). A `base` resolved from workspace files alone holds `openai/gpt-4o`, and an `init` entry carrying only `providers.llm.aliases.coder` cannot move it.

The root error was treating an `init` entry as "the partial that source declared". It has to be what the source *contributed*: the difference between resolving sources up to and including it and resolving the sources before it, in load order. Then the alias-map override carries the resolved `assistant.model.id` too, and the replay lands where ordinary resolution lands. Alias flattening is the clearest case; inquiry inheritance and instruction ordering are computed during resolution the same way.

Added a **What an `init` entry holds** subsection with that definition, and the invariant it buys stated outright: folding `base` then every `init` entry produces the same `AppConfig` as resolving the whole invocation at once. Phase 3's tests gained that as a case, with your alias-only user-workspace override as the fixture.

Same fix resolves your user-global precedence comment, which is the second half of the same mistake.

docs/rfd/070-negative-config-deltas.md
JeanMertz: In-memory invariant, and I wrote it as though it were the wire format. Fixed.

Confirmed the encoding: `to_parts()` writes `to_value(self.base_config.to_partial())` (`stream.rs:2000-2002`), and `from_parts()` reads through `compat::deserialize_partial_config` then `finalize_recovered_config` (`:1972-1988`). So the file has always held a partial-shaped object, and read literally my text would have had an implementer serialize `AppConfig` directly — losing the schema-aware stripping and the required-field fallback that #1110 added.

The amendment now says it explicitly: the stored type is `AppConfig`, total and defaults-baked; the stored encoding is unchanged, and that compat path is retained as-is. Only what gets resolved into `base` narrows.

Both contradictions you named are real and corrected — User-facing commands ("the familiar `PartialAppConfig`") and Phase 3's Storage layer (`{ base: PartialAppConfig, … }`), plus two more of mine in the same vein: "plain `AppConfig` under `base`" in the RFD 054 paragraph, and the legacy-wrapping claim that the file "already holds a resolved `AppConfig`". The legacy paragraph now gives the actual reason wrapping suffices: both shapes hold the same partial-encoded snapshot, read through the same path.

docs/rfd/070-negative-config-deltas.md
JeanMertz: Real conflict, and worse than unspecified.

`ConfigPipeline::new:223-227` starts from `PartialAppConfig::default()` when `NONE` is present and never calls `load_base`, with the doc noting the gate is decided before any file I/O precisely to keep the escape hatch working when implicit config is broken. Phase 2b as written made creation load those files and fail — the invocation `NONE` exists to rescue.

And there is no valid workspace-only `base` to store for it: no files, and `ModelIdConfig` carries `#[setting(required)]` on both `provider` and `name`, so a defaults-only resolution does not finalize. The model arrives from `-c ./mre.toml`, which the amendment had put in `init`.

Specified as: under `NONE`, creation resolves `base` from the whole invocation — today's behaviour — and `init` is empty. Phase 2b's check does not apply, since the user has said in the invocation itself that this conversation is not built on workspace config. `-C` is a no-op on those fields, exactly like a legacy conversation. Phase 2b gained a test that `jp query --new --cfg=NONE --cfg=./mre.toml hello` still creates a conversation in a workspace whose own files are broken.

Worth noting what I did *not* do: weaken the invariant to "`base` + `init` resolves". Holding `base`'s validity fixed is what forces the fallback, and it keeps the pinning guarantee uniform across every conversation.

docs/rfd/070-negative-config-deltas.md
JeanMertz: Confirmed in full, and it breaks the very rule this amendment exists to protect.

`persist_config_reset` writes a `ResetDelta` followed by `Apply(workspace)` and `Apply(post)`, and `fold_config_delta:408-411` implements `Reset` as `*state = PartialAppConfig::default()`. Today a fresh conversation never stores that sequence — `query.rs` guards the reset path with `if !fresh` and absorbs the keyword into the resolved snapshot. Moving creation-time directives into `init` puts the `Reset` in the replay path, so `base` is discarded and every field the reset's own `Apply` does not restore takes its default from whichever binary reads next.

Fix keeps `Reset` semantics untouched: creation continues to absorb reset directives into `base`, and `init` holds only the sources following the last reset. The cost is provenance, not pinning — `-C` cannot reach a creation-time reset or anything it discarded, which costs nothing real, since a reset's floor is `base` and `base` is what the reset resolved to. Written up in the new **Creation-time resets, and `NONE`** section alongside the `NONE` case, which lands in the same place for a different reason.

I considered redefining `Reset`'s floor as `base` instead. Rejected: that changes what mid-conversation `--cfg=NONE` means for shipped conversations, which is an RFD 038 decision and not one to make implicitly here.

Your point about the test is the one that stung — the pinning test I specified has no reset in `init`, so it would have passed while the guarantee was broken. Phase 3 now carries the case twice: an ordinary conversation, and one created with a reset directive.

docs/rfd/070-negative-config-deltas.md
JeanMertz: Correct. Verified the order: `load_partial_configs_from_files:1148-1193` pushes user-global, then workspace `.jp/config.toml`, then the `.jp.toml` chain, then user-workspace, and `load_partials_with_inheritance` merges in that order with later layers overriding. So user-global is the *lowest*-precedence file source and the workspace overrides it. Recording what user-global declared and replaying it after `base` inverts that, and `run = "ask"` silently becomes `unattended`.

Not an intentional precedence change — an oversight. The amendment reasoned about *where* each source belongs (workspace state vs user intent) and never checked what folding order that implied.

Fixed by the same contribution-based definition as your alias comment: an `init` entry carries the difference between successive resolutions in load order, so a field the workspace overrides contributes nothing from user-global and the replay order cannot change the outcome. Both cases are written up in the new **What an `init` entry holds** subsection.

One asymmetry worth recording, since it explains why only user-global bites: user-workspace config is already the last file loaded, so moving it into `init` preserves its precedence. Only user-global inverts.

Phase 3's tests gained your conflicting case: `unattended` in user-global against `ask` in the workspace resolves to `ask`, both through ordinary resolution and through the `base` + `init` replay.

PR #1129, opened 2026-09-08
feat(config): Record a field the user cleared

JeanMertz:
A delta can say that a field went away. Clearing a field with `--cfg
assistant.name=null` reached the runtime config, because `--cfg` assigns
directly rather than merging, but no delta could carry it: schematic keeps
the previous value when the next layer has none, so `delta_opt` dropped the
difference.

Such a field now reports its path in the delta's `unsets`, and the fold
clears it before merging, so the conversation's own resolution honors it.
Scalars and the lists that carry their own merge strategy are both covered.

A sweep over every field the schema names asserts the law directly: clear it,
resolve, take the delta, fold it back, and the field is still clear. It
populates its own fixture, since a field left unset by the fixture is how the
first version of this quietly covered nothing. Ten paths are listed as known
gaps with a reason each. The three fields a delta never stores are named in
one place rather than restated in each impl.

The invocation pipeline is not covered by this. A conversation layer fills
from the file layer, and filling reads a cleared field as one the conversation
does not mention, so the next `jp query` refills it from the workspace
config. Tracked in T-0gqh7z7.

### DIFF
--- a/crates/jp_config/src/assistant.rs
+++ b/crates/jp_config/src/assistant.rs
@@ -20,7 +20,9 @@ use crate::{
         sections::{PartialSectionConfig, SectionConfig},
         tool_choice::ToolChoice,
     },
-    delta::{PartialConfigDelta, delta_mergeable_vec, delta_opt, delta_opt_partial, path},
+    delta::{
+        PartialConfigDelta, delta_mergeable_vec, delta_opt, delta_opt_at, delta_opt_partial, path,
+    },
     fill::{FillDefaults, fill_opt},
     internal::merge::{string_with_strategy, vec_with_strategy},
     model::{ModelConfig, PartialModelConfig},
@@ -140,7 +142,7 @@ impl PartialConfigDelta for PartialAssistantConfig {
 
     fn delta_with_unsets(&self, next: Self, prefix: &str, unsets: &mut Vec<String>) -> Self {
         Self {
-            name: delta_opt(self.name.as_ref(), next.name),
+            name: delta_opt_at(&path(prefix, "name"), self.name.as_ref(), next.name, unsets),
             system_prompt: delta_opt_partial(self.system_prompt.as_ref(), next.system_prompt),
             instructions: delta_mergeable_vec(&self.instructions, next.instructions),
             system_prompt_sections: delta_mergeable_vec(
@@ -151,7 +153,9 @@ impl PartialConfigDelta for PartialAssistantConfig {
             model: self
                 .model
                 .delta_with_unsets(next.model, &path(prefix, "model"), unsets),
-            request: self.request.delta(next.request),
+            request: self
+                .request
+                .delta_with_unsets(next.request, &path(prefix, "request"), unsets),
         }
     }
 }
--- a/crates/jp_config/src/assistant.rs
+++ b/crates/jp_config/src/assistant.rs
@@ -20,7 +20,7 @@ use crate::{
         sections::{PartialSectionConfig, SectionConfig},
         tool_choice::ToolChoice,
     },
-    delta::{PartialConfigDelta, delta_opt, delta_opt_partial, path},
+    delta::{PartialConfigDelta, delta_mergeable_vec, delta_opt, delta_opt_partial, path},
     fill::{FillDefaults, fill_opt},
     internal::merge::{string_with_strategy, vec_with_strategy},
     model::{ModelConfig, PartialModelConfig},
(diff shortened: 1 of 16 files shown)


crates/jp_config/src/assistant.rs
```diff
@@ -127,17 +129,11 @@ impl PartialConfigDelta for PartialAssistantConfig {
...
-            system_prompt_sections: next
-                .system_prompt_sections
-                .into_iter()
-                .filter(|v| !self.system_prompt_sections.contains(v))
-                .collect(),
+            instructions: delta_mergeable_vec(&self.instructions, next.instructions),
+            system_prompt_sections: delta_mergeable_vec(
+                &self.system_prompt_sections,
+                next.system_prompt_sections,
+            ),
```
JeanMertz: blocking: On a conversation with nonempty instructions, changing only `style.code.color` can erase those instructions from the stored config. `get_config_delta_from_cli` produces a sparse delta with empty instruction lists and no unsets. `ConversationStream::add_config_delta` then calls `config.to_partial().delta(*delta)` again (`stream.rs:591`). These calls interpret the omitted lists as complete empty snapshots and emit `replace: []`. The same problem affects `conversation.attachments`; appending an instruction also turns the first diff's tail into a replacement during the second diff.

Separate full-snapshot diffing from sparse-delta suppression, preserving the distinction between an omitted list and an explicit empty replacement. Add a regression through `get_config_delta_from_cli` and `add_config_delta` that changes a scalar and asserts the existing lists survive. The direct merge in the law test bypasses this path.

crates/jp_config/src/delta_law_tests.rs
```diff
@@ -77,6 +79,185 @@ fn assert_law(before: &[&str], after: &[&str]) {
...
+        let mut unsets = Vec::new();
+        let delta = prev.delta_with_unsets(next, "", &mut unsets);
+
+        let mut folded = prev.clone();
+        for cleared in &unsets {
+            folded.unset(cleared).expect("a reported path is a field");
+        }
+        folded.merge(&(), delta).expect("folding cannot fail");
+
+        if crate::util::build(folded).ok().as_ref() != Some(&expected)
```
JeanMertz: blocking: With `assistant.name = "Bot"` in workspace config, query once with `--cfg 'assistant.name:=null'`, then resume without the flag. The stored stream folds to `None`, but `Query::apply_conversation_config` converts that resolved config back to a partial, and `ConfigPipeline::partial_with_conversation` calls `conversation.fill_from(self.base.clone())`. `PartialAssistantConfig::fill_from` restores `Some("Bot")` from the file layer. The next query therefore silently restores the value the user cleared.

This assertion stops before the configuration pipeline that the next invocation actually runs. Preserve explicit clears across that boundary so they are not treated as missing configuration, and add a two-invocation regression with a nonempty file-layer value.

crates/jp_config/src/assistant.rs
JeanMertz: The second diff this depends on is gone, so the erasure does not happen.

`add_config_delta` at this head (`stream.rs:613-624`) records the delta as given and only drops an apply that merges nothing and clears nothing. Grepping `to_partial().delta(*delta)` across `jp_conversation` and `jp_cli` finds nothing. Line 591, cited here as the call site, is now inside the doc comment explaining why that call must not exist: "A delta is not a snapshot [...] an appended list element becomes the whole list, and a field left alone becomes a field emptied."

That came from #1135 (`7d9d1daf`), which is in this PR's base — `stream.rs` is not among the changed files, so the version I read is the base's. It closed `0ffsw7e`, filed for exactly this failure on `access.fs` and `access.env`.

Traced with one diff: unchanged instructions give `next.starts_with(prev)`, so the delta is the empty tail `MergeableVec::Vec([])`. Folded through `vec_with_strategy` that is an append of nothing — `[i] ++ []`, dedup leaves `[i]`. The list survives. Same for `conversation.attachments`, and for the appended-instruction case, whose tail stays a tail.

The ambiguity underneath your point is real and already tracked. `0fgdq7z` is about exactly this: `MergeableVec::Vec([])` means both "this layer says nothing" and "this field is empty", and the two delta helpers are entitled to opposite readings. It records why it is not fixed yet — "T-0ffsw7e removed the only caller that read a delta as though it were a snapshot, so nothing currently reaches the ambiguity. It is unreachable, not absent" — and names #1130 and #1131 as what widens it from two rule lists to the whole config tree. That is where I would keep this, rather than changing `assistant.rs` here.

crates/jp_config/src/delta_law_tests.rs
JeanMertz: Correct, and the PR claims otherwise, so the claim is what I am changing here.

Verified the chain: `apply_conversation_config` turns the stream's resolved config into a partial (`query.rs:2392`), where the cleared field is `None`; `partial_with_conversation` calls `conversation.fill_from(self.base.clone())` (`config_pipeline.rs:322`); `PartialAssistantConfig::fill_from` is `self.name.or(defaults.name)` (`assistant.rs:166`). `config_pipeline.rs` never mentions `unset`, so the cleared paths are consumed by the fold and never cross the boundary.

One detail beyond what you describe: the reversal persists. The next `get_config_delta_from_cli` diffs the stream (`None`) against the runtime config (`Some("Bot")`), `delta_opt_at` takes its `(None, next) => next` arm, and a delta setting the field back is written. So the clear is not just ineffective on the next turn, it is undone in the stream.

And you are right about the test. The sweep does delta → unset → merge → `build`, entirely inside `jp_config`. It proves the stream resolves to `None`, which is a different claim from "the next invocation runs with it cleared".

Where I differ is scope. Distinguishing "cleared" from "not stated" is what `unsets` exists for, and a partial cannot carry it, so the fix needs `ConversationStream` to accumulate cleared paths across its deltas — dropping a path a later delta sets — and `partial_with_conversation` to apply them after filling, before `--cfg`. That is `jp_conversation` and `jp_cli`, in a PR that is `jp_config` only, and it is the same shape of change as #1135. Filed as `T-0gqh7z7` with your two-invocation regression as the verification.

What lands here is the honest claim. The PR body opens "Clearing a field in a conversation sticks", naming this exact scenario, and that is not what it delivers: a delta can state a clear, the fold applies it, and the conversation's own resolution honors it. The invocation pipeline still refills from the file layer. I am rewording it rather than leaving a merge commit that overstates the fix.

PR #1130, opened 2026-09-08
feat(config): Let every list state how it merges

JeanMertz:
A list-valued setting can say how it combines with the layer beneath
it:

    [editor]
    envs = { value = ["MY_EDITOR"], strategy = "replace" }

Six fields gain it: `editor.envs`, `assistant.model.parameters.stop_words`,
`providers.llm.anthropic.beta_headers`, `providers.mcp.*.arguments` and
`.variables`, and `config_load_paths`. A plain list merges the way its
field declares, so nothing changes for a config that does not ask.

What it buys beyond user control: a conversation can record a removed or
reordered element, which an appending merge cannot express. Four helpers
that existed only to guess at a strategy are deleted.

`arguments` and `variables` keep duplicates, since a repeated flag on a
command line is not a duplicate. Every other list deduplicates.

Reviewer note: the first commit changes the vendored `schematic` macro
so `partial_via` works on a field whose element type is not a nested
config, which is what a list of strings needs.


crates/jp_config/src/delta.rs
```diff
@@ -123,6 +95,24 @@ pub fn delta_mergeable_vec<T: Clone + PartialEq>(
...
+pub fn delta_opt_mergeable_vec<T: Clone + PartialEq>(
+    prev: Option<&MergeableVec<T>>,
+    next: Option<MergeableVec<T>>,
+) -> Option<MergeableVec<T>> {
+    let next = next?;
+    let Some(prev) = prev else {
+        return Some(next);
+    };
+
+    let delta = delta_mergeable_vec(prev, next);
```
JeanMertz: blocking: Continuing a conversation whose MCP arguments are `["serve"]` with `--cfg 'providers.mcp.bookworm.arguments+=--verbose'` persists `["--verbose"]`, dropping `serve`. `get_config_delta_from_cli` first computes the correct append delta `["--verbose"]`, but `ConversationStream::add_config_delta` then diffs that delta against the current config again. This helper interprets the tail as the target list and emits a replacement because it does not start with `["serve"]`. The saved configuration is wrong on subsequent queries.

Separate recording an already-computed delta from diffing a target configuration, so the query path computes the diff once. Add a regression covering query delta production, `add_config_delta`, and replay; the direct replacement test does not exercise this path.

crates/jp_config/src/assignment.rs
```diff
@@ -1051,6 +1034,62 @@ impl KvAssignment {
...
+            let merged =
+                serde_json::from_value(value).map_err(|error| kv_error(&self.key, error))?;
+
+            *vec = Some(merged);
+            return Ok(());
+        }
+
+        let mut elements = vec.take().map(MergeableVec::into_vec).unwrap_or_default();
+        self.try_vec(&mut elements, parser)?;
+        *vec = Some(elements.into());
```
JeanMertz: blocking: With `editor.envs = ["EDITOR"]` in user-global config, writing a workspace override via `jp config set --cfg 'editor.envs:={"value":["VISUAL"],"strategy":"replace"}' --cfg 'editor.envs+=MY_EDITOR'` loses the requested replacement. `into_vec()` discards the wrapper metadata, so the second assignment produces a plain list and `config set` saves `["VISUAL", "MY_EDITOR"]`. Reloading appends it to the global list, leaving `EDITOR` first rather than replacing it.

Mutate the wrapped value through `DerefMut` for element edits and appends so `strategy`, `dedup`, and `discard_when_merged` survive. Add a sequential-assignment regression that also merges the result over a lower layer.

crates/jp_config/src/editor.rs
```diff
@@ -62,8 +66,13 @@ pub struct EditorConfig {
...
     /// is tried).
     #[setting(
-        default = vec!["JP_EDITOR".into(), "VISUAL".into(), "EDITOR".into()],
-        merge = schematic::merge::append_vec,
+        default = MergeableVec::from(vec![
+            "JP_EDITOR".to_owned(),
+            "VISUAL".to_owned(),
+            "EDITOR".to_owned(),
+        ]),
+        partial_via = MergeableVec::<String>,
```
JeanMertz: nit: The advertised `envs = { value = ["MY_EDITOR"], strategy = "replace" }` form is accepted by deserialization but absent from the generated schema. `Field::generate_schema_type` still infers from `self.value` (`Vec<String>`), so schema consumers see an array-only field and cannot discover the wrapper or validate this spelling. The same mismatch applies to the other newly opted-in fields.

Have the input schema include the `partial_via` shape, and add an assertion that it describes both the plain list and the strategy object.

crates/contrib/schematic_macros/src/common/field.rs
```diff
@@ -150,7 +150,11 @@ impl Field<'_> {
             }
             value_type
         } else {
-            FieldValue::value(result.value)
+            // `partial_via` applies to a plain field too, so a list of scalars
+            // can carry a wrapper that knows its own merge strategy. The
+            // partial holds the via type and `generate_from_partial_value`
+            // converts back to the field's own type.
+            FieldValue::value(result.partial_via_ty.as_ref().unwrap_or(result.value))
```
JeanMertz: theoretical: A plain `#[setting(partial_via = MergeableVec::<String>)]` field declared as `Option<Vec<String>>` or `Box<Vec<String>>` loses its outer type information here. `FieldValue::value` inspects only the via type, so `is_nullable()` and `is_outer_boxed()` are false, and the conversion emits `Into::into(partial.field.unwrap_or_default())` rather than the intended optional/boxed conversion. The nested branch above explicitly copies this information from the declared field.

Copy the outer information for the plain branch too and add macro coverage. I found no current optional or boxed scalar-list `partial_via` field in JP, so this has no named user input today and does not gate the PR.

crates/jp_config/src/assignment.rs
```diff
@@ -1051,6 +1034,62 @@ impl KvAssignment {
...
+            return Ok(());
+        }
+
+        // An object declares a strategy alongside the value, so it is parsed as
+        // the wrapper rather than element by element.
+        if let KvValue::Json(value @ Value::Object(_)) = self.value.clone() {
+            let merged =
+                serde_json::from_value(value).map_err(|error| kv_error(&self.key, error))?;
+
+            *vec = Some(merged);
```
JeanMertz: blocking: With an existing MCP argument list `["serve"]`, `--cfg 'providers.mcp.bookworm.arguments:+={"value":["--verbose"],"strategy":"append"}'` silently replaces it with `["--verbose"]`. This branch ignores `self.is_merge()` and overwrites the accumulated partial. `apply_cfg_args` assigns directly into that partial, so there is no later merge against the discarded `["serve"]` for the wrapper's `append` strategy to act on.

Handle the merge-assignment operator in the object branch as well as the list branch, applying the declared list strategy against the existing value. Add a test starting with a nonempty list; accepting an object into an empty partial does not catch this.

crates/contrib/schematic_macros/src/common/field.rs
JeanMertz: Accepted and fixed locally, not pushed yet. The plain `partial_via` branch preserves the declared field's optional/boxed flags. Non-nullable boxed via fields also unwrap the partial (or report a missing required field) before conversion and boxing.

Added compile-and-run coverage in `crates/jp_config/src/list_strategy_tests.rs` for `Option<Vec<String>>`, `Box<Vec<String>>`, and `Option<Box<Vec<String>>>`, including absent values and required-field validation. The fixture failed to compile before the fix and passes with it.

crates/jp_config/src/assignment.rs
JeanMertz: Accepted and fixed locally, not pushed yet. Object merge assignments call the field's merge function against the accumulated value. MCP arguments and variables use `ordered_vec_with_strategy`, so repeated flags survive; the other opted-in lists use `vec_with_strategy`. Ordinary object assignment still replaces the value.

Added nonempty-list regressions for append, prepend, replacement, repeated MCP flags, and the editor field's dedup policy in `list_strategy_tests.rs`. The append/prepend regressions reproduced the discarded prefix before the fix and pass with it.

crates/jp_config/src/assignment.rs
JeanMertz: Accepted and fixed locally, not pushed yet. Appends and indexed assignments mutate the wrapped vector through `DerefMut`, preserving `strategy`, `dedup`, and `discard_when_merged`. Whole-list assignment still replaces the wrapper, and null still clears the optional field.

The sequential `replace` then append regression checks the exact serialized wrapper, round-trips the partial through TOML, and merges it over `editor.envs = ["EDITOR"]`. The result is exactly `["VISUAL", "MY_EDITOR"]`. A separate indexed-edit regression checks preservation of all wrapper metadata. Both reproduced the metadata loss before the fix.

crates/jp_config/src/delta.rs
JeanMertz: This is already resolved at the reviewed head by commit `7d9d1daf` (#1135). `ConversationStream::add_config_delta` in `stream.rs:613-623` records nonempty applies unchanged; it does not diff them again. The query path computes its diff in `turn_config_delta` before recording it.

No additional production change is needed here. The existing query-shaped append regression, `appending_an_access_rule_keeps_the_rules_already_there`, passes, as do the full `jp_conversation` and `jp_cli` test suites.

crates/jp_config/src/editor.rs
JeanMertz: Accepted and fixed locally, not pushed yet. `generate_schema_type` infers from the `partial_via` type when present, so the input schema includes the wrapper's accepted shapes.

Added `scalar_list_schema_describes_array_and_strategy_object`: it checks that `editor.envs` has both an array variant and a strategy-object variant, including the value array and metadata fields. It failed against the array-only schema and passes with the fix. The full `jp_config`, `jp_conversation`, and `jp_cli` suites pass.

PR #1131, opened 2026-09-08
feat(config): Let every map state how it merges

JeanMertz:
A map-valued setting can say how it combines, the same way a list
can. Ten fields gain it: `providers.mcp`, `conversation.tools`,
`providers.llm.aliases`, `plugins.command`, `template.values`, and a
tool's `parameters`, `questions`, `options` and `properties`.

Removing an entry from any of them reaches the conversation, which an
entry-wise merge could not express: the conversation kept the tool or
server the user had deleted, and recomputed the same empty delta every
turn.

A conversation's own snapshot is emitted key by key rather than as a
replacement, so a server, tool or alias the workspace config gained
after the conversation was created still reaches it.

Also fixes how a generic type is named in the schema.
`MergeableMap<ToolConfig>` and `MergeableMap<ToolParameterConfig>` both
answered `MergeableMap`, so a consumer resolving a reference by name
walked a value against whichever it met first. That reached users
through stored conversations: either valid keys were deleted, or a stale
one survived for typed deserialization to reject, which discards the
whole stored config rather than the key. The arguments are appended to
the name.

`merge_nested_indexmap` and the four strategy-less delta helpers are
deleted; nothing needs to guess any more.


crates/jp_config/src/delta.rs
```diff
@@ -95,6 +97,47 @@ pub fn delta_mergeable_vec<T: Clone + PartialEq>(
...
+    T: PartialConfigDelta + PartialEq,
+{
+    if prev.keys().any(|key| !next.contains_key(key)) {
+        // Stated rather than inherited from `next`'s shape: a plain map
+        // deep-merges on the fold and brings the dropped key back.
+        return MergeableMap::Merged(MergedMap {
+            value: next.into_map(),
+            strategy: Some(MergedMapStrategy::Replace),
+            discard_when_merged: false,
+        });
```
JeanMertz: blocking: A model-only change on a conversation with configured tools or MCP servers clears those unrelated maps. `ConversationStream::add_config_delta` calls `config.to_partial().delta(*delta)` with a sparse patch, including the already-reduced patch from `get_config_delta_from_cli`. Its absent maps are empty, so this branch emits `replace {}` for them. `jp config set --id <id> --cfg assistant.model.id=…` reaches the same path and persists the deletions while reporting success.

Separate snapshot diffing from sparse-patch application, and avoid re-diffing an already-computed delta as if it were a complete snapshot. Apply the distinction to `delta_mergeable_value_map` too. A stream-level regression test should change only the model and assert that unrelated maps survive persistence and reload.

crates/jp_conversation/src/compat.rs
```diff
@@ -392,7 +433,8 @@ fn strip_struct<'a>(
         return 0;
     };
 
-    let entry_schema = flattened_entry_schema(struct_type);
+    let flattened = flattened_field_schema(struct_type);
+    let entry_schema = flattened.and_then(map_value_schema);
```
JeanMertz: blocking: Removing one of two tools produces a stored delta shaped like `{"conversation":{"tools":{"value":{"bash":{"source":"local"}},"strategy":"replace"}}}`. Here the flattened field is always reduced to the plain map's entry schema, so `strip_struct` walks `value` as a `ToolConfig`. It removes `bash` as an unknown field, turning the retained-tools map into `{}`. Reloading the conversation then applies an empty replacement and loses the tool the user kept.

Select the flattened field's wrapper/plain variant from the leftover object before descending. For the wrapper form, walk `value` as the map and preserve the strategy metadata. Add a `deserialize_partial_config` regression test using a non-empty flattened replacement.

crates/jp_config/src/providers.rs
```diff
@@ -85,11 +106,9 @@ impl ToPartial for ProviderConfig {
...
         Self::Partial {
             llm: self.llm.to_partial(),
-            mcp: self
-                .mcp
-                .iter()
-                .map(|(k, v)| (k.clone(), v.to_partial()))
-                .collect(),
+            // Per key rather than `replace`: a server the workspace config
+            // gained after this conversation was created still reaches it.
+            mcp: map_to_partial_per_key(self.mcp.iter()),
```
JeanMertz: blocking: If `bookworm` is still declared in workspace config, running a query with `--cfg 'providers.mcp.bookworm:=null'` removes it from the stream, but the next ordinary query brings it back. `Query::apply_conversation_config` resolves the stream and calls this `to_partial`; the plain map forgets the explicit removal. `ConfigPipeline::partial_with_conversation` then calls `fill_from`, whose `fill_map` inserts `bookworm` from the workspace again.

The snapshot needs to preserve enough deletion information to distinguish an intentionally removed entry from a newly added workspace entry. Merely emitting every snapshot per key cannot preserve both behaviors. Please test a removal followed by a fresh pipeline build against the unchanged workspace, as well as the existing workspace-addition case.

crates/jp_config/src/delta.rs
```diff
@@ -95,6 +97,47 @@ pub fn delta_mergeable_vec<T: Clone + PartialEq>(
...
+        .filter_map(|(key, next)| {
+            let Some(prev) = prev.get(&key) else {
+                return Some((key, next));
+            };
+
+            if prev == &next {
+                return None;
+            }
+
+            let delta = prev.delta(next);
```
JeanMertz: blocking: A replacement layer that keeps a map key but removes fields inside its value is not recorded correctly. For example, replace `conversation.tools.bash.parameters` with a `cmd` parameter that keeps `type = "string"` but omits its former `enum`. No parameter key disappears, so this calls `PartialToolParameterConfig::delta`; its `delta_opt` drops the `Some(enum) -> None` change and the stored enum survives. Similarly, clearing a tool's `style` never reaches the new `PartialToolConfig::delta_with_unsets`, because this helper calls `delta` directly.

Handle changes that deep merging cannot reproduce inside surviving entries, either by propagating their unsets or by replacing the affected entries with complete values. The plain-value helper needs the same consideration for object-valued `JsonValue`s: carrying the new object whole still deep-merges its removed properties back in. Add a fold test where the outer keys stay unchanged but a nested field is removed by `strategy = "replace"`.

crates/jp_config/src/conversation/tool.rs
```diff
@@ -517,16 +587,26 @@ pub struct ToolConfig {
...
     /// Unknown options are silently forwarded.
-    #[setting(nested, merge = merge_nested_indexmap)]
-    pub options: IndexMap<String, JsonValue>,
+    ///
+    /// Entries merge by key, so an option set in a later layer joins the ones
+    /// an earlier layer set.
+    /// Declare the map as `{ value = { … }, strategy = "replace" }` to drop
+    /// them instead.
+    #[setting(nested, merge = map_with_strategy)]
+    pub options: MergeableMap<JsonValue>,
```
JeanMertz: blocking: The documented wrapper is parsed differently through `--cfg`. For example, `--cfg 'conversation.tools.cargo_check.options:={"value":{"profile":"release"},"strategy":"replace"}'` goes through `assign_to_entry(&mut self.options)`, which dereferences the wrapper to an `IndexMap` and inserts literal `value` and `strategy` entries. It never selects `MergeableMap::Merged`. The command succeeds, but `cargo_check` receives no top-level `profile` and uses its default build profile.

Make whole-map assignment wrapper-aware while retaining per-entry assignment and clear semantics. Please cover the same configuration through TOML and CLI JSON; the provider map assignment also treats the metadata as entry names.

JeanMertz: Correct in every detail, including that the command reports success.

Traced it: `assign_to_entry` takes the whole-collection branch (`assignment.rs:311`), `Strategy::Set` clears the map, and `:326-343` inserts each top-level key of the object as an entry. `&mut self.options` deref-coerces from `MergeableMap` to `IndexMap`, so `Merged` is never selected. The tool gets options named `value` and `strategy`, and no `profile`.

What earns the fix beyond the outcome: the PR's own doc comments advertise this form on four fields, and lists already accept both shapes through `--cfg` via `try_some_mergeable_strings`. So it documents a spelling that works in TOML and silently misbehaves on the command line, and it is inconsistent with the list it is modelled on.

Fixed with `assign_to_mergeable_entry`, which tells the wrapper from a map the way the map's own deserializer does — both `value` and `strategy` present. Per-entry assignment and `:=null` clearing are unchanged, and a named entry still lands inside whatever wrapper the map already carries, since naming one entry says nothing about how the map combines.

The test asserts the two spellings agree rather than that one of them works: the same configuration through TOML and through `--cfg` JSON, checked for the strategy, the option the tool receives, and the absence of an option named `strategy`. It was red on the `cfg` half with exactly the shape you describe — `Map({"value": …, "strategy": …})`.

Applied to the provider map too, and to `labels`, `aliases`, `tools` and `template.values`, which all had the same arm.

crates/jp_config/src/delta.rs
JeanMertz: The re-diff this depends on is gone, so the deletions do not happen.

`add_config_delta` at this head (`stream.rs:647-658`) records the delta as given and drops only an apply that merges nothing and clears nothing. Grepping `to_partial().delta(*delta)` across `jp_conversation` and `jp_cli` finds nothing. That came from #1135, which is in this PR's base.

`jp config set --id` does not re-diff either. `set.rs:61` calls `config_pipeline::override_to_record`, which merges the override into the conversation's partial, compares the two *resolved* configs, and returns **the override verbatim** when it changes anything (`config_pipeline.rs:64-78`). The sparse patch is never read as a snapshot.

I checked the remaining callers rather than assuming: three sites reach `.delta(…)`, and each diffs two complete states — `query.rs:2344` (two resolved configs), `lib.rs:1109` (a reset's state against the post-reset partial), `stream.rs:1657` (successive per-event configs). Nothing hands a sparse patch to this branch.

The ambiguity underneath your point is real and tracked. `0fgdq7z` is about exactly it: `MergeableMap::Map({})` means both "this layer says nothing" and "this map is empty", and it records why it is not fixed yet — "T-0ffsw7e removed the only caller that read a delta as though it were a snapshot, so nothing currently reaches the ambiguity. It is unreachable, not absent" — while naming this PR as what widens its surface from two rule lists to the whole config tree. That is where I would keep it.

crates/jp_config/src/delta.rs
JeanMertz: Taking it, with one correction to the framing.

The sharpest part is the second half, and it is exact: this helper calls `prev.delta(next)`, and the only route to a `PartialToolConfig` is `PartialToolsConfig::tools`, so the `delta_with_unsets` this PR *adds* had no caller at all. Dead code, in a commit whose claim is that clears are recorded. Clearing a tool's `style` was not recorded, and neither was anything else inside a surviving entry.

Where I would push back: "not recorded correctly" reads as a regression, and for tools it is not one. On main, `PartialConversationConfig::delta_with_unsets` called `self.tools.delta(next.tools)` and `delta_map` also used plain `delta`, so your `parameters.cmd.enum` case was already lost. The PR does delete `delta_map_with_unsets`, which main used for `providers.mcp` — but `PartialMcpProviderConfig` has no `delta_with_unsets`, so that descent hit the default and reported nothing either. The machinery was there and inert.

Fixed by wiring rather than deleting, since #3956060484 needs the same signature. `delta_mergeable_map_at` takes a prefix and reports: a dropped key's path, and each surviving entry diffed with `delta_with_unsets` under its own dotted path. Every `delta_with_unsets` call site moved over, and `PartialToolConfig::delta_with_unsets` is now reachable.

Two entry types needed completing before the descent bought anything. `PartialToolParameterConfig` had no `delta_with_unsets`, which is your `enum` example, and `PartialMcpProviderConfig` had one declined by a comment claiming no field needed a path — true of `arguments` and `variables`, false of `checksum`. Both have one now.

The regression test is at the helper: clearing `checksum` inside a surviving `kagi` entry reports `providers.mcp.kagi.checksum`. Red before the change with an empty list.

On the `JsonValue` half of your note — an object-valued option whose properties are removed still deep-merges them back. That one I have left, and it is the `delta_mergeable_value_map` path: its values carry no partial, so there is no per-field diff to report through. It wants the same treatment as the outer map, which is a value-level strategy on `JsonValue` rather than a path, and that is `0fgdq7z` territory.

crates/jp_config/src/providers.rs
JeanMertz: Confirmed step by step, and your last sentence is the one that settles the design.

The chain: `--cfg providers.mcp.bookworm:=null` removes the entry (`assignment.rs:294-297`); the delta takes the `Replace` branch and the stream resolves without it; the next query calls `config_unsets()` (`stream.rs:567-586`), which accumulates only `apply.unsets`. `delta_mergeable_map` never wrote to `unsets` — it put the removal in the value — so the path was absent, and `partial_with_conversation`'s `fill_map` put `bookworm` back.

"Merely emitting every snapshot per key cannot preserve both behaviors" is exactly right, and it is why the fix is not in `to_partial`. Per-key is what makes `a_server_added_to_the_workspace_reaches_an_existing_conversation` pass; the same property is what loses the removal. The two need different carriers.

So a dropped key is now said twice. The value carries `replace`, which is what the conversation's own fold applies. The path joins `unsets`, which is what stops the layer above from restoring the key after filling. `config_unsets`'s existing filter already handles the re-add case: a key set again by a later delta fails its "clearing this changes nothing" probe and drops out of the list.

Two tests, since the halves are in different crates. `a_removed_mcp_server_is_recorded` now asserts the path is reported — its old assertion was `unsets.is_empty()`, which is the premise this overturns. And `a_removed_server_survives_the_next_invocation` runs the round trip you asked for: the producer computes the delta, the pipeline is built against the unchanged workspace, and the server stays gone while the one the user kept is still there. It takes the paths from the delta rather than naming them, so it fails if either half regresses — which I checked by suppressing the reporting and watching it go red.

crates/jp_conversation/src/compat.rs
JeanMertz: Correct, and the worst of the five: it is data loss on read, in the layer that exists to prevent exactly that.

Walked it through. `map_value_schema` returns the Map variant's value type unconditionally (`compat.rs:500-506`), so `entry_schema` is `ToolConfig`. `has_flatten` keeps `value` and `strategy` in the object, then the loop finds no non-flattened field by either name and walks both against `ToolConfig`. `value` is `{"bash": …}`, `ToolConfig` flattens nothing, `retain` drops `bash` as unknown, and the stored replacement becomes empty. `strategy` is a string, so `as_object_mut` fails and it survives untouched.

The doc comment on `map_value_schema` states the assumption that fails — "Flattened, its entries are sibling keys of the struct around it, which is the plain map's shape" — true of the plain form, false of the wrapper the producer writes.

Fixed as you suggest: `strip_struct` tests the leftover keys for the wrapper pair first, and when it matches, walks `value`'s entries against the map's value schema and leaves the metadata alone. The `value`+`strategy` test is factored into `states_a_strategy` so `strategy_carrying_variant` and this share one rule rather than spelling it twice.

Both tests you asked for. The stripping one pins the retained tool and the untouched strategy; the `deserialize_partial_config` one asserts `bash` is still in the parsed config. Each was red first, and the second's failure output is the bug stated plainly: `Merged(MergedMap { value: {}, strategy: Some(Replace) })`.

A tool named `value` still resolves as a tool — it needs the sibling `strategy` before the object reads as a wrapper, which is what `strip_descends_into_a_tool_named_after_the_flattened_field` covers.

crates/jp_config/src/assignment.rs
```diff
@@ -352,6 +355,39 @@ impl KvAssignment {
...
+        V: AssignKeyValue + Default + Clone + DeserializeOwned,
+    {
+        if self.key.is_empty()
+            && let KvValue::Json(Value::Object(object)) = &self.value
+            && object.contains_key("value")
+            && object.contains_key("strategy")
+        {
+            let value = self.value.clone().into_value();
+            *map = serde_json::from_value(value).map_err(|error| kv_error(&self.key, error))?;
+            return Ok(());
```
JeanMertz: blocking: With `providers.mcp.bookworm` already configured, `--cfg 'providers.mcp:+={"value":{"kagi":{"type":"stdio","command":"kagi"}},"strategy":"deep_merge"}'` drops `bookworm`. This branch replaces the accumulated map without checking `self.is_merge()`, so a merge assignment silently becomes a replacement and the query records the unintended deletion.

For `:+=`, merge the parsed wrapper with the existing map using `map_with_strategy`; retain direct assignment for `:=`. `try_some_mergeable_vec` already makes this distinction. Please test a wrapper merge against a non-empty map, including `keep` preserving an existing entry.

crates/jp_config/src/plugins.rs
```diff
@@ -56,26 +61,29 @@ impl AssignKeyValue for PartialPluginsConfig {
...
-                        None => next,
-                    };
-                    Some((name, next))
-                })
-                .collect(),
+            command: delta_mergeable_map_at(
+                &path(prefix, "command"),
+                &self.command,
+                next.command,
+                unsets,
```
JeanMertz: blocking: If the workspace declares `plugins.command.serve`, a query loading a TOML layer with `[plugins.command]`, `value = {}`, and `strategy = "replace"` removes it for that turn, but the next query restores it. The reported `plugins.command.serve` unset does not delete an entry: `PartialPluginsConfig::assign` uses `entry(name).or_default().assign(kv)`, and assigning null merely empties the plugin config. After the replacement folds, `config_unsets()` probes that path, creates an empty `serve` entry, sees a difference, and discards the unset. `fill_map` then restores the workspace entry on the next invocation.

Make plugin-entry clearing remove the key without inserting absent entries, as `assign_to_entry` does. Add a stream-to-pipeline test that checks `config_unsets()` after folding the replacement, not just the paths produced by this helper.

crates/jp_config/src/delta.rs
JeanMertz: blocking: At `c08e8dd`, the `parameters.cmd.enum` example still fails at the fold, despite the added path reporting. `PartialToolParameterConfig::delta_with_unsets` emits `conversation.tools.bash.parameters.cmd.enum`, but `PartialAppConfig::unset` delegates to assignment, and `PartialToolConfig::assign` accepts only the whole `"parameters"` key. The nested path reaches `missing_key`; `fold_config_delta` warns and continues, and deep merging retains the enum. Please make the emitted paths executable or use a replacement that does not depend on them. A test must apply the unsets and fold the delta, not merely assert the reported path.

The acknowledged `JsonValue` case also has a concrete input in this repo: replacing `web_fetch.options` with the same top-level keys while removing one hostname from `options.domains`. The plain-map delta deep-merges that hostname back. This need not wait for a new value type: the value-map helper already carries changed entries whole, so emitting those entries with the existing shallow `MergedMapStrategy::Merge` can replace their objects without replacing unrelated entries. Please cover that fold as well.

crates/jp_config/src/providers.rs
JeanMertz: blocking: The query path reports the removal, but `config set --id` still loses it across invocations. Start with a conversation and workspace both declaring `bookworm`, then run `jp config set --id <id> --cfg 'providers.mcp:={"value":{},"strategy":"replace"}'`. `Set::set_in_conversations` passes the override returned by `override_to_record` directly to `add_config_delta`. Conversion from `PartialAppConfig` creates an `ApplyDelta` with empty `unsets`, so the replacement removes the server from the stream but `config_unsets()` has no removal to report. The next query fills `bookworm` back from the workspace.

Preserve the raw override's merge semantics, but also record the removed-entry paths derived from applying it to the accumulated state. Please test the `config set` producer through `config_unsets()` and a fresh pipeline build; the added test exercises snapshot diffing directly and does not cover this producer.

crates/jp_config/src/conversation/tool.rs
JeanMertz: nit: The options example is handled, but the whole tools map still rejects the CLI wrapper. For example, `--cfg 'conversation.tools:={"value":{"bash":{"source":"local"}},"strategy":"replace"}'` takes the empty-key arm in `PartialToolsConfig::assign`, which calls `try_merge_object` before `assign_to_mergeable_entry` can inspect the wrapper. It dispatches `value` and `strategy` as tool names and fails assignment.

This is a visible error rather than silent misconfiguration. Recognize the wrapper at the whole-`ToolsConfig` assignment boundary, preserving the separate `*` defaults, or document that this map's strategy currently requires a config file. The options-only CLI test does not exercise the flattened map.

crates/jp_config/src/assignment.rs
JeanMertz: Taken. The merge assignment was a replacement, and the deletion it makes is recorded.

Confirmed the reach before fixing it: `--cfg` assigns onto the partial `partial_with_conversation` builds (`config_pipeline.rs:325-331`), which is the conversation filled from the file layer — so the map holds every server the workspace declares by the time this branch runs. `*map = serde_json::from_value(value)?` then took the whole thing, and the turn's own delta saw a dropped key, stamped `replace`, and reported `providers.mcp.bookworm` as an unset. Silent, and persisted.

`try_some_mergeable_vec` making the distinction is the part that settles it: the same wrapper spelled for a list keeps what is already there, and spelled for a map did not.

Fixed as you describe, with one departure. The vec helper takes its merge function as a parameter because lists have two of them (`vec_with_strategy` and `ordered_vec_with_strategy`); maps have one, and every field that declares a strategy declares `map_with_strategy`, so `assign_to_mergeable_entry` calls it directly and the eight call sites stay as they are. `:=` states the map outright, `:+=` combines under the strategy the wrapper names.

Four tests at the helper. A `deep_merge` wrapper merged into a map holding `bookworm` keeps it and adds `kagi`; a `keep` wrapper leaves an entry that is already there at its own value, which is the one strategy whose result names itself; `:=` still replaces; and `value` without `strategy` still names an entry. The two merge tests were red first, one of them with `no entry found for key` — the entry the user never mentioned.

crates/jp_config/src/conversation/tool.rs
JeanMertz: Taken, though not on the severity — on the rule.

My read was that this one stops at a failed command: `try_merge_object` dispatches `value` as a tool name, `PartialToolConfig::assign` answers `UnknownKey`, and the user sees it. Contained and visible, where I would normally leave it and note that the flattened map's strategy is a config-file spelling.

The rule that decides it instead is the one the options fix was made under: a spelling that works in a config file works on the command line. Two spellings for one concept that disagree is the defect, whichever way the disagreement surfaces.

So the whole-map boundary now recognises the wrapper before dispatching: the `""` arm tests for `value` beside `strategy` and hands the object to the map, and everything else still goes key by key, including the `*` defaults block, which is a sibling field and never reached that arm.

The test is the two spellings side by side rather than the CLI alone — `tools_map_accepts_a_replace_strategy` now builds the same map from TOML and from `--cfg` and asserts the same three things of each: the strategy survives the flatten, `my_tool` is there, and neither `value` nor `strategy` became a tool. Red first on the `cfg` half with exactly what you predicted: `KvAssignmentError { key: "value.my_tool", error: UnknownKey { … } }`.

crates/jp_config/src/delta.rs
JeanMertz: First half taken and fixed at the paths, as you asked. Second half not in this change; reasoning below.

**The dead paths.** Correct: `PartialToolConfig::assign` matched `"parameters"` exactly, so `conversation.tools.bash.parameters.cmd.enum` fell through to `missing_key`, `fold_config_delta` warned, and the enum survived. Everything the added `delta_with_unsets` emitted under `parameters` was unexecutable.

I made them executable rather than stop emitting them. `PartialToolParameterConfig` now implements `AssignKeyValue` — `type`, `default`, `required`, `summary`, `description`, `examples`, `enum`, `items`, `properties` — and the arm routes by prefix through `assign_to_mergeable_entry`. `Box<T>` forwards `AssignKeyValue` so `items` descends into the same shape. A side effect worth having: `--cfg 'conversation.tools.bash.parameters.cmd.enum:=null'` is now something a user can type.

Two tests. One walks every field of a parameter in by path and then clears `enum`, asserting the parameter keeps its `type`. The other is the one you asked for — it applies the unsets and folds the delta: the delta reports `conversation.tools.bash.parameters.cmd.enum`, `unset` applies it, the merge lands, and the enum is gone while `type` stands. Red first with the `UnknownKey` you predicted, at the `expect` on `unset`.

One correction to the framing, in the same direction as my earlier one. The dropped-*key* paths (`…parameters.cmd`) were never the live case: `fill_map` keeps a tool both layers hold whole, so nothing restores a removed parameter, and the value's `replace` already does that work. What was actually lost is the case you named — a field cleared inside a surviving entry.

**The `JsonValue` half.** Not fixed here, and your input is real: `.config/jp/tools` parses `web_fetch`'s `options.domains` as a map of hostnames, and dropping one of them deep-merges it back through `merge_values`.

Your mechanism is right, and it needs one guard I would not want to land without. Stamping the carried entries `MergedMapStrategy::Merge` does exactly what you say — `map_with_strategy` inserts the whole value for that strategy, so the changed option's object replaces and its siblings are untouched, no new value type needed. But `MergeableMap::is_empty` answers `false` for a `Merged` carrying a strategy even with no entries (`types/map.rs:110-115`), so stamping unconditionally makes every sparse partial carry a non-empty `options` and `template.values`, and the stream gains a config delta event per turn that changes nothing. That is the trap the `compaction.rules` comment in this PR records, arriving from the other side. The helper has to wrap only when there is at least one changed entry and return a plain empty map otherwise, with a test for the no-change case alongside the fold test.

Happy to do it here if you want the five landing together; otherwise I will file it as a follow-up ticket naming the `domains` input and the empty-case guard, so it is not carried by this thread alone.

crates/jp_config/src/plugins.rs
JeanMertz: Taken. Your chain holds line for line, and grepping the set makes it a one-liner: `plugins.command` was the only map this PR converted that still assigned through `entry(name).or_default().assign(kv)`. `labels`, `tools`, `parameters`, `properties`, `options`, `mcp`, `aliases` and `template.values` all route through `assign_to_mergeable_entry`, whose null branch removes the key.

So clearing the path inserted an empty `serve` instead of taking it, `config_unsets`'s `cleared == partial` probe saw a difference and dropped the path, and `fill_map` put the workspace entry back on the next invocation.

Fixed by routing it like the rest, which also gets the whole-map form and the wrapper for free.

On reachability, since it changes what the fix is worth: before this, `plugins.command:={…}` hit `missing_key`, so the only producer was a config file layer carrying the wrapper. That is no longer true after the fix, which is a second reason the removal has to stick rather than an argument against it.

The test is the pipeline round trip: the delta reports `plugins.command.serve`, the pipeline is built against a workspace still declaring it, and the plugin stays gone while the one the user kept is still there. Red first with its own message — `an entry that is there is not a gap for filling to close`.

crates/jp_config/src/providers.rs
JeanMertz: Taken. Confirmed the whole path: `set_in_conversations` hands `override_to_record`'s return to `add_config_delta`, and `From<PartialAppConfig> for ConfigDelta` (`stream.rs:313-317`) builds an `ApplyDelta` with empty `unsets`. The `replace` removed the server from the stream, `config_unsets()` had nothing to report, and filling restored it.

Worth naming: this hole is new in this PR. Before it, `providers.mcp` was a plain `IndexMap` merged by `merge_nested_indexmap`, and no override could drop a key. Lists never needed the equivalent, because filling does not merge list elements back in — it is maps that have gaps to close.

Fixed as you describe, keeping the two carriers separate. `override_to_record` returns the paths alongside the override: it already builds `current + override`, so diffing `current` against that merged state reports the keys the override removed. The recorded value is still the override verbatim — its merge semantics are the point — and only the paths are harvested. The harvest is tight by construction: merging never clears a field, so the only paths it can produce are the keys a `replace`-stamped collection dropped.

Both producers build the `ApplyDelta` explicitly now, so the empty-`unsets` conversion is no longer on this route: `config set --id`, and the editor-provided config in `query.rs`, which had the same shape.

The test runs the producer rather than the diff: it parses the same `--cfg 'providers.mcp:={"value":{…},"strategy":"replace"}'` string a user types, asserts the reported path is `providers.mcp.bookworm`, folds the returned delta onto the conversation, and builds a fresh pipeline against the unchanged workspace — the removed server stays gone, the kept one is there. Red first at the first assertion, `left: []`.

PR #1132, opened 2026-09-08
refactor(config)!: Flatten the model parameter collector

JeanMertz:
A model parameter JP does not model is collected into `other`, which
was a real key on the wire: `presence_penalty = 0.5` was stored as
`other = { presence_penalty = 0.5 }`. It is flattened, so the block
holds the parameter directly and `other` is gone from the generated
`config.toml` and from the schema as a key to write.

A config file or stored conversation written with the nested table is
hoisted on read, so nothing needs migrating by hand.

Deleting the hand-rolled collector removes a hand-maintained list of
known keys, so a newly added typed field can no longer be silently
forwarded to the provider as a raw parameter. It also inverts a
data-loss risk: stripping skips a struct holding a flattened field, so
provider parameters survive a load that previously kept them only
because `other` happened to be named in the schema.

BREAKING CHANGE: `other` is no longer a reserved parameter name

`--cfg ...parameters.other` names a parameter called `other`, not the
collector. Write an unrecognized parameter directly:
`--cfg ...parameters.presence_penalty=0.5`.


crates/jp_config/src/model/parameters.rs
```diff
@@ -91,93 +91,58 @@ pub struct ParametersConfig {
...
-    /// [assistant.model.parameters.other]
-    /// presence_penalty = 0.5
-    /// ```
-    #[setting(default, merge = schematic::merge::merge_iter)]
+    /// Flattened, so the parameters reach the wire under their own names and
+    /// this field is never a key anyone writes.
+    /// That is also what keeps them: the compat layer strips whatever the
+    /// schema does not name, and skips a struct holding a flattened field for
+    /// exactly this reason.
+    #[setting(flatten, default, merge = schematic::merge::merge_iter)]
```
JeanMertz: blocking: A config with `assistant.model.parameters.verbosity = "high"` and `conversation.inquiry.assistant.model.parameters.temperature = 0.2` silently loses the inherited verbosity on inquiry requests. Serde deserializes the flattened `Option<IndexMap<...>>` as `Some(empty)` when the block contains no unrecognized keys. `PartialParametersConfig::fill_from` uses `self.other.or(defaults.other)`, so that empty map prevents `PartialAssistantConfig::inherit_from` from taking the parent's provider parameters. The former collector left this case as `None`. Preserve absence when nothing was collected, and add a TOML-to-`util::build` regression test that checks inheritance after setting a typed inquiry parameter.

crates/jp_config/src/model/parameters.rs
```diff
@@ -91,93 +91,58 @@ pub struct ParametersConfig {
...
-    let has_explicit = explicit.is_some();
-    if let Some(explicit) = explicit {
-        let explicit: IndexMap<String, JsonValue> =
-            serde_json::from_value(explicit).map_err(DeError::custom)?;
-        other.extend(explicit);
+    // Hoisted after the siblings so a nested entry still wins a collision with
+    // one of the same name, which is what the nested form did when it was the
+    // documented spelling.
+    if let Some(serde_json::Value::Object(legacy)) = map.remove("other") {
+        map.extend(legacy);
```
JeanMertz: blocking: For an existing `.jp/config.toml` containing `[assistant.model.parameters.other]` with `verbosity = "low"`, `jp config set --cfg assistant.model.parameters.verbosity=high` reports success but leaves verbosity at `low` on the next load. `Set::set_in_file` calls `ConfigFile::merge_delta`, which writes the flattened `parameters.verbosity` without removing the legacy entry; this `map.extend(legacy)` then overwrites the value just written. Preserve the promised legacy collision precedence by migrating or updating the legacy table in the file-editing path before merging the delta. Please cover setting a provider parameter in a legacy TOML file and reloading it.

crates/jp_config/src/model/parameters.rs
```diff
@@ -91,93 +91,58 @@ pub struct ParametersConfig {
...
-    let explicit = map.remove("other");
-    let has_explicit = explicit.is_some();
-    if let Some(explicit) = explicit {
-        let explicit: IndexMap<String, JsonValue> =
-            serde_json::from_value(explicit).map_err(DeError::custom)?;
-        other.extend(explicit);
+    // Hoisted after the siblings so a nested entry still wins a collision with
+    // one of the same name, which is what the nested form did when it was the
+    // documented spelling.
+    if let Some(serde_json::Value::Object(legacy)) = map.remove("other") {
```
JeanMertz: theoretical: The advertised parameter named `other` cannot round-trip through `PartialModelConfig`. `--cfg assistant.model.parameters.other=5` serializes as `"other": "5"`, but `map.remove("other")` removes it even when the `Object` pattern fails. An object-valued parameter such as `other := {"mode":"fast"}` is interpreted as legacy data and becomes a different parameter, `mode`, on reload. The added test checks assignment only, so it misses both cases. I haven't identified a supported provider that actually accepts a parameter named `other`, hence the non-blocking classification. Leave non-object values untouched; for objects, distinguish legacy input from current data or keep the name reserved. Extend the test through serialization and the parent deserializer.

crates/jp_config/src/model/parameters.rs
JeanMertz: Taken, and it is a regression this PR introduces.

Verified the chain: `lib.rs:439-443` calls `inherit_from`, which ends in `self.fill_from(parent)` (`assistant.rs:206`), which reaches `other: self.other.or(defaults.other)` (`parameters.rs:280`). Serde's `flatten` collects into `Some({})` where the hand-rolled collector set the field only `if has_explicit || !other.is_empty()`, and `Some({}).or(Some({"verbosity": …}))` keeps the empty map.

The fingerprint was in the PR's own diff, which is the part that stings. `deserialize_leaves_other_unset_when_every_key_is_known` was renamed and its `assert_eq!(p.other, None)` loosened to `is_none_or(IndexMap::is_empty)`, in four places. The representation changed from absent to present-and-empty and the assertions were relaxed to accept it rather than the invariant being held. One of those four was mine, from a rebase — I saw `Some({})`, checked the delta path where `to_partial` puts `Some(map)` on both sides so nothing moves, and did not check `fill_from`.

Fixed at the single choke point, in `deserialize_hoisting_legacy_other`: `other` goes back to `None` when the collected map is empty. `ModelConfig::parameters` is the only owner of a `ParametersConfig`, so that covers every read, and `None` then means the same thing to `fill_from`, `delta_opt` and `is_empty` alike.

Three of the four assertions are tightened back to `assert_eq!(p.other, None)` — they are the invariant, and loosening them is what let this through. I left the fourth, `assign_clears_a_collected_parameter`, permissive: `--cfg …parameters.seed=null` genuinely does say something ("no provider parameters"), unlike a block that never mentioned them, so `Some({})` there is defensible and a future normalization of that path should not break the test.

Your test, added as `an_inquiry_inherits_the_assistant_s_provider_parameters`: a TOML fixture with `verbosity` on the assistant and `temperature` on the inquiry, through `ConfigLoader` and `util::build`, asserting the inquiry resolves with both. Red first with `left: None` on the inherited parameter.

One thing I did not fix, because it predates this PR: `ToPartial for ParametersConfig` (`parameters.rs:295`) also emits `Some({})` for an empty resolved map, feeding the conversation layer's `fill_from`. The resolved `other` was already an empty `IndexMap` for a config without provider parameters, so that is older than the flatten and the choke-point fix does not reach it.

crates/jp_config/src/model/parameters.rs
JeanMertz: Taking the scalar half, narrowing the claim on the object half.

Both mechanisms are exact. `map.remove("other")` is the scrutinee of the `if let`, so the key is gone whether or not the `Object` pattern matches; a scalar `other = "5"` was removed and then dropped. And `other = { mode = "fast" }` is read as the legacy table and becomes a parameter named `mode`.

I took the scalar case despite the `theoretical:` label, because the PR advertises this spelling — the `BREAKING CHANGE` note says `--cfg ...parameters.other` names a parameter called `other` — and `a_parameter_named_other_is_not_a_wrapper` only checked assignment, upstream of where it broke. A claim the commit makes should survive a round trip.

Fixed by matching on the value before deciding: a table is hoisted, anything else goes back into the block under its own name. `deserialize_keeps_a_scalar_parameter_named_other` covers the read directly, and `a_parameter_named_other_is_not_a_wrapper` now carries the assigned value through JSON and TOML serialization and back through the parent deserializer. Both red first; the round trip failed with `Some({})`.

The object case I left as it is and documented instead. `other = { mode = "fast" }` is byte-identical to the legacy table this function exists to migrate, so there is nothing in the document to tell them apart. Keeping the name reserved would contradict what this PR is for, so the field docs now say a parameter may be called `other` except that a *table* by that name is read as the older nested form.

crates/jp_config/src/model/parameters.rs
JeanMertz: Correct, and a regression: before the flatten the delta's `other` serialized as a real `other` key, so `merge_delta` wrote into the same nested table and overwrote it.

Verified: `merge_delta` (`fs.rs:155-167`) deep-merges and removes nothing, so the file ends up holding `verbosity = "high"` at the parameters level beside the untouched `[assistant.model.parameters.other] verbosity = "low"`, and `map.extend(legacy)` let the nested one win. The command reports success and the next load resolves `low`.

I fixed it in the reader rather than the file-editing path. The hoist now inserts a legacy entry only where the block does not already name it, so the current spelling wins a collision. The old precedence belonged to the model where the two forms were alternative spellings and the explicit one was the more deliberate. In this model the nested form is input being migrated, and migrated data overriding current data is backwards. Doing it in the reader also covers a hand edit or any future writer, where a `config set`-specific migration covers only `config set`.

`deserialize_prefers_the_explicit_other_entry_on_collision` asserted the old precedence, so it is now `deserialize_prefers_the_current_spelling_on_collision` and asserts the opposite.

Your regression is `merge_delta_of_a_provider_parameter_beats_a_legacy_other_table` in `fs_tests.rs`: a legacy file, `merge_delta` of the flattened parameter, reparse, assert `high`. I checked it against the old precedence and it goes red, and the failure output is the case you described — both values in the file, `low` winning.

Having the writer also drop the now-dead nested table would leave the file tidier. That's a follow-up, not part of the fix.

PR #1133, opened 2026-09-08
fix(config): Record a change to compaction rules and labels

JeanMertz:
Reordering a compaction rule, or dropping one, reaches the
conversation. Both were computed and discarded, so the conversation kept
compacting by the rules it was created with and recomputed the same
non-delta on every turn.

A label rule the workspace config gained after a conversation was
created reaches that conversation, matching how every other map now
behaves.

The list naming fields whose clear could not be recorded is gone. No
field is exempt from the law that a clear survives a fold.


crates/jp_config/src/conversation/compaction.rs
```diff
@@ -87,23 +87,16 @@ impl AssignKeyValue for PartialCompactionConfig {
...
+            // An empty `rules` says nothing about rules rather than asking for
+            // none: `fill_from` substitutes the built-in defaults for it, so
+            // that is what the config language already means by an empty list.
+            // Every sparse partial reaching this delta carries one, and
+            // replacing with it would record zero rules the user never asked
+            // for, making a later `jp conversation compact` do nothing.
+            rules: if next.rules.is_empty() {
+                MergeableVec::default()
+            } else {
+                crate::delta::delta_mergeable_vec(&self.rules, next.rules)
```
JeanMertz: blocking: Appending a compaction rule to an existing conversation silently removes its earlier rules. For example, with the built-in rule `A`, `jp query --cfg 'conversation.compaction.rules:+=[{"keep_first":2,"reasoning":"strip"}]' 'Continue' resolves to `[A, B]`. `get_config_delta_from_cli` computes the append delta `[B]`, but `ConversationStream::add_config_delta` diffs that delta against `[A]` again. This call then emits `replace [B]`, so the persisted configuration loses `A` and subsequent compaction stops applying its tool-call policy.

Keep already-computed deltas out of snapshot diffing, or otherwise preserve their append semantics. Add a regression test that passes the CLI-produced delta through `add_config_delta` and checks `events.config()`, rather than stopping at the first delta calculation.

crates/jp_config/src/conversation.rs
```diff
@@ -201,7 +201,9 @@ impl ToPartial for ConversationConfig {
             tools: self.tools.to_partial(),
             compaction: self.compaction.to_partial(),
             attachments: vec_to_mergeable_partial(&self.attachments),
-            labels: map_to_mergeable_partial(self.labels.iter()),
+            // Per key rather than `replace`: a label rule the workspace config
+            // gained after this conversation was created still reaches it.
+            labels: map_to_partial_per_key(self.labels.iter()),
```
JeanMertz: blocking: A workspace that gains `conversation.labels.area = "config"` after a conversation was created still does not pass that rule to the resumed conversation. `ConfigPipeline::partial_with_conversation` calls `conversation.fill_from(base)`, not `base.merge(conversation)`. `PartialConversationConfig::fill_from` delegates labels to `MergeableMap::fill_from`, which keeps the existing entries and copies only wrapper metadata, so the workspace-only `area` entry is silently dropped even with this plain-map representation.

Apply per-key gap filling to labels in `PartialConversationConfig::fill_from`, following the MCP map's `fill_map` handling while preserving explicit strategy wrappers. Exercise `ConfigPipeline::partial_with_conversation` in the regression test; the new test's direct `merge` does not cover the runtime path.

JeanMertz: Correct, and worse than a gap: this commit does not deliver what it says it does.

Verified the chain. `partial_with_conversation` is `conversation.fill_from(self.base.clone())` (`config_pipeline.rs:339`), `PartialConversationConfig::fill_from` delegates to `self.labels.fill_from(defaults.labels)`, and `MergeableMap::fill_from` says so in its own doc: "Entries are always kept from `self`; only the `Merged` wrapper metadata is inherited."

The part that stings is that the change is inert on that path. Before it, `to_partial` stamped `Merged{Replace}`, which hits `fill_from`'s early return and keeps `self`'s entries. After it, `Map` falls through to `Self::Map(entries)` and keeps `self`'s entries. Same outcome, different representation. The `to_partial` change is a prerequisite rather than wrong — per-key filling can only apply to a `Map`, since a map stating a strategy is deliberately left alone — but it was half the fix shipped as the whole one.

And you are right about why the test did not catch it. `files.merge(&(), conversation.to_partial())` is a deep merge, which keeps both keys whatever `fill_from` does, so it cannot fail for this bug. It passed for a reason unrelated to the claim.

Fixed by routing `labels` through `fill_map` the way `providers.mcp`, `plugins.command`, `providers.llm.aliases` and `conversation.tools` already do, with the `Merged` passthrough kept.

The regression is `a_workspace_label_reaches_an_existing_conversation` in `config_pipeline_tests.rs`, built against a base declaring `area` and `topic` with a conversation holding only its own `topic`. It asserts both halves: the workspace-only rule arrives, and the conversation's own value wins for the rule it holds. Red first with `Map({"topic": Static("rust")})`.

One thing worth noting for the stack: `a_server_added_to_the_workspace_reaches_an_existing_conversation` in `lib_tests.rs` has the same merge-based shape. It is not a false positive, because `providers.mcp` genuinely fills per key, but it proves the merge path rather than the one the runtime takes.

crates/jp_config/src/conversation/compaction.rs
JeanMertz: The step that does the damage here is gone, so the rules survive.

Your premise about the resolved value is right, and I checked it rather than assuming: the conversation layer is a resolved snapshot stamped `replace` by `vec_to_mergeable_partial`, `:+=` appends onto that, and `fill_from` leaves a `Merged` alone, so the turn resolves to `[A, B]`. I had expected the built-in to be displaced and it is not.

Where the chain breaks is `add_config_delta`. At this head (`stream.rs:330-341`) it records the delta as given and drops only an apply that merges nothing and clears nothing; grepping `to_partial().delta(` and `.delta(*delta)` across `jp_conversation` and `jp_cli` finds nothing. The doc comment above it is now the explanation for why that call must not exist: "an appended list element becomes the whole list, and a field left alone becomes a field emptied." It went with #1135, which is in this branch's base.

Without it: `delta_mergeable_vec(&[A], [A, B])` finds `starts_with` true with no repeat and returns the tail as a plain `MergeableVec::Vec([B])`, an append rather than a replacement. Folding that next turn runs `vec_with_strategy` with `strategy: None`, which appends — `[A] ++ [B]` is `[A, B]`. The stream resolves to what the turn ran with, so the following turn computes no delta at all.

Also worth saying: this commit does not change the append case either way. The previous body was a set difference (`next.filter(|v| !self.contains(v))`), which produces the same `Vec([B])` here. What it changes is reorder and removal, which that body dropped.

The test you ask for is still worth having, and for the reason you give rather than the bug: nothing currently drives a CLI-produced delta through `add_config_delta` and asserts on `events.config()`, and that is the seam where a re-diff would reappear. `0fgdq7z` already tracks the ambiguity that makes one dangerous, so I would file the test against that rather than hold this commit for it.

PR #1135, opened 2026-09-08
fix(conversation, cli): Record a config delta as it was computed

JeanMertz:
A `--cfg` that granted a tool one more `access.fs` rule dropped the
rules it already had, and one that changed only `access.env` cleared
that tool's filesystem rules entirely. An empty `fs` list means
unrestricted (workspace-confined) access, so the tool silently gained
reach the user never granted, and the widening persisted for every later
turn. `conversation.labels` was cleared the same way.

Worse, the conversation never converged. A stored rule set of `[docs]`
where the invocation wanted `[src, docs]` made the next identical run
compute a delta re-adding `src` and dropping `docs`, and the run after
that swap them back, so every `jp` run appended a config delta and moved
the resolved rules again.

`ConversationStream::add_config_delta` reduced each apply to its diff
against the conversation's current config, which read the delta as
though it stated a complete configuration. It does not: it states a
change. An appended list element read that way becomes the whole list,
and a field left alone becomes a field emptied. The delta is now
recorded as it was computed, and an apply that merges nothing and clears
nothing is dropped.

Stream reconstruction was corrupted by the same reading, so a summary
request built from a rebuilt stream saw rules the conversation never
had. Stored deltas also get smaller, since the second diff wrapped every
list in a merge-strategy envelope that is now just the list.

### DIFF
--- a/crates/jp_cli/src/config_pipeline.rs
+++ b/crates/jp_cli/src/config_pipeline.rs
@@ -111,10 +111,9 @@ impl ConfigReset {
 /// The `post` partial is a partial-level diff from the reset point's state to
 /// the invocation's final partial, so it captures post-keyword `--cfg`
 /// directives and command CLI overrides without pinning program defaults.
-/// It is computed directly instead of routing through the empty-diff
-/// suppression path, because that path resolves the stream's current config —
-/// which is not a valid configuration between a `Reset` and whichever `Apply`
-/// restores the required fields.
+/// It is diffed against the reset point rather than against the conversation's
+/// current config: the `Reset` discards that state, so a diff from it would
+/// leave out every field the reset is about to drop.
 ///
 /// [RFD 038]: https://jp.computer/rfd/038
 #[derive(Debug, Clone)]
--- a/crates/jp_cli/src/config_pipeline.rs
+++ b/crates/jp_cli/src/config_pipeline.rs
@@ -15,7 +15,7 @@ use std::path::Path;
 
 use camino::Utf8PathBuf;
 use jp_config::{
-    AppConfig, FillDefaults as _, PartialAppConfig, PartialConfig as _,
+    FillDefaults as _, PartialAppConfig, PartialConfig as _,
     assignment::{AssignKeyValue as _, KvAssignment},
     fs::{load_partial, user_global_config_dir},
     loader::PartialLoaderConfig,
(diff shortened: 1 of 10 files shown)


crates/jp_cli/src/config_pipeline.rs
```diff
@@ -29,6 +29,42 @@ use tracing::{debug, error};
...
+        .merge(&(), overrides.clone())
+        .map_err(jp_config::Error::from)?;
+
+    // Both sides go through the same transform, so the comparison does not rest
+    // on `to_partial` round-tripping a config exactly.
+    let (Ok(before), Ok(after)) = (build(base), build(merged)) else {
+        return Ok(Some(overrides));
+    };
+
+    Ok((before != after).then_some(overrides))
```
JeanMertz: blocking: `jp config set --id <id> --cfg assistant.system_prompt.dedup=false` on a conversation with a nonempty prompt is silently dropped here. Merging this override leaves the prompt text unchanged, and `build` resolves `assistant.system_prompt` to `Option<String>`, so `before == after` even though the override changes how subsequent appends must behave. The command reports success without recording the requested policy.

There is also a false no-op when an earlier stored delta already carries `dedup = false`: `current.to_partial()` removes that metadata, so an append of an existing prompt block is tested with default deduplication and suppressed, although folding it into the actual stream would append it.

Use the stream's accumulated partial state for the check, and preserve overrides that change persistent merge metadata even when resolved values match. A sequence test that sets the dedup policy and then appends an existing block would cover both cases.

JeanMertz: Both halves reproduced, one fixed, one documented as open. Fixed in `6ab818d3`.

**Your diagnosis is exactly right on the mechanism.** `assistant.rs:105-110` inserts a bare `Merged(PartialMergedString::default())` for a nested `system_prompt.*` key, so the promotion at `string.rs:92-99` never fires and no value or `Replace` is injected. `string_with_strategy` then falls to `(Some(p), None) => Some(p)` at line 80 — text unchanged, `dedup` preserved on the partial (line 96) — and `assistant.rs:54` declares the resolved type as `Option<String>`, so `build` erases it and `before == after`. Two red tests confirmed both cases before any change: the metadata-only override recorded 0 deltas, and the append landed once instead of twice.

**What landed.** `ConversationStream::config_partial` exposes the accumulated state `config` was folding privately before resolving it; `config` is now that plus the resolve it always did. `override_to_record` takes that partial and merges the override onto it, so a recorded `dedup` governs the outcome. That fixes your second case — the false no-op — and `an_override_is_compared_under_the_conversations_dedup_policy` pins it, watched red against a simulated round-trip through resolution.

**Where I had to amend your fix.** "Preserve overrides that change persistent merge metadata even when resolved values match" needs the comparison to read metadata, which means comparing two partials. That does not work, and it took me three attempts to establish why: **merging is not shape-preserving, and the artifact is not uniform.** Schematic skips a nested struct whose incoming partial is empty, so setting any one field in `conversation` collapses `attachments` from `Merged{[], Replace}` to `Vec([])` while leaving `compaction.rules` stamped. Which collections collapse depends on which sibling fields the override happens to touch, so no fixed baseline normalizes it and equal metadata can sit in unequal shapes. I tried `merged != current`, then `merged != baseline` with an empty-merge baseline; both fail on every override.

So your first case — an override changing *only* metadata — is **still dropped**, and I've written that into `override_to_record`'s doc comment with the reason rather than leaving it implicit. I'd rather ship the half whose correctness I can demonstrate than a comparison I can't. Happy to be argued out of that if you see a shape-insensitive way to reach the metadata.

**Two bounds worth having, both verified.** `partial_via = MergeableString` appears exactly once, on `assistant.system_prompt` — the only field whose metadata is writable from `--cfg` today. And `MergeableVec`/`MergeableMap` have no `AssignKeyValue` impl at all, so their metadata is unwritable from the command line in either direction; `types/vec.rs:32-37` contemplates adding one, and that is when the remaining gap widens beyond one field.

Also recorded on T-0fgdq7z (`23628c0f`): I closed that ticket earlier this session on the finding that empty and absent behave identically for collections, because the empty value is the identity element of their merge. Your comment shows that reasoning does not extend to `dedup` and `discard_when_merged`, which are read from the accumulated side and are not inert. The closure was narrower than it read, and now says so.

PR #1136, opened 2026-09-08
enhance(cli): Format `--quote` seed with standard markdown rules

JeanMertz:
`jp q --quote` and `--quote=false` now run the quoted assistant message through the same markdown formatter used to render turns to the terminal, using the `style.markdown` wrap width and table settings, before quoting it. A table copied out of a prior response is padded into aligned columns, and long paragraphs rewrap at the configured width, instead of the raw source markdown being quoted as-is.

The formatter's ANSI escapes are stripped back out and horizontal rules are kept as plain `---`, since the seed is markdown source fed back into the next request, not terminal output.

One behavioral side effect: a soft-wrapped assistant message (a raw `\n` with no blank line between the two halves) now reflows into a single line before rewrapping, matching how that same text already renders in the terminal.

### DIFF
--- a/crates/jp_cli/src/cmd/query.rs
+++ b/crates/jp_cli/src/cmd/query.rs
@@ -107,12 +107,14 @@ use jp_llm::{
     },
 };
 use jp_mcp::{StartupSet, id::McpServerId};
+use jp_md::format::Formatter;
 use jp_printer::{LineSink, PrintableExt as _, Printer, RegionStyle, StatusRegion};
 use jp_storage::backend::{FsStorageBackend, Projection};
 use jp_task::task::TitleGeneratorTask;
 use jp_term::width::{display_width, truncate_to_width};
 use jp_workspace::{ConversationHandle, ConversationLock, Id as WorkspaceId, Workspace};
 use minijinja::{Environment, UndefinedBehavior};
+use strip_ansi_escapes::strip_str;
 use tokio::sync::broadcast::error::RecvError;
 use tool::{TerminalExecutorSource, ToolCoordinator};
 use tracing::{debug, trace, warn};
@@ -833,7 +835,7 @@ impl Query {
         // (e.g. brand new conversation) degrades to a warning and the editor
         // opens with whatever else was seeded.
         if let Some(prefixed) = self.input.quote
-            && !seed_quoted_reply(&mut chat_request, view, prefixed)
+            && !seed_quoted_reply(&mut chat_request, view, prefixed, config)
         {
             warn!("--quote: no prior assistant message in this conversation");
         }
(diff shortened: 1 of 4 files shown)


crates/jp_cli/src/cmd/query.rs
```diff
@@ -1366,29 +1368,55 @@ fn blockquote(text: &str) -> String {
...
+/// stripped, since the result is markdown source, not terminal output.
+fn reformat_quoted_message(message: &str, config: &AppConfig) -> String {
+    let markdown = &config.style.markdown;
+    let formatter = Formatter::with_width(markdown.wrap_width)
+        .table_max_column_width(markdown.table_max_column_width)
+        .table_continuation_edge(markdown.table_continuation_edge)
+        .pretty_hr(false);
+
+    let rendered = formatter
+        .format_terminal(message)
```
JeanMertz: blocking: With the default wrap width, quoting an assistant reply containing the inline command `grep 'a  b' file` changes it to `grep 'a b' file`. `TerminalFormatter::format_code` sends the literal through `TerminalWriter::output(literal, true)`, which collapses consecutive spaces. Stripping ANSI cannot restore them, so a visually subtle change to a command's meaning becomes part of the next request.

Use source-oriented Markdown formatting that preserves inline-code literals while wrapping prose, rather than treating terminal rendering as a lossless serialization. Please add a quote regression test with significant consecutive spaces inside inline code.

crates/jp_cli/src/cmd/query.rs
```diff
@@ -1366,29 +1368,55 @@ fn blockquote(text: &str) -> String {
...
+/// paragraphs and aligned tables, using the wrap width and table settings from
+/// `style.markdown`.
+///
+/// Horizontal rules are kept as plain `---` rather than the terminal's
+/// decorative unicode line, and any styling escapes the formatter injects are
+/// stripped, since the result is markdown source, not terminal output.
+fn reformat_quoted_message(message: &str, config: &AppConfig) -> String {
+    let markdown = &config.style.markdown;
+    let formatter = Formatter::with_width(markdown.wrap_width)
+        .table_max_column_width(markdown.table_max_column_width)
```
JeanMertz: blocking: An assistant table with a header such as `Execution time before optimization (milliseconds)` loses part of that header when quoted with the default column limit of 40. `jp_md::table::format_table` truncates headers with `…`; its `test_wide_header_is_truncated_not_wrapped` explicitly verifies this behavior. Long body cells are also split into separate physical rows, which become separate rows when parsed as Markdown.

These are display transformations, but here the truncated text and altered table structure become part of the user's next request. Align seed tables without truncating cells or splitting logical rows, allowing columns to exceed the display limit, and add quote tests with headers and body cells wider than the configured cap.

JeanMertz: Agreed, and fixed in 4fcf4dd.

`reformat_quoted_message` now passes `table_max_column_width(0)`. With no cap and no `terminal_width`, `fit_columns` returns early on the natural widths, so `truncate_to_visual_width` is a no-op (content always fits its column) and no cell can spill onto a continuation line. Columns are still padded and aligned, which is the part of the reformatting worth having; only the display cap is dropped, since nothing on this path has a terminal to fit. I dropped the `table_continuation_edge` builder call at the same time — with no wrapping possible it was dead configuration.

Two tests added, both of which fail without the change: `quote_keeps_a_table_header_wider_than_the_display_cap` uses your 49-column benchmark header, and `quote_keeps_a_wide_body_cell_on_one_row` pins a 49-column body cell to a single row.

crates/jp_cli/src/cmd/query.rs
JeanMertz: The defect is real and is fixed in 96936430, though not the way you suggested — the remedy as described isn't available.

`Formatter::format` (comrak's `format_commonmark`) is the source-oriented path in this codebase, and its `format_code` passes `wrap = true` into an `output` carrying the identical space-skipping loop, so it collapses the same runs. It also writes tables as `| cell |` with no padding, which is the feature this PR exists to deliver. Switching formatters trades one defect for a worse one.

So the fix went into `jp_md::render` instead: `write_code_literal` splits the span's literal on runs of spaces, writes a run of two or more verbatim with no break point, and keeps routing a lone space through the wrapping path. Breaking at a lone space is safe because CommonMark reads the line ending back as a single space, so the span's content survives the round trip; breaking inside a run is not, because the break drops a space just as collapsing did. Long spans still wrap, so the existing inline-code wrap tests are untouched.

Worth noting the collapse predates this PR — JP's terminal already printed `grep 'a b' file` for that input, with nothing on screen to say the text had been altered. What this PR added was the path from display output into request text. I confirmed `test_inline_code_keeps_a_run_of_spaces` fails against the unfixed renderer before committing, and added `quote_keeps_consecutive_spaces_in_inline_code` for the seed end of it.

One thing I checked while I was in there, since it's the same class of leak: headings keep their `#`, links keep `[text](url)` with the URL written unwrapped, and fenced blocks go out through `write_raw`, so those survive the strip intact. Tables and code spans were the two exceptions, not the tip of a pile.

PR #1138, opened 2026-09-08
feat(printer): Hold output while a widget owns the terminal

JeanMertz:
Output produced while a prompt is up now waits for the answer instead of
landing in the middle of the question. A tool result arriving from a
parallel tool, a title generator finishing, an MCP server logging to
stderr: each used to print straight over the widget's frame, on a
terminal in raw mode where a line feed no longer returns to column zero.
The same holds while an external `$EDITOR` has the screen.

A handover is bracketed by the guard `suspend_status` and the prompt
writers already take, so the printer now knows exactly when the terminal
is not its own. For that window the worker keeps ordinary output in a
queue and writes it, in the order it was produced, once the terminal
comes back. Handovers nest, and only the outermost hand-back releases
anything: a permission prompt opening the inline reply widget on top of
itself never gives the screen away mid-question.

A print task says who it belongs to, and only a prompt session's own
writes reach a terminal someone else owns. Nothing is dropped: if a run
ends with a widget still holding the terminal, whatever it was holding
back is written on the way out, because that output is the answer the
user asked for and an untidy final frame is the smaller loss.

### DIFF
--- a/crates/jp_printer/src/printer.rs
+++ b/crates/jp_printer/src/printer.rs
@@ -1,6 +1,7 @@
 //! The printer module.
 
 use std::{
+    collections::VecDeque,
     fmt::{self, Write},
     io,
     sync::{
@@ -158,6 +159,7 @@ impl Printer {
                     rx,
                     delay_control,
                     regions: RegionStack::new(),
+                    held: VecDeque::new(),
                 };
                 worker.run();
             })
@@ -292,13 +294,17 @@ impl Printer {
         StatusRegion::new(id, self.tx.clone(), buffer, refresh)
     }
 
-    /// Suspend status-region rendering until the returned guard drops.
+    /// Hand the terminal to a writer outside the printer until the returned
+    /// guard drops.
     ///
-    /// Blocks until the worker has erased any drawn region and entered the
-    /// suspended state, so a caller handing the terminal to a writer outside
-    /// the printer — an external `$EDITOR`, say — knows the rows are gone
+    /// Blocks until the worker has erased any drawn status region, so a caller
+    /// handing the terminal to an external `$EDITOR` knows the rows are gone
     /// before the child paints.
-    /// Returns immediately when regions are disabled.
+    /// For the guard's lifetime the worker also holds back ordinary output,
+    /// writing it once the terminal comes back: whoever holds the terminal owns
+    /// the cursor, and JP keeps producing while they have it.
+    ///
+    /// Handovers nest; the terminal comes back when the outermost guard drops.
     #[must_use]
     pub fn suspend_status(&self) -> SuspendGuard {
         self.suspend_regions()
@@ -306,10 +312,6 @@ impl Printer {
 
     /// Enqueue a suspension and wait for the worker to apply it.
     fn suspend_regions(&self) -> SuspendGuard {
-        if !self.chrome_repaints() || !self.terminal.permits_regions(self.format) {
-            return SuspendGuard::inert();
-        }
-
         let (ack, rx) = mpsc::channel();
 
         if self
@@ -493,6 +495,7 @@ impl Printer {
         PrinterWriter {
             printer: self,
             target: PrintTarget::Out,
+            origin: PrintOrigin::Content,
         }
     }
 
@@ -502,6 +505,7 @@ impl Printer {
         PrinterWriter {
             printer: self,
             target: PrintTarget::Err,
+            origin: PrintOrigin::Content,
         }
     }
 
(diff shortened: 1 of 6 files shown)


crates/jp_printer/src/printer.rs
```diff
@@ -995,9 +1021,47 @@ impl<O: io::Write, E: io::Write> Worker<O, E> {
...
+    /// Decide whether `task` may be written now, holding it back if not.
+    ///
+    /// Only the prompt session's own writes reach a terminal someone else owns;
+    /// everything else waits for it to come back.
+    fn admit(&mut self, task: PrintTask) -> Option<PrintTask> {
+        if !self.regions.is_suspended() || task.origin == PrintOrigin::Prompt {
+            return Some(task);
+        }
+
+        self.held.push_back(task);
```
JeanMertz: nit: In an `Ask` tool permission prompt, choose `e` (or `r`), then press `Ctrl+X` when the configured editor cannot start. `ToolPrompter::prompt_ask` keeps its outer canvas alive through `inline_edit`, and `report_editor_failure` uses `eprintln`, so this branch holds the recovery notice until the entire permission prompt returns. The user gets the inline widget back without the explanation that the editor failed.

Route this session's recovery notice through a prompt-origin writer without releasing the outer hold. A regression test should check that the notice is visible when the replacement inline prompt opens; checking output after `prompt_permission` returns would miss this.

JeanMertz: Verified and fixed. `prompt_ask` takes `self.canvas()` at the top of the loop body (`prompter.rs:283`), so the `PromptWriter` outlives the `Ok('e')` / `Ok('r')` arms; `inline_edit`'s own `owned_prompt_writer` is moved into `inline_reply` and dropped on return, but the canvas keeps `suspensions > 0`, so the `eprintln` in `report_editor_failure` was queued as `Content`.

Two corrections to the framing, neither of which changes the outcome:

- The notice is deferred, not lost. In the `EditResult::Emptied` case it lands when the iteration's canvas drops, just before the permission prompt redraws; in the other two it lands after `prompt_ask` returns. Late enough to be useless, which is why it's worth fixing, but nothing is dropped.
- This is the only affected call site. `handler.rs:256`/`:343` pass `&mut printer.prompt_writer()` as a temporary that dies with the call expression, and `handler.rs:481` / `dispatch.rs:180` move the owned writer into `inline_reply`. None of them hold a suspension across the failure report.

The fix is in `report_editor_failure` itself rather than at the one call site, so the helper doesn't write on two channels depending on caller:

```rust
printer.prompt_println(format!(
    "\n⚠ Couldn't open your editor: {error}. {recovery}"
));
```

`prompt_println` already exists for exactly this: a line that travels with its question rather than as chrome. The notice's whole job is explaining why the widget re-opened, and on the chrome channel `--quiet` swallowed it outright. The cost is that all four sites move from stderr to `/dev/tty`, so `2>` no longer captures the notice and `--quiet` no longer silences it; the non-tty fallback is `out`, but every one of these paths needs an interactive widget to reach, so I don't think that state occurs. Also switched the sibling `"No editor configured; returning to the menu."` line in `editor_unavailable`, which would otherwise sit on a different channel from its neighbour in the same `match`.

Regression test is `an_editor_failure_reaches_the_user_while_the_permission_prompt_is_open` in `prompter_tests.rs`: it holds an outer `printer.prompt_writer()` across `try_edit_arguments` and asserts the exact notice before dropping it — the mock backend has no hook to observe from inside `inline_reply`, so the condition is reproduced directly instead. Confirmed red against the `eprintln` version (`left: ""`).

JeanMertz: Correction to the previous reply: the fix keeps the notice on stderr rather than moving it to the prompt stream. The diagnosis and the call-site analysis above still stand; only the shape of the fix changed.

`Printer::prompt_eprintln` writes to `PrintTarget::Err` with `PrintOrigin::Prompt` — same stream, same JSON record shape, so `--quiet` still silences it and `2>` still captures it — but the task is admitted while a prompt session owns the terminal instead of queueing behind it. Its doc names the constraint that makes that safe: the caller bypasses the wait, so it is only correct between a widget's frames, when the terminal is back in cooked mode. That is exactly where `report_editor_failure` is called from.

`report_editor_failure` and the sibling `"No editor configured; returning to the menu."` line in `editor_unavailable` both use it, so the two arms of that `match` stay on one channel.

Two tests, both seen red first against `PrintOrigin::Content`:

- `a_prompt_notice_on_the_chrome_stream_is_not_held` (`printer_tests.rs`) — an ordinary `eprintln` and a `prompt_eprintln` issued under the same held session; only the notice reaches `err`.
- `an_editor_failure_reaches_the_user_while_the_permission_prompt_is_open` (`prompter_tests.rs`) — as described above, now asserting the exact notice on `err`.

PR #1139, opened 2026-09-08
refactor(md, term, cli): Move ANSI stream machinery into `jp_term`

JeanMertz:
`ansi` and `shade` move from `jp_md` to `jp_term`, along with the
region-background types they interpret, which land in a new
`jp_term::background`. Nothing about what they do is markdown: `segments`
tokenizes an escape stream, `AnsiState` tracks what a stream left active,
and `ShadedWriter` keeps a background showing across cursor rewrites and
split escapes. `jp_md` gains a `jp_term` dependency and imports them like
any other consumer.

The move puts them where the printer can reach them. `jp_printer` depends
on `jp_term` and deliberately not on `jp_md`, so shading a prompt meant
either pulling a markdown renderer into the output layer or writing the
state machine a second time.

`DefaultBackground`, `BackgroundFill`, and `line_fill` leave
`jp_md::format` for `jp_term::background`, so the types and the one place
that interprets them travel together. `AnsiState`'s methods widen from
`pub(crate)` to `pub` now that its callers are in another crate.


crates/jp_term/src/ansi.rs
```diff
@@ -1,8 +1,11 @@
...
 //!
-//! This module provides the escape sequences, state tracking, and visual width
-//! computation used by both the terminal renderer (`render.rs`) and the table
-//! formatter (`table.rs`).
+//! [`segments`] is the tokenizer everything else here is built on: it splits a
+//! byte stream into visible text and complete escape sequences.
+//! [`AnsiState`] tracks which attributes a stream has left active, so a writer
+//! can close them at a line break and re-open them on the next line.
+//! [`visual_width`] and [`advance_column`] measure text as the display lays it
+//! out, skipping escapes and advancing a tab to its stop.
```
JeanMertz: nit: For tabbed text such as `"a\tb"`, `visual_width` returns `3`, while `advance_column(0, ...)` returns `9`, as the existing tab test records. This summary attributes tab-stop handling to both helpers. Please distinguish them here so callers choosing an API for column padding don't assume `visual_width` models cursor movement.

JeanMertz: Taken. The summary was wrong, and it was wrong as of this commit: the pre-move version made no tab claim at all, so I introduced it while rewriting the paragraph to drop its references to `render.rs` and `table.rs`. Worse than a stale comment, it contradicted `visual_width`'s own doc eleven lines below, which says a tab counts as one column and sends you to `advance_column` when the cursor position matters.

Each helper gets its own sentence now, and the numbers you quoted are in the text rather than only in the test:

```rust
//! [`visual_width`] measures how many columns text occupies, skipping escapes.
//! [`advance_column`] answers a different question: where the cursor ends up,
//! which a tab or a carriage return moves somewhere the column count does not
//! predict.
//! `"a\tb"` occupies 3 columns and leaves the cursor at 9, so padding to a
//! fixed column has to go through `advance_column`.
```

That covers the case you named directly: a caller choosing an API for column padding reads the concrete divergence before it picks one.

Worth knowing while you are in this file: the same tab discrepancy exists a second time, between these two and `jp_term::width::display_width`, which has no notion of a tab either and now sits in the same crate after this move. `T-0fk64t3` records it, with a first step of testing both over the same inputs to find out whether the rest of their behaviour agrees. That is a consolidation question rather than a docs one, so it stays out of this PR.

PR #1140, opened 2026-09-08
fix(printer, cli): Shade every prompt with the reasoning background

JeanMertz:
A tool call made from a reasoning block sits inside that block's shading,
and its prompts now do too. The approval prompt and the tool's questions
were already shaded; the inline reply widget, reached by `e` to edit
arguments or `r` to skip with a reason, was not, and cut an unshaded hole
in the middle of the block.

The printer knows which background prompts are drawn against, and applies
it to every prompt writer it hands out. Whoever owns the region names it
once and no prompt site has to remember: `Printer::set_prompt_background`
replaces the copy `ToolPrompter` was keeping, and the shaded-writer
plumbing that lived beside it in `jp_cli` is gone. The owned prompt
writer, which is the one the reply widget takes, is covered by the same
code as the borrowed one rather than needing its own.

Closes: T-0fg8cjs


crates/jp_cli/src/cmd/query/tool/coordinator.rs
```diff
@@ -553,13 +553,14 @@ impl ToolCoordinator {
...
-        // shading, and a prompt is a visual row like any other (RFD 095). The
-        // prompter writes through the printer rather than through the
-        // renderer, so it has to be handed the background rather than finding
-        // it.
-        prompter.set_background(tool_renderer.current_region());
+        // shading, and a prompt is a visual row like any other (RFD 095). This
+        // is the one place holding both the renderer that owns the region and
+        // the printer that draws the prompts, and the region is per tool, so
+        // the read happens here rather than at the tool-call boundary.
+        printer.set_prompt_background(tool_renderer.current_region());
```
JeanMertz: blocking: Clear the shared prompt background when rendering leaves the reasoning region. A plausible sequence is a tool called from shaded reasoning, followed by a normal assistant message, then Ctrl+C during that message. This assignment leaves `Some(background)` on the shared `Printer`; the transition to message rendering never clears it, and `InterruptHandler::handle_streaming_interrupt` takes its prompt writer from that same printer. The interrupt menu and its reply editor therefore inherit the earlier tool's shading outside the region. Dropping a prompt writer resets the terminal background but does not clear this stored value, so the stale formatting state survives into subsequent prompts until another tool decision replaces it.

Tie this state to the region's lifetime, including the transition out of reasoning, and add a regression covering a shaded tool followed by an interrupt during unshaded message output. The printer test that explicitly calls `set_prompt_background(None)` does not exercise that lifecycle.

crates/jp_printer/src/printer.rs
```diff
@@ -619,10 +645,15 @@ impl Printer {
...
-        Box::new(OwnedPrinterWriter {
+        let writer = OwnedPrinterWriter {
             tx: self.tx.clone(),
             target: self.prompt_target(),
-            _suspension: self.begin_prompt_session(),
+        };
+        let suspension = self.begin_prompt_session();
+
+        Box::new(OwnedPromptWriter {
+            writer: Canvas::new(writer, self.prompt_background.lock().as_ref()),
```
JeanMertz: nit: Handle CRLF before routing the reply editor through `ShadedWriter`. Pressing `e` on a shaded tool with non-empty JSON arguments seeds the editor with pretty-printed, multi-line JSON. The pinned reedline implementation converts buffer newlines to `\r\n` (`PromptLines::new` calls `coerce_crlf`). `ShadedWriter::process_text` forwards the `\r` immediately, then inserts `\x1b[K` when it encounters the `\n`. For example, `first\r\nsecond` becomes `BG first\r ESC[K ESC[49m\n BG second`: the erase runs from column zero and deletes `first`. The editor consequently blanks each completed line of the arguments while retaining them in its input buffer.

This is a visible rendering defect contained to the prompt, rather than loss of the input buffer. Teach the shading writer to fill before the carriage return of a CRLF pair, including pairs split across writes, and add a multi-line owned-writer regression. The mock's single message write does not exercise the real editor's newline stream.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
JeanMertz: Taken, and reproduced before fixing. Confirmed in `c536bbd5`.

`set_prompt_background` had exactly one production caller, and the value it wrote came from `ToolRenderer::current_region`, which only moves at `set_region` and `complete`. Both of those are reached only from a `ToolCallPart::Start` event. Nothing hears about the transition out of reasoning, so the last tool call of a turn kept the printer coloured for the rest of the process, and `handler.rs:256` takes the interrupt menu's writer from that same printer.

The staleness predates this PR: `ToolPrompter::background` had the same lifetime, set from the same single call site. What changed is reach, from the three `ToolPrompter` methods to every prompt the printer hands out. That is what makes it this PR's problem.

## Where the clear landed

Not where I first sketched it. I had thought `TurnCoordinator` could own the whole thing and the `printer` parameter threaded through `resolve_tool_call_decision` could go away. It cannot: the per-tool region depends on `conversation.tools.<name>.style.joins_reasoning`, which only the tool-call path knows, and `ToolRenderer::complete` realigns `current_region` per tool so parallel calls in different regions render correctly. Reading it anywhere else would be reading it too early. The parameter is doing real work, so it stays.

What was missing was only the other end. A region is open exactly while the last chat response was reasoning, and `ChatRenderer` is what knows that, so it publishes it:

```rust
fn set_response_kind(&mut self, kind: Option<ContentKind>) {
    self.last_response_kind = kind;

    if self.flow == RenderFlow::Live {
        self.printer.set_prompt_background(self.open_region());
    }
}
```

All five assignments to `last_response_kind` route through it. A tool call still names its own region over the top, since only it can answer the per-tool question, but the next message takes back whatever was left open. Replays do not publish at all: a stored turn is rendered into a document and never prompted against, so its regions have no business colouring a live prompt.

One thing this also buys, which was not the point but follows: an interrupt taken *during* reasoning is shaded, which is what RFD 095 asks for and what it never did before.

## Tests

Two, in `chat_tests.rs`, because the lifecycle is the thing under test rather than the printer's mechanism. You were right that `clearing_the_background_unshades_later_prompts` cannot catch this; it calls the setter by hand.

`a_tool_calls_region_does_not_outlive_the_reasoning_it_came_from` is your scenario: reasoning, a tool call, the background named directly on the printer the way `resolve_tool_call_decision` names it, then an ordinary message, then a prompt. With the publish disabled it fails with exactly the symptom you described:

```
  left: "\u{1b}[48;5;236mInterrupted\u{1b}[49m"
 right: "Interrupted"
```

`a_message_after_reasoning_closes_the_region_for_later_prompts` pins both directions, so the open half cannot quietly regress into never shading anything.

crates/jp_printer/src/printer.rs
JeanMertz: Taken, and your trace is right at every step. Fixed in `983bd59d`.

I verified the reedline premise rather than taking it on faith: `Cargo.lock` pins the fork at `afcd57f3`, and `PromptLines::new` there calls `coerce_crlf` on `before_cursor`, `after_cursor` and `hint`, with the function documented as "Ensures input uses CRLF line endings. Needed for correct output in raw mode." So every buffer newline reaches the writer as `\r\n`, and `try_edit_arguments` seeds that buffer with `serde_json::to_string_pretty`, which is multi-line for any tool call worth editing.

I would put this above `nit:`, though not at blocking. It is contained to the prompt and plainly visible, so it is not gating by the severity rules. But this PR's headline claim is that the reply widget was cutting an unshaded hole in the block and no longer does, and on the common input it was rendering that path *worse* than before rather than better. Worth fixing here rather than after.

## The fix, and the trap next to it

`\r\n` is one line terminator, so the line ends before the carriage return:

```rust
b'\r' if bytes.get(i + 1) == Some(&b'\n') => {
    self.end_line(&text[start..i], "\r\n")?;
    i += 2;
    start = i;
}
```

A carriage return on its own still rewrites the row in place, which is what the progress line needs and what `carriage_return_rewrite_keeps_the_background_active` pins.

Worth recording why the obvious precedent is wrong here. `jp_md::format::apply_line_background` guards this exact hazard, in almost these words, by stripping the `\r`:

```rust
// A CRLF line's `\r` puts the cursor back at column 0, so a fill written
// after it overwrites the content instead of extending it. The newline
// below terminates the line on its own.
line.strip_suffix('\r').unwrap_or(line)
```

That works there because the markdown path writes to a cooked terminal and `ONLCR` supplies the carriage return. Copying it here would drop the `\r` on a path that is in raw mode precisely because reedline knows it has to, and would reintroduce the staircase `#1137` fixed one commit down this stack. Fill before the `\r` and keep it, as you said.

## Split pairs

Done, using the buffer that was already there for half-formed escapes. A trailing `\r` says nothing about which of the two readings applies until the next byte lands, so it is held until it does, and `finish` releases one whose `\n` never came.

## Tests

Four in `shade_tests.rs` and one in `printer_tests.rs`. Before the fix, all four of the CRLF cases fail with the output you predicted:

```
  left: "\u{1b}[48;5;236mab\r\u{1b}[K\u{1b}[49m\n..."
 right: "\u{1b}[48;5;236mab\u{1b}[K\u{1b}[49m\r\n..."
```

The `Column` fill fails the same way with spaces instead of the erase, which is the variant a host laying out its own sub-window renders.

`an_owned_prompt_writer_keeps_a_crlf_line_intact` is the owned-writer regression you asked for, and you were right that the mock's single message write would not have exercised it; the test writes a CRLF-terminated JSON body through `owned_prompt_writer` directly. The two split-write tests also assert that holding the `\r` stays invisible to the rewrite case, since that is the thing the hold could plausibly break.

crates/jp_cli/src/render/chat.rs
```diff
@@ -793,6 +793,42 @@ impl ChatRenderer {
...
+    /// Every change is published, so a prompt taken once the region has ended
+    /// is drawn without it.
+    fn set_response_kind(&mut self, kind: Option<ContentKind>) {
+        self.last_response_kind = kind;
+
+        // A replay builds a document out of stored events and is never
+        // prompted against; only the live renderer's region governs what a
+        // prompt shows.
+        if self.flow == RenderFlow::Live {
+            self.printer.set_prompt_background(self.open_region());
```
JeanMertz: blocking: Clear the published region when reasoning gives way to a structured response too. A plausible input is an Anthropic query with `--schema`, full reasoning display, and a configured reasoning background, followed by Ctrl+C while the JSON answer streams. The provider emits `Reasoning` followed by `Structured`, but `TurnView::render_chat_response_chunk` handles the latter with `self.chat.flush(); self.structured.render_chunk(resp)` (`turn_view.rs:184-186`). Neither call updates `last_response_kind` or clears the printer's background. The interrupt path also only flushes before acquiring its prompt writer, so the reasoning background survives into the interrupt menu and reply editor outside the reasoning region, including subsequent interrupts after Continue.

The message-transition regression covers the ordinary-text path, but not this dispatch around `ChatRenderer`. Explicitly end the prompt region at the structured-response boundary and add a regression through `TurnView`: reasoning, a structured chunk, then a prompt that must be unshaded. Avoid clearing on every `flush()`, since interrupting reasoning itself also flushes and should retain its shading.

JeanMertz: Taken, and correct at every link. Fixed in `ddf0bb62`.

The dispatch is what you quoted, and neither half reaches the publish. `ChatRenderer::flush` is `cancel_reasoning_timer` + `drain_buffer` + `emit_pending_separator(false)`, and `drain_buffer` routes through `handle_event` rather than `render_content`, so nothing on that path touches `last_response_kind`. There is no second chance either: `render_response` answers `ChatResponse::Structured { .. } => {}`, so even a direct call would be a no-op.

The input is more ordinary than your description allows. Reasoning-then-structured is the *normal* shape for a schema answer from a reasoning model, not a corner: Anthropic maps `TextDelta` to `Event::structured` when `is_structured` and `ThinkingDelta` to `Event::reasoning` (`anthropic.rs:2470`, `:2473`), and Cerebras, llama.cpp and Ollama literally flush reasoning and then emit structured in consecutive statements (`cerebras.rs:871-877`, `llamacpp.rs:265-271`, `ollama.rs:191-197`).

One detail worse than you wrote it. On Continue, `reset_preserving_region` saves `last_response_kind`, calls `reset()` (which publishes `None`), then restores it through `set_response_kind`, which publishes again. With the value stale at `Some(Reasoning)`, Continue actively re-asserts the region rather than merely failing to clear it.

## Where it landed

Your caveat about `flush()` is right, and I checked it rather than taking it: `TurnCoordinator::flush_renderer` is documented as "Call this before `Printer::flush_instant()` on interrupt" and forwards to `TurnView::flush` to `ChatRenderer::flush`. Clearing there would unshade an interrupt taken *during* reasoning, which is the one thing `c536bbd5` gained beyond the bug fix.

So the chat renderer gained a way to be told, and the structured branch tells it:

```rust
if resp.is_structured() {
    self.chat.flush();
    self.chat.leave_response_region();
    self.structured.render_chunk(resp);
}
```

`ContentKind` has no `Structured` variant and adding one would ripple into `flush_on_transition` and the separator logic for nothing: `last_response_kind` is only ever compared against `Some(ContentKind::Reasoning)`, so `None` and `Some(Message)` behave identically here. `None` is the honest value. `render_chat_response` delegates to the chunk path, so the complete-response route came along.

## Test

`a_structured_response_closes_the_region_for_later_prompts` in `turn_view_tests.rs`, through `TurnView` as you asked. It needed its own fixture: the existing `view_owing_separator` builds with `RenderFlow::Replay`, which never publishes at all.

Both directions are pinned, so the fix cannot regress into never shading anything. With the call commented out it fails exactly as you predicted:

```
  left: "\u{1b}[48;5;236mInterrupted\u{1b}[49m"
 right: "Interrupted"
```

On the label: this is the same class and the same containment as the coordinator comment above, so by the severity rules it is a suggestion rather than a gate. I am not arguing it down, for the reason that `c536bbd5`'s message claims "a prompt taken after the assistant moved on from reasoning is drawn without a background again", and that was false for structured output. A half-true claim is worse than no claim.

PR #1142, opened 2026-09-09
docs(rfd): Add RFD 107 discoverable tools

JeanMertz:
RFD 107 proposes a third tool enablement state. A `discoverable` tool sits in the catalogue and is permitted, but its definition is withheld from the model until a `search_tools` builtin finds it, at which point a `ConfigDelta` promotes it to offered for the rest of the conversation. This keeps a large MCP catalogue out of the context window without hiding capability from the user, and it works on every provider; only cache preservation across a discovery is Anthropic-specific.

The design turns on two boundaries. The single `Vec<ToolDefinition>` that today serves the provider array, the `describe_tools` docs map, executor eligibility, and MCP startup splits into a catalogue and an offered subset, because a discoverable tool has to be searchable and callable while staying out of the provider's tools array. Promotion crosses the boundary RFD 078 closed against builtin config mutation, so it travels a host-owned channel carrying only tool names rather than `outcome.config`, and is folded against the state at its own position so it cannot silently revert a disable the user approved earlier in the same cycle.

RFD 105 gains a paragraph recording why deferring tool search is not neutral: its initial directive offers every enabled tool and withdraws every disabled one, leaving discovery nowhere to operate until the third state exists. The relationship is recorded as `Extends` on both RFD 081, whose enablement model 107 widens, and RFD 078.

### DIFF
--- a/docs/rfd/107-discoverable-tools.md
+++ b/docs/rfd/107-discoverable-tools.md
@@ -189,11 +189,9 @@ Consumers divide along those two sets:
 | Executor eligibility   | permitted                                  | a promoted tool is callable in the same turn         |
 | MCP server startup     | permitted                                  | an unstarted server has no metadata to search        |
 
-The search row is stated in terms of availability rather than as "permitted
-minus offered" because the forced-tool exemption puts an `off` tool into both
-sets.
-Deriving search eligibility by subtraction would leak exactly the tool the user
-disabled.
+The search row names an availability rather than a subtraction because it says
+what the set *is*, and because it stays correct if the permitted set later grows
+a member that is neither offered nor discoverable.
 
 **Discovery saves context, not startup.** A discoverable MCP tool's server
 starts with the conversation and its schema is fetched up front, exactly as an
--- a/docs/rfd/107-discoverable-tools.md
+++ b/docs/rfd/107-discoverable-tools.md
@@ -1,13 +1,3 @@
-<!--
-  This template is a starting point, not a constraint. Delete sections that
-  don't apply, add sections that do, or restructure entirely. The only
-  requirement is the metadata header (Status, Authors, Date).
-
-  Use HTML comments like this one for draft-time notes and review markers.
-  They do not appear in the rendered output and can be removed when the RFD
-  advances to Discussion status.
--->
-
 # RFD 107: Discoverable Tools
 
 - **Status**: Discussion
@@ -18,10 +8,10 @@
 
 ## Summary
 
-Tool enablement grows a third state: a tool can be in the catalogue and
-permitted, but not offered to the model until the model asks for it.
-A `search_tools` builtin searches the catalogue and promotes matches to offered
-for the rest of the conversation.
+Tool enablement grows a third state: a tool can be permitted, but not offered to
+the model until the model asks for it.
+A `search_tools` builtin searches the permitted set and promotes matches to
+offered for the rest of the conversation.
 This keeps large tool catalogues out of the context window without hiding
 capability from the user.
 
@@ -60,7 +50,7 @@ A third value for a tool's enablement:
 [conversation.tools.fs_read_file]
 enable = "on"
 
-# In the catalogue, permitted, not offered until the model searches for it.
+# Permitted, but not offered until the model searches for it.
 [conversation.tools.github_list_workflow_runs]
 enable = "discoverable"
 
@@ -112,7 +102,7 @@ values:
 | `-t` / `--tool`    | → `on` | → `on`         | no-op   |
 | `-T` / `--no-tool` | no-op  | → `off`        | → `off` |
 
-A bulk `-t` therefore loads the whole catalogue into context, for every tool
+A bulk `-t` therefore loads every discoverable tool into context, for every tool
 whose `allow_toggle` permits it.
 That is the correct default: `-t` means "give the model everything," and a user
 who wants a discoverable tool exempt from it has `allow_toggle = "if_named"`
@@ -130,7 +120,7 @@ precedence.
 
 The consequence matters for the next section: because forcing can put an `off`
 tool into the resolved definition set, that set is not a statement about
-availability and must not be used as the search catalogue.
+availability and must not be used as the search surface.
 
 ### The state
 
@@ -148,7 +138,7 @@ combination unrepresentable:
 pub enum Availability {
     /// Not offered, not discoverable.
     Off,
-    /// In the catalogue, permitted, offered once discovered.
+    /// Permitted, offered once discovered.
     Discoverable,
     /// Offered up front.
     On,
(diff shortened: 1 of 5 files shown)


docs/rfd/107-discoverable-tools.md
```diff
@@ -0,0 +1,650 @@
...
+
+Promoting unconditionally would fold the promotion over the approved disable and
+silently undo it.
+
+A promotion is therefore evaluated against the folded state **at its own
+position** in the buffer, not against the cycle-start snapshot: a tool that is
+no longer `discoverable` by the time the promotion is folded is not promoted.
+The search response reports what committed, so a tool that was matched but not
+promoted is reported as found and unavailable rather than silently listed as
+usable.
```
JeanMertz: blocking: If the model issues parallel `search_tools("deploy")` and `search_tools("staging")` calls that both match a discoverable `deploy_staging`, the first promotion makes it `on`. At the second call's position it is no longer `discoverable`, so this rule skips promotion and reports it as unavailable, even though the next request offers it. That falsely tells the model not to use a tool it successfully discovered.

Can the response distinguish `on` (usable, no additional delta needed) from `off` (unavailable), rather than treating every skipped promotion as unavailable? Add an overlapping-search test alongside the approved-disable case.

docs/rfd/107-discoverable-tools.md
```diff
@@ -0,0 +1,650 @@
...
+
+Consumers divide along that line:
+
+| Consumer               | Set                     | Why                                                  |
+| ---------------------- | ----------------------- | ---------------------------------------------------- |
+| Provider `tools` array | offered                 | the context cost this RFD exists to avoid            |
+| `search_tools`         | catalogue minus offered | already-offered tools need no discovery              |
+| `describe_tools`       | catalogue               | a found tool must be explainable before it is called |
+| Executor eligibility   | catalogue               | a promoted tool is callable in the same turn         |
+| MCP server startup     | catalogue               | an unstarted server has no metadata to search        |
```
JeanMertz: nit: With `deploy.enable = "off"` and `jp q -u deploy`, the forced-tool exemption puts `deploy` in the offered set but not the catalogue. These rows therefore exclude its definition from the executor and, for an MCP tool, omit its server from startup. `TerminalExecutorSource::create` requires that definition to construct an executor, so the model receives a tool it cannot run, contrary to the forced-use contract above.

Execution and MCP startup need the forced-tool exception too, for example by consuming `catalogue ∪ offered`, while search remains restricted to discoverable tools. A forced-off MCP test would cover the distinction.

docs/rfd/107-discoverable-tools.md
```diff
@@ -0,0 +1,650 @@
...
+discoverable tool at the point the turn starts:
+
+```
+search_tools(query: string, limit?: integer = 10) -> [{ name, summary }]
+```
+
+Registration is decided once and holds for the conversation.
+Deciding it per request would withdraw the builtin the moment the last
+discoverable tool is promoted — a second change to the tools array on top of
+the promotion, and a tool disappearing under the model for no reason it can see.
```
JeanMertz: question: Is registration latched when a discoverable tool first appears, or frozen at conversation creation? A user continuing an existing conversation with `--cfg conversation.tools.deploy.enable=discoverable` needs `search_tools` even if the conversation started without discoverable tools. Conversely, resuming after the last promotion must retain the builtin despite the current config having no discoverable tools.

“At the point the turn starts” and “once ... for the conversation” leave these cases ambiguous, and phase 3 says “when the conversation opens.” Please specify how registration survives separate CLI invocations while still allowing discovery to be introduced into an existing conversation, and cover both cases in the plan.

docs/rfd/107-discoverable-tools.md
```diff
@@ -0,0 +1,650 @@
...
+The search response reports what committed, so a tool that was matched but not
+promoted is reported as found and unavailable rather than silently listed as
+usable.
+A response that promised a tool the fold left `off` would put the model in a
+state where its next call fails for reasons the transcript does not explain.
+
+### Provider behaviour
+
+Nothing in the mechanism is provider-specific.
+The tools array carries the offered set, which grows when a delta lands.
```
JeanMertz: question: How does the offered-only provider contract compose with RFD 105? For a local discoverable tool, its first promotion grows this array, but RFD 105 preserves the cache by keeping the array fixed and emitting `tool_addition`. It also defines its catalogue to include `off` tools, whereas this RFD's catalogue excludes them.

Please distinguish the search catalogue from RFD 105's stable provider catalogue and state the provider-boundary exception here and in the consumer table. The directive-enabled provider needs access to the stable definitions, not just the offered subset; otherwise the claimed cache preservation does not follow even without MCP churn.

docs/rfd/107-discoverable-tools.md
JeanMertz: Correct, and more than a nit — this reverts documented behaviour. `ctx.rs:229` is `if !cfg.is_enabled() && forced_tool != Some(name) { continue; }`, with a doc comment right above it explaining why: "Its backing server has to start regardless of the tool's enable state, or `tool_definitions` drops the tool again as unreachable and the forced choice cannot be satisfied." Routing MCP startup to a set that excludes `off` tools undoes that, and the failure is the silent kind. On the executor side `TerminalExecutorSource::create` opens with `self.definitions.get(&request.name)?`, so an absent definition means no executor at all.

Fixed differently than suggested. `catalogue ∪ offered` patches the two rows but leaves offered as a non-subset, and the `search_tools` row ("catalogue minus offered") would then only read correctly by accident. Instead the forced tool now widens the *definition*:

- **permitted set** = availability `discoverable` or `on`, plus the forced tool, minus anything locked off
- **offered set** = availability `on`, plus the forced tool — now a strict subset

Executor eligibility and MCP startup stay pointed at the permitted set and are correct with no further edits. The search row is restated as "availability `discoverable`, minus offered" rather than by subtraction, since subtraction would leak a forced-off tool into results — with a paragraph saying so. Added the forced-off MCP case to phase 2's tests.

docs/rfd/107-discoverable-tools.md
JeanMertz: Three phrasings, three different rules, and I wrote all of them: "at the point the turn starts" (per-turn), "decided once and holds for the conversation" (latched), "when the conversation opens" (at creation).

The sharper problem your question exposes is that "holds for the conversation" had nowhere to live. Each `jp query` is a fresh process that re-derives config from the stream, so a latch that isn't derivable from stream state doesn't survive to the next invocation.

Replaced all three with one rule: `search_tools` is offered once **any config position in the conversation's stream has held at least one tool at `discoverable`**, and stays offered from there. Derived from the stream like availability itself ([RFD 054]), monotone within a conversation, no new durable state. Both your cases fall out of it — a `--cfg` delta introducing a discoverable tool turns the builtin on next turn, and promoting every discoverable tool doesn't withdraw it, because an earlier position still holds one. The existing argument against per-request evaluation now explains the monotonicity instead of contradicting it.

The cost is stated rather than hidden: a conversation that once had discoverable tools and no longer does keeps a builtin whose searches come back empty. A wasted definition, not a wrong answer. Both cases are now phase 3 tests.

docs/rfd/107-discoverable-tools.md
JeanMertz: Agreed, and the input is ordinary rather than contrived — [RFD 078] folds by tool-call index precisely because a cycle can carry several calls, so two searches with overlapping matches needs no special setup.

The rule collapsed two different reasons a tool stops being `discoverable` at fold position into one outcome. Now three-way:

| Folded state at this position | Promotion | Reported as |
| --- | --- | --- |
| `discoverable` | applied | usable |
| `on` | no-op, none needed | usable |
| `off` | skipped | found, unavailable |

The prose calls out the `on` row explicitly with your `deploy` / `staging` / `deploy_staging` example, and says why collapsing it into `off` is wrong: it would tell the model a tool is unavailable in the same request that offers its definition. The closing sentence now names both failure directions rather than one — promising a tool the fold left `off`, and withholding one the fold left `on`.

Overlapping-search test added to phase 4 alongside the approved-disable case: two searches in one cycle matching the same tool report it usable both times, with one delta between them.

docs/rfd/107-discoverable-tools.md
JeanMertz: Both halves are right, and the first was a straight contradiction: "the tools array carries the offered set, which grows when a delta lands" sat directly above a table awarding directive-enabled Anthropic a cache-preserved "yes". An array that grows on promotion is the prefix change [RFD 105] exists to prevent, so that cell was unearned even before MCP churn.

"Provider behaviour" now describes two shapes instead of one. Without directives the offered set *is* the array and grows on promotion — tokens saved, cache lost, which is the trade every other provider makes. With [RFD 105] the array is not the offered set at all: 105 sends every tool it knows about in a fixed deferred array, and this RFD's offered set feeds `tool_addition` / `tool_removal` derivation instead, so the array stays byte-identical. Stated explicitly that the directive-enabled provider reads 105's array rather than this RFD's offered set.

On the naming collision: this RFD's set is renamed **permitted set** throughout, and "catalogue" is left to 105's wider meaning. A short paragraph records that the two are different extents and that 105's array is a superset, since an `off` tool belongs in 105's array (a later directive might name it) and never in this RFD's permitted set. The remaining uses of "catalogue" in the document are either generic English or explicit references to 105's array.

docs/rfd/107-discoverable-tools.md
```diff
@@ -0,0 +1,721 @@
...
+| `search_tools`         | availability `discoverable`, minus offered | an `off` tool never appears in search, forced or not |
+| `describe_tools`       | permitted                                  | a found tool must be explainable before it is called |
+| Executor eligibility   | permitted                                  | a promoted tool is callable in the same turn         |
+| MCP server startup     | permitted                                  | an unstarted server has no metadata to search        |
+
+The search row is stated in terms of availability rather than as "permitted
+minus offered" because the forced-tool exemption puts an `off` tool into both
+sets.
+Deriving search eligibility by subtraction would leak exactly the tool the user
+disabled.
```
JeanMertz: nit: With `deploy.enable = "off"` and `jp q -u deploy`, `deploy` belongs to both the permitted and offered sets, so `permitted minus offered` removes it. Subtraction does not leak the disabled tool into search. Under the resolved-set definitions above, that difference gives the same search candidates as filtering for `discoverable` and excluding offered tools.

The explicit availability predicate is fine, but please remove or correct this leakage explanation. It describes a failure that the shared forced-tool exemption already prevents.

JeanMertz: You're right, and the paragraph contradicted its own premise. With `D` = availability `discoverable`, `O` = `on`, `F` = the forced tool, and `D ∩ O = ∅` because `Availability` is three-valued:

```
permitted \ offered = (D ∪ O ∪ F) \ (O ∪ F) = D \ F
availability `discoverable`, minus offered  = D \ (O ∪ F) = D \ F
```

Identical. And for `deploy.enable = "off"` with `-u deploy`, `deploy` isn't in `D` at all, so neither form contains it. Being in both sets is exactly what makes subtraction cancel it out, which is the fact the paragraph cited before concluding the opposite.

It was never true, either. Against the definitions this replaced (permitted excluded the forced tool, offered included it), `(D ∪ O) \ (O ∪ F) = D \ F` as well. I wrote the justification while fixing the forced-tool row above and didn't check the arithmetic against either version.

Kept the availability predicate, replaced the justification with the honest one:

> The search row names an availability rather than a subtraction because it says what the set _is_, and because it stays correct if the permitted set later grows a member that is neither offered nor discoverable.

That second clause isn't hypothetical: the permitted set is where exemptions accumulate, and the forced tool became one in this review cycle. The table row's own "an `off` tool never appears in search, forced or not" was always sound, via `D` unforced and via `minus offered` forced.

Note for whoever reads the thread above: my reply on the forced-tool comment ends by repeating the leak claim. Same error, same correction.

PR #1143, opened 2026-09-09
fix(cli, rfd)!: Stop the workspace picker recording a selection

JeanMertz:
Answering the workspace picker recorded the choice as the session's active workspace, so a command run from outside every workspace asked once and then silently reused that answer. Standing in `~/scratch`, `jp c ls` prompted, and the next `jp c ls` did not: a listing had acquired durable user-global state on the way past, and nothing said so on later runs, because the foreign-root notice stays quiet when the launch cwd has no workspace of its own to compete.

RFD 087 never said the picker records, and says the opposite everywhere else. The cwd-vs-active prompt separates `c` (use it now) from `C` (use it and make it session-active), which is a distinction the picker was quietly deciding on the user's behalf. `jp w use` is described as the command that mutates session state. And the whole surface is meant to mirror `jp conversation`, where the picker leaves the session mapping alone and only `use`, `fork`, and `q` write it. The picker now resolves the run and nothing more, so a session that never ran `jp w use` is asked again, exactly like a conversation that was never `use`d. Resolution may still repair a recorded selection whose checkout is gone, but it never creates one: `picker` takes no session at all, so the invariant holds by construction rather than by review.

The bootstrap requirement declaration was not driving anything. `run_inner` returned early for `jp w` before reading it, leaving `WorkspaceCommand::workspace_requirement` unreachable and its `Resolve` variant dead behind that unreachable call. The declaration also described the wrong relationship: `jp w show <id>` reports *on* a workspace rather than running inside one, and `jp w use <id>` records a selection for later runs, so both resolve their own target against the pre-workspace `TargetEnv` and want no root selected for them. That is now a `Subject` requirement, and the startup dispatch matches on the requirement instead of on the command, so each arm performs exactly the pre-workspace work its commands asked for. `Resolve` had no client left once `jp w use` became `Subject`, and is dropped rather than kept for a hypothetical caller.

The targeting keyword for global recency becomes `recent` / `r`. `jp conversation` already spells this concept `recent` and reserves `newest` for the distinct one, so `latest` was a third word for something the grammar it borrows from names twice. RFD 087 was corrected to `recent` while this implementation was in flight, and merged without picking the change up.

BREAKING CHANGE: The `l` / `latest` workspace target is now `r` / `recent`

Spell `jp -w latest` as `jp -w recent`, and `jp w use l` as `jp w use r`. The old spellings are no longer keywords and fall through to fuzzy matching, so they will report no matching workspace rather than resolving. `jp w use help` lists the current grammar.

### DIFF
--- a/crates/jp_cli/src/bootstrap.rs
+++ b/crates/jp_cli/src/bootstrap.rs
@@ -46,11 +46,18 @@ pub(crate) enum WorkspaceRequirement {
     /// plugin child cwd, path parsing — simply do not run.
     None,
 
-    /// Resolve and validate a target root, without loading the conversation
-    /// index.
-    Resolve,
+    /// The workspace is the command's *subject* rather than its context.
+    ///
+    /// `jp w show <id>` reports on a workspace instead of running inside one,
+    /// and `jp w use <id>` records a selection for later runs to resolve.
+    /// Both resolve their own [`WorkspaceTarget`] against the pre-workspace
+    /// [`TargetEnv`], from outside every workspace and possibly to no workspace
+    /// at all, so selecting a root on their behalf would answer a question they
+    /// have not asked.
+    Subject,
 
-    /// Resolve, construct the [`Workspace`], and load the conversation index.
+    /// Resolve a root, construct the [`Workspace`], and load the conversation
+    /// index.
     Load,
 }
 
@@ -74,8 +81,7 @@ pub(crate) enum RootSource {
     /// w use`).
     SessionActive,
 
-    /// The interactive picker fallback, recorded as the session's new active
-    /// workspace.
+    /// The interactive picker fallback, resolving this run only.
     Picker,
 }
 
@@ -191,10 +197,14 @@ impl ExecutionContext {
 /// 5. Else the session-active workspace, while the workspace still has a live
 ///    checkout — recovering through surviving checkouts when the recorded one
 ///    is gone.
-/// 6. Else the picker, recorded as the new session-active workspace.
+/// 6. Else the picker, resolving this run only.
 ///
 /// Non-interactive runs ignore the session layer entirely (steps 2–3 and
 /// 5–6), so scripts never depend on hidden per-session state.
+///
+/// Resolution may *repair* the session's recorded selection when its checkout
+/// is gone, but never creates one: attaching a workspace to a session is `jp w
+/// use`'s job, or the `C` / `A` answers to the conflict prompt.
 pub(crate) fn resolve(
     target: Option<&WorkspaceTarget>,
     session: Option<&Session>,
@@ -261,7 +271,7 @@ fn source_for(target: &WorkspaceTarget) -> RootSource {
         WorkspaceTarget::Session
         | WorkspaceTarget::SessionPicker
         | WorkspaceTarget::Picker
-        | WorkspaceTarget::Latest
+        | WorkspaceTarget::Recent
         | WorkspaceTarget::Fuzzy(_) => RootSource::CliSelector,
         WorkspaceTarget::Cwd | WorkspaceTarget::Help => {
             unreachable!("resolved before source mapping")
@@ -311,7 +321,7 @@ fn ladder(env: &TargetEnv<'_>, session: &Session) -> Result<(Utf8PathBuf, RootSo
         }
 
         // Step 6: the picker.
-        (None, None) => picker(env, session),
+        (None, None) => picker(env),
     }
 }
 
@@ -531,21 +541,14 @@ fn apply_conflict_choice(
 
 /// The ladder's last step: pick from every known workspace.
 ///
-/// The choice is recorded as the session's new active workspace: an
-/// unrecordable choice would be re-made on every invocation, which is why the
-/// session layer requires a session identity at all.
-fn picker(env: &TargetEnv<'_>, session: &Session) -> Result<(Utf8PathBuf, RootSource)> {
+/// The choice resolves this run only.
+/// Taking no session identity is the point: a fallback the user was never asked
+/// to keep cannot quietly become the answer to every later run from the same
+/// terminal.
+fn picker(env: &TargetEnv<'_>) -> Result<(Utf8PathBuf, RootSource)> {
     let Some(selected) = workspace_target::pick_known_workspace(env, "Select a workspace")? else {
         return Err(no_workspace_error(env));
     };
-
-    if let Some(id) = &selected.id
-        && let Err(error) = env
-            .store
-            .record_selection(session, id, &selected.root, Utc::now())
-    {
-        warn!(%error, "Failed to record the workspace selection.");
-    }
 
     Ok((selected.root, RootSource::Picker))
 }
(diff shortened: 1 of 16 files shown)


crates/jp_cli/src/bootstrap.rs
```diff
@@ -531,22 +541,15 @@ fn apply_conflict_choice(
...
 /// The ladder's last step: pick from every known workspace.
 ///
-/// The choice is recorded as the session's new active workspace: an
-/// unrecordable choice would be re-made on every invocation, which is why the
-/// session layer requires a session identity at all.
-fn picker(env: &TargetEnv<'_>, session: &Session) -> Result<(Utf8PathBuf, RootSource)> {
+/// The choice resolves this run only.
+/// Taking no session identity is the point: a fallback the user was never asked
+/// to keep cannot quietly become the answer to every later run from the same
+/// terminal.
```
JeanMertz: question: With `picker` taking no session, why does step 6 still require one?

`resolve_from` only enters `ladder` for `(Some(session), true)` (L234-236). An interactive run with no session identity still errors from outside every workspace, and `no_workspace_error` (L567-571) tells the user to set `$JP_SESSION` "to select one". The reason for that gate was the doc this PR deletes: "an unrecordable choice would be re-made on every invocation, which is why the session layer requires a session identity at all". Re-making the choice on every invocation is now the intended behaviour, so that reason no longer holds.

The `ladder` doc at L291-292 is also stale: "a choice that cannot be prompted for or recorded is not made at all". The step-6 choice is no longer recorded, and it is still made.

Two consistent options:
- Let identity-less interactive runs reach `picker(env)`. This needs a separate branch in `resolve_from`, since `ladder` needs the session for steps 2-5. Then drop the session hint from the error.
- Keep the gate, give it a reason that doesn't depend on recording, and fix L291-292.

Which one did you intend?

docs/rfd/087-session-scoped-active-workspace.md
```diff
@@ -443,6 +453,13 @@ Interactive ladder, in order:
 5. Else use the session-active workspace when live.
 6. Else picker.
 
+The picker at step 6 resolves the run only; it does not record a selection.
+Resolution may **repair** a recorded selection whose checkout is gone, but never
+creates one — attaching a workspace to a session is `jp w use`'s job, or the
+`C` / `A` answers below.
+A session that never selected one is asked again on the next run, mirroring the
+conversation picker, which likewise leaves the session mapping untouched.
```
JeanMertz: nit: This paragraph contradicts "No session identity" further down (L514-520, not in the diff). That section says `jp q` without an identity "does not fall back to a non-persisted one-shot picker: a choice that cannot be recorded would have to be re-made on every invocation". After this change, step 6 *is* a non-persisted one-shot picker, and re-asking is described here as the intended behaviour.

Whatever you decide on the `bootstrap.rs` question about gating step 6, update that bullet's rationale in this PR so the RFD states one position.

crates/jp_cli/src/cmd/workspace/use_.rs
```diff
@@ -173,6 +177,27 @@ impl Use {
...
+    }
+
+    match target::resolve(&WorkspaceTarget::Session, env) {
+        Ok(resolved) => Ok(resolved),
+        Err(error) => {
+            debug!(%error, "No previous workspace to return to; picking instead.");
+            target::resolve(&WorkspaceTarget::Picker, env)
+        }
+    }
+}
```
JeanMertz: nit: Bare `jp w use` now means `s` with a picker fallback; before this PR it always opened the picker. That is a user-visible behaviour change, but the PR description and the `BREAKING CHANGE` footer only mention `latest` → `recent`. The RFD 087 `jp w use <target>` bullet (L315-324) also doesn't cover the bare form, even though RFD 112 L185-190 describes it as existing behaviour.

Concrete case: a user who types `jp w use` to open the picker, in a session with two workspaces in its history, gets switched to the other one with no prompt. The output says "Switched…", so they will see it. Still, it belongs in the squash message and in the RFD 087 bullet (for example: "Bare `jp w use` returns to `s`, falling back to `?`, mirroring bare `jp c use`").

crates/jp_cli/src/cmd/workspace/use_tests.rs
```diff
@@ -364,6 +364,55 @@ fn use_cwd_clears_the_selection() {
     assert!(env.store.load(&session).is_none());
 }
 
+// Bare `jp w use` is the `cd -` of workspaces, mirroring bare `jp c use`: it
+// returns to where the session came from rather than asking again.
+#[test]
+fn bare_use_returns_to_the_previous_workspace() {
```
JeanMertz: nit: This covers the `s` branch of `resolve_target`, but nothing covers the fallback. You can test it without a terminal. With no session history and an empty registry, bare `use` should fail with "No known workspaces" (the `Picker` arm), not "No previously active workspace recorded" (the `Session` arm). That assertion would catch a regression where the `Session` error is returned instead of swallowed, or where the fallback target changes.

crates/jp_cli/src/lib.rs
```diff
@@ -591,57 +591,60 @@ fn build_printer(globals: &Globals, format: OutputFormat) -> Printer {
...
-            .map_err(Into::into);
-    }
+        // `jp w` takes a workspace as its subject: it resolves its own target
+        // against the pre-workspace `TargetEnv` — from outside every
+        // workspace, possibly to no workspace at all — so it needs a session
+        // identity but never a selected root, and never a `Ctx`.
+        WorkspaceRequirement::Subject => {
+            let Commands::Workspace(args) = cli.command else {
+                unreachable!("`Subject` is declared by `jp workspace` alone");
+            };
```
JeanMertz: nit: The comment says "no command can reach a path another arm owns". That is only enforced at runtime. `workspace_requirement` and these two `unreachable!` arms are a pair the compiler doesn't check against each other. If someone adds a command that declares `None` or `Subject`, it compiles and then panics on first run. The old code had the same weakness for `Init`, so this isn't new, but the comment now claims more than the code guarantees.

`workspace_requirement` is now a pure function of the top-level variant (`Workspace(_) => Subject` no longer looks at the subcommand), so it adds nothing the variant doesn't already say. Two options:
- Match on `cli.command` here (`Init | Provider` / `Workspace` / rest). That gives exhaustiveness at compile time. Keep `workspace_requirement` as the declaration, with a `debug_assert_eq!` or a unit test that ties each arm to its declared requirement.
- Keep this shape, and soften the comment to say the pairing is checked by `unreachable!`.

docs/rfd/112-hierarchy-targets-and-explicit-selection-clearing.md
```diff
@@ -0,0 +1,315 @@
+# RFD 112: Hierarchy Targets and Explicit Selection Clearing
```
JeanMertz: question: Should RFD 112 and the RFD 039 "Depth is unbounded" / "Path-shaped targeting" sections ship in this PR? The description covers the picker fix, the `Subject` requirement, and the `recent` rename. It doesn't mention a new RFD, the RFD 039 additions, or the bare `jp w use` change. Two of the three commit subjects are "review feedback" and "fixes", so squashing with the PR body would leave these changes unexplained in history.

If they belong together, a line in the description is enough. Otherwise, putting the RFDs in their own PR would let RFD 112 stay in "Discussion" without being tied to a `fix!` release.

PR #1144, opened 2026-09-09
feat(llm, anthropic, config): Serve `flex` as a batch request

JeanMertz:
`jp q --tier flex` against `anthropic` used to fail, because Anthropic sells no flex tier. It now runs the request through their Message Batches API, which carries the same 50% discount other providers' flex tiers offer. Nothing streams: the answer arrives complete, typically after tens of minutes, and the status row reads `⏱ Waiting… 240.0s (waiting on the batch)` until it does. A batched response is replayed as the events a streamed one produces, so tool calls, chaining past the token ceiling, and thinking-block repair behave the same either way.

Interrupting the turn cancels the batch, so an abandoned request stops billing. Waiting longer than
`providers.llm.anthropic.batch_max_wait_secs` (one hour by default) gives up and reports the batch id, leaving it running so the answer can still be fetched by hand; `0` waits the full 24 hours Anthropic keeps a batch alive. `batch_poll_interval_secs` sets how often the status is checked.

Inquiries no longer inherit the assistant's service tier. An inquiry blocks the tool call that raised it, so it cannot pay for a discount in latency; setting
`conversation.inquiry.assistant.model.parameters.service_tier` on the inquiry still applies.

### DIFF
--- a/crates/jp_cli/src/cmd/conversation/summarize.rs
+++ b/crates/jp_cli/src/cmd/conversation/summarize.rs
@@ -266,7 +266,7 @@ fn summarize_events(events: Vec<Event>) -> StreamOutcome {
             // else consumes them on this path, so keep them for the rebuild.
             Event::Patch(mut p) => patches.append(&mut p),
             // `KeepAlive` is a liveness signal.
-            Event::KeepAlive => {}
+            Event::KeepAlive { .. } => {}
         }
     }
 
--- a/crates/jp_cli/src/cmd/query/turn/coordinator.rs
+++ b/crates/jp_cli/src/cmd/query/turn/coordinator.rs
@@ -357,7 +357,7 @@ impl TurnCoordinator {
 
             // Patch is handled by the caller before reaching here; KeepAlive is
             // a liveness signal with nothing to record or render.
-            Event::Patch(_) | Event::KeepAlive => HandleEventOutcome::new(Action::Continue),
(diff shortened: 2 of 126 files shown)


JeanMertz:
Closing unmerged. The implementation works; the design does not.

A batch holding one request uses none of what the batch API is for.
Amortization across many requests is the whole mechanism, and with a single
request you pay the queue wait for nothing in return. Measured here:
`jp q -n --tier flex "Hello there, is this working?"` sat past 2078s. Queue
latency has no relationship to prompt size.

Three further problems, in order of how hard they are to design around:

- **Naming it `flex` is a one-way door.** Anthropic may ship a real `flex`
  tier. `--tier flex` persists as a conversation config delta, so the wrong
  meaning would already be on disk in users' workspaces.
- **The cache economics invert.** The default 5-minute cache TTL cannot
  survive the wait, so each tool round re-sends the conversation as a cache
  write rather than a read. With default settings, batched agentic work costs
  *more* than streaming.
- **It is per request, not per turn.** Three tool calls is three queue waits.

`--tier batch` was considered as a more faithful rename and rejected: batch is
not a tier. A tier means "same request, different capacity"; batch changes the
interaction model, and the two are independent on Anthropic's wire anyway.

The shape that would work is `jp batch` as its own command, batching multiple
pre-built queries, where the discount is a side effect rather than the
motivation. Findings and recovery notes: T-0g1w1mw.

PR #1146, opened 2026-09-09
fix(cli, llm): Correct the token estimate, stop gating summaries

JeanMertz:
Summarizing a range that does not fit the model's context window now reports the provider's own rejection, carrying the request's real token count. Running `jp conversation compact --turn=2..50 --summary` on a large conversation passed a local character-based check and then failed with a bare stream error; it now reports a summarization failure naming the model, the actual token count against the window, and the two ways to get under it.

The pre-flight check is removed rather than made accurate. A token count request ships the same payload as the request it would guard, so it saves nothing when the range is too large and costs an extra upload every time the range fits. The provider's message is also more precise than anything measured locally, and the error it raises was already classified as `ContextWindowExceeded`.

`CHARS_PER_TOKEN` drops from 3 to 2. A 4,220,150-byte Anthropic request body counted 1,317,976 input tokens, which is roughly 1.9-2.0 chars per token once JSON framing is backed out, so the old ratio overstated the available window by about half. This tightens the truncation budget for title generation and tool inquiries too, since they share the same estimator.

One detail is no longer reported: the failure message does not name the turn range, which the caller supplied on the command line and the provider's token counts describe more usefully.

### DIFF
--- a/crates/jp_cli/src/cmd/conversation/summarize.rs
+++ b/crates/jp_cli/src/cmd/conversation/summarize.rs
@@ -10,13 +10,12 @@ use jp_conversation::{
     thread::ThreadBuilder,
 };
 use jp_llm::{
-    Provider,
+    Provider, StreamErrorKind,
     event::{Event, EventPatch, FinishReason, record_patches},
     event_builder::EventBuilder,
     model::ModelDetails,
     provider,
     retry::{RetryConfig, collect_with_retry},
-    window,
 };
 use tracing::debug;
 
@@ -86,25 +85,6 @@ pub async fn generate_summary(
     let provider = provider::get_provider(model_id.provider, &app_cfg.providers.llm)?;
     let model_details = provider.model_details(&model_id.name).await?;
 
-    // The instructions ride in the system prompt and the request in its own
-    // turn; both share the window with the range.
-    let overhead =
-        window::estimate_overhead_chars(Some(instructions), &[], &[], &[]) + user_message.len();
-
-    if let Some(overflow) = window_overflow(&stream, model_details.context_window, overhead) {
-        return Err(Error::Summarize {
-            model: model_id.to_string(),
-            // Stored indices are 0-based; turn numbers shown to the user are
-            // 1-based.
-            reason: format!(
-                "turns {}..{} {overflow}; compact a smaller range (`--from`/`--to`) or summarize \
-                 with a larger-window model",
-                range_from + 1,
-                range_to + 1,
-            ),
-        });
-    }
-
     summarize_stream(
         provider.as_ref(),
         &model_details,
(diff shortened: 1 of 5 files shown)


crates/jp_cli/src/cmd/conversation/summarize.rs
```diff
@@ -86,25 +85,6 @@ pub async fn generate_summary(
     let provider = provider::get_provider(model_id.provider, &app_cfg.providers.llm)?;
     let model_details = provider.model_details(&model_id.name).await?;
 
-    // The instructions ride in the system prompt and the request in its own
-    // turn; both share the window with the range.
-    let overhead =
-        window::estimate_overhead_chars(Some(instructions), &[], &[], &[]) + user_message.len();
-
-    if let Some(overflow) = window_overflow(&stream, model_details.context_window, overhead) {
```
JeanMertz: blocking: An oversized multi-turn range summarized with an OpenAI model can succeed after silently dropping part of the input. `openai::create_request` (`crates/jp_llm/src/provider/openai.rs:585`) unconditionally sets `truncation: Some(types::Truncation::Auto)`, which lets OpenAI discard items from the beginning of the conversation rather than reject the request. The resulting `Completed` response passes `summarize_events`, and `build_compaction_for_range` records that tail-only summary against the entire selected range. Future requests then omit turns the summary never covered.

Make summary requests explicitly non-truncating (for OpenAI, `Truncation::Disabled`) before relying on provider rejection, and add a request-construction regression test. The mock rejection test does not exercise this provider policy.

JeanMertz: Verified and fixed in `fba097e9`, with one change of scope from what you proposed.

The chain checks out end to end. `openai.rs:585` sets `truncation: Some(types::Truncation::Auto)` unconditionally (introduced in `c223cab`, no recorded rationale). `Truncation::Auto`'s own docs say the model "will truncate the response to fit the context window by dropping input items **in the middle** of the conversation" — the middle, not the beginning, which for a summary is worse: it can remove content from inside the very span the summary claims to cover, and can orphan a tool call from its response. `Truncation::Disabled` is documented as "the request will fail with a 400 error", which is exactly the behavior this PR's design assumes every provider has.

**This PR owns the regression rather than merely inheriting it.** I checked whether the deleted gate ever protected OpenAI, since that decides fix-vs-defer: it did. OpenAI models carry a static `context_window` (`openai.rs:634–1287`, 128k–1,050k), so `window_overflow`'s `let context_window = context_window?` never short-circuited for them. For a 400k model the old gate rejected above ~1,080,000 chars while true capacity is ~800,000, so the false-accept band was already exposed — but grossly oversized ranges were caught. Removing the gate exposed the whole domain.

And the damage both spreads and hides: `build_compaction_for_range` (`compact.rs:507–530`) attaches `SummaryPolicy::generated(text)` to the full `range.from_turn..range.to_turn` with no check on coverage, the command prints success, and every later request projects that summary in place of turns it never read.

**Where I amended.** The fix can't be scoped to "summary requests" as written, because `create_request(model, query)` has no notion of request purpose and `chat_completion_stream` is a trait method with no room for one. So `ChatQuery` gains a `truncation: Truncation` field (`Allowed` / `Forbidden`), summarization asks for `Forbidden`, and OpenAI maps it. That's ~50 mechanical call-site updates, all compiler-caught.

I deliberately did **not** flip the default to `Disabled` globally, tempting as it looks. `truncate_to_fit` has exactly two callers (`title.rs:151`, `inquiry.rs:254`) and the main query path (`turn_loop.rs:304`) is not among them — it relies on OpenAI's auto-truncation to keep long conversations answerable at all. A global flip converts working sessions into hard 400s with no fallback. Defensible, arguably even correct, but a separate behavior decision.

**On the test.** Added `provider::openai::tests::truncation`, asserting the serialized `truncation` field is `"auto"` for `Allowed` and `"disabled"` for `Forbidden`. I confirmed it actually catches the regression by reverting the match arm to always emit `Auto` and watching it fail (`left: Some(String("auto")), right: Some(String("disabled"))`) before restoring the fix.

**Two related things picked up along the way.** `summarize_events`' doc comment justified rejecting every non-`Completed` reason on the grounds that a truncated response would be "stored as the summary and replace the turns it was meant to stand in for" — true for *output* truncation, but defeated by the *input* case it doesn't cover. I extended the comment to say which half is handled where. I also checked whether OpenAI is unique here: Anthropic and Google both reject (`openai.rs:1573`, `google_tests.rs:1099`, `error_tests.rs:583`), and llama.cpp sets no truncation parameter at all — its overflow behavior is a server-side deployment concern JP doesn't opt into. So OpenAI is the only provider where JP itself asked for silent truncation.

PR #1149, opened 2026-09-10
refactor(storage): Generalize conversation locking into a resource locker

JeanMertz:
Conversation locks were the only cross-process locking in the tree, and
their mechanics were welded to their policy: OS advisory locks, guard
lifetimes, holder-info serialization, and the conversation-id naming all
lived in one module. A second consumer could not reuse the locking
without also inheriting conversations.

`ResourceLocker` is that mechanism on its own: acquire a named resource,
release on drop, probe whether it is held, and read opaque holder info.
Three implementations back it. The filesystem locker keeps one
`<resource>.lock` per resource under a directory it owns, held through
`flock` or `LockFileEx`. The in-memory locker serializes within a
process, and blocking waiters park on a condition variable. The null
locker grants everything, for flows that deliberately run without
exclusion.

`lock.rs` becomes the domain layer above it: conversation ids as
resource names, `LockInfo` as the holder diagnostics, and the
user-versus-workspace placement policy. The in-memory storage backend
drops its bespoke lock set and guard in favour of the shared locker, so
both backends now produce guards the same way.

One mechanic is worth naming, because it is a trap. Removing the lock
file on guard drop is never fully sound: the OS lock releases when the
handle closes, so a contender that opens the path before the unlink
locks an inode about to disappear while the next acquisition locks a
fresh file at the same path. A non-blocking acquisition bounds that to
the close/unlink window; a blocking waiter is already parked on the
doomed inode when the unlink lands, so `lock` refuses to run with
`with_remove_on_drop`. Conversation lock files double as presence
markers for the orphan scan, so they take that option and inherit the
residual window; RFD 106 names it the release-then-unlink defect and
T-0gnv6ez tracks the fix.

### DIFF
--- a/crates/jp_storage/src/resource_lock.rs
+++ b/crates/jp_storage/src/resource_lock.rs
@@ -1,0 +1,470 @@
+//! Cross-process resource locking.
+//!
+//! [`ResourceLocker`] serializes access to named resources.
+//! It is the locking *primitive*: acquire, release on drop, probe, and opaque
+//! holder info.
+//! Domain policies — which resource names exist, what the holder info
+//! contains, where lock files live — belong to the consumer.
+//!
+//! Three implementations:
+//!
+//! - [`FsResourceLocker`] — one `<resource>.lock` file per resource, held via
+//!   OS advisory locks (`flock` on Unix, `LockFileEx` on Windows).
+//! - [`InMemoryResourceLocker`] — in-process locking for tests and
+//!   memory-backed setups.
+//! - [`NullResourceLocker`] — every acquisition succeeds; for flows that
+//!   deliberately run without cross-process exclusion.
+
+use std::{
+    collections::HashMap,
+    fmt::Debug,
+    fs::{File, OpenOptions},
+    io::{self, Read as _, Seek as _, Write as _},
+    sync::{Arc, Condvar, Mutex},
+};
+
+use camino::Utf8PathBuf;
+
+/// Failure to acquire or inspect a resource lock.
+#[derive(Debug, thiserror::Error)]
+#[error("failed to lock resource {resource:?}: {source}")]
+pub struct LockError {
+    /// The resource whose lock operation failed.
+    pub resource: String,
+
+    #[source]
+    pub source: io::Error,
+}
+
+impl LockError {
+    fn new(resource: &str, source: io::Error) -> Self {
+        Self {
+            resource: resource.to_owned(),
+            source,
+        }
+    }
+}
+
+/// Serializes access to named resources across processes.
+pub trait ResourceLocker: Send + Sync + Debug {
+    /// Attempt to acquire the resource's exclusive lock without blocking.
+    ///
+    /// Returns `Ok(None)` when another holder has it.
+    /// `info` is opaque holder metadata recorded best-effort at acquisition,
+    /// readable through [`Self::holder_info`] while the lock file (or in-memory
+    /// entry) exists.
+    /// `None` records nothing, and clears anything an earlier holder left.
+    fn try_lock(
+        &self,
+        resource: &str,
+        info: Option<&str>,
+    ) -> Result<Option<Box<dyn ResourceGuard>>, LockError>;
+
+    /// Block until the resource's exclusive lock is acquired.
+    fn lock(&self, resource: &str, info: Option<&str>)
+    -> Result<Box<dyn ResourceGuard>, LockError>;
+
+    /// Whether another holder currently has the resource.
+    ///
+    /// A non-destructive probe: no lock file is created or removed.
+    fn is_held(&self, resource: &str) -> bool;
+
+    /// Read the info recorded by the most recent acquisition (best-effort).
+    ///
+    /// Not proof that anyone holds the resource: a file-based locker that keeps
+    /// its lock files reports the last holder's info after that holder's guard
+    /// dropped.
+    /// Pair this with [`Self::is_held`] when the current holder is what
+    /// matters.
+    fn holder_info(&self, resource: &str) -> Option<String>;
+}
+
+/// A held resource lock.
+/// Released on drop.
+pub trait ResourceGuard: Send + Sync + Debug {}
+
+/// File-based locking: one `<resource>.lock` file per resource under a fixed
+/// directory, held via OS advisory locks.
+///
+/// Killing the holding process releases the OS lock automatically; the lock
+/// file may remain, but a leftover file does not block the next acquisition.
+#[derive(Debug, Clone)]
+pub struct FsResourceLocker {
+    dir: Utf8PathBuf,
+
+    /// Remove the lock file when the guard drops.
+    ///
+    /// Unlinking is never fully sound.
+    /// The OS lock releases when the handle closes, so a contender that opens
+    /// the path in the window before the unlink takes a lock on an inode about
+    /// to disappear, while the next acquisition creates and locks a fresh file
+    /// at the same path: two holders at once.
+    /// RFD 106 calls this the release-then-unlink defect.
+    ///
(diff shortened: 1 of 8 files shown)


crates/jp_storage/src/resource_lock.rs
```diff
@@ -0,0 +1,448 @@
...
+    }
+}
+
+impl Drop for FsResourceGuard {
+    fn drop(&mut self) {
+        // Drop the file handle first to release the OS lock.
+        self.file.take();
+
+        if self.remove_on_drop {
+            let _err = std::fs::remove_file(&self.path);
```
JeanMertz: blocking: Concurrent JP commands targeting the same conversation can obtain overlapping guards even when every acquisition uses `try_lock`. On Unix, A can close its handle at line 232, B can open and lock that inode, A can unlink it here, and C can create and lock a replacement file. B and C both report success and can write the same conversation without exclusion.

This race is inherited from `ConversationFileLock`, but the new reusable policy explicitly claims `try_lock` makes removal sound. It does not: a non-blocking contender can also be descheduled between `open` and `flock`, so simply unlinking before closing is insufficient.

Prefer stable lock files and adjust the conversation orphan cleanup so it does not unlink them (`cleanup_stale_files` also removes them), or introduce a coordinated acquisition/deletion protocol. Add a regression test that controls the open/acquire/drop interleaving; the sequential removal test does not exercise this case.

crates/jp_storage/src/resource_lock.rs
```diff
@@ -0,0 +1,448 @@
...
+
+    fn lock(
+        &self,
+        resource: &str,
+        info: Option<&str>,
+    ) -> Result<Box<dyn ResourceGuard>, LockError> {
+        debug_assert!(
+            !self.remove_on_drop,
+            "blocking lock with remove_on_drop races on the unlinked inode"
+        );
```
JeanMertz: theoretical: The documented restriction on `with_remove_on_drop().lock(...)` disappears in release builds because this is a `debug_assert!`. A blocking waiter can then acquire an unlinked inode while another caller locks the replacement path, violating exclusion. I found no production caller using this combination; conversation locking uses `try_lock`.

If this option remains, enforce the restriction in release builds too, preferably by returning `LockError` before opening the file, or by making the unsupported combination unavailable through the API. Add a test for rejection that also runs with release assertions disabled.

crates/jp_storage/src/resource_lock.rs
```diff
@@ -0,0 +1,448 @@
...
+        let mut guard = FsResourceGuard {
+            file: Some(file),
+            path,
+            remove_on_drop: self.remove_on_drop,
+        };
+
+        // Holder info is purely diagnostic; failing to record it must not
+        // fail the acquisition.
+        if let Some(info) = info {
+            let _err = guard.write_info(info);
```
JeanMertz: question: Should an acquisition with `info: None` clear metadata from the previous acquisition? With the default filesystem locker, acquiring `"res"` with `Some("holder A")`, dropping it, and acquiring it with `None` leaves `holder_info("res")` reporting `"holder A"` while the new guard is held. The in-memory implementation returns `None` for that sequence.

Conversation locking supplies metadata, so this is about the new trait's contract rather than an observed conversation regression. If `None` means the current holder has no metadata, truncate the file best-effort on that path too and extend the shared tests with this reacquisition sequence. Otherwise, document that `None` preserves earlier metadata for filesystem lockers.

crates/jp_storage/src/resource_lock.rs
JeanMertz: The mechanism is right and the claim it contradicts was mine, so the docs are fixed here. The behavioral fix is going to a separate change, and I don't think this gates the merge.

On the claim: `try_lock` buys something narrower than "sound". A blocking waiter is *guaranteed* to be parked on the doomed inode when the unlink lands; a non-blocking contender only loses if it is descheduled inside the close/unlink gap. Deterministic versus racy, not sound versus unsound. The `remove_on_drop` doc, `with_remove_on_drop`, the `try_lock_conversation` comment, and the commit message now say that, and name it as RFD 106's release-then-unlink defect.

On scope: the PR does not widen the exposure. The deleted `ConversationFileLock::drop` did the identical close-then-unlink, and the same hazard exists independently in `cleanup_stale_files` → `list_orphaned_lock_files`, which unlinks lock files while another process may be between its `open` and its `flock`. RFD 106 already ruled on it — "The release-before-unlink defect still affects write exclusion on *existing* conversations; that is a pre-existing bug, out of scope here" — and designs the allocator lock around a stable inode for exactly this reason. Gating a behavior-preserving refactor on it would reverse that decision, and the fix you describe changes presence-marker semantics across `is_conversation_locked`, `lock_file_path`, and `cleanup_stale_files`.

Filed as `T-0gnv6ez`: both unlink sites, the presence-marker question that has to be answered before the files can persist, and your point that a regression test has to control the interleaving rather than run the steps in order.

crates/jp_storage/src/resource_lock.rs
JeanMertz: Taken, with a bit more than the truncate.

`None` now truncates: `guard()` calls `write_info(info.unwrap_or(""))`, and `holder_info`'s empty-buffer check turns that into `None`, so both lockers agree on your sequence. `test_holder_info_roundtrip` gained the drop-and-reacquire step; I checked it fails on the `fs` locker without the fix (`left: Some("{\"pid\":42}")`, `right: None`).

The truncate only closes half of it, though. Even after it, a file locker without `remove_on_drop` still reports the last holder's info once that guard has dropped, where the in-memory locker reports `None` — same divergence, one step later. That one is inherent to a lock file that outlives its holder, so it is documented rather than fixed: `ResourceLocker::holder_info` now says the info belongs to the most recent acquisition, is not proof anyone holds the resource, and should be paired with `is_held` when the current holder is what matters.

No conversation regression either way: `try_lock_conversation` always passes `Some` (serializing a `LockInfo` cannot fail) and removes the file on drop.

crates/jp_storage/src/resource_lock.rs
JeanMertz: Agreed, and it stops being theoretical sooner than the "no production caller" reading suggests. RFD 106 specifies the ID allocator lock as blocking (`ALLOCATOR_LOCK_WAIT`, one second) on a lock file that must never be unlinked. Once that lands, a removing/`try_lock` locker and a stable/`lock()` locker both exist in production and the illegal combination is one call away.

`lock()` now returns `Err(LockError)` when `remove_on_drop` is set, before the file is opened, so there is no debug/release split left. `test_fs_blocking_lock_refused_when_removing_on_drop` asserts the error and that no lock file was created.

I skipped the other two parts of the suggestion. Making the combination unrepresentable needs a second locker type, which is more API surface than the hazard is worth for one option. And a release-assertions test variant is moot once the check is a plain `Err` return rather than an assertion.

crates/jp_storage/src/resource_lock.rs
JeanMertz: The explicit scope decision in RFD 106, the follow-up in T-0gnv6ez, and the removal of the claim that `try_lock` makes unlinking sound address my reason for gating this refactor. The deleted guard has the same close-then-unlink sequence. I am withdrawing the merge block for this PR; the exclusion defect remains a durable-state bug for the follow-up.

One clarification for the follow-up's interleaving: a non-blocking contender need not acquire during the close/unlink gap. B can open inode I while A still holds it, pause before `flock`, and resume after A has both closed and unlinked I. B then locks I successfully while C can lock the replacement inode J. Thus `try_lock` does not bound the race to the few instructions between close and unlink. The existing warning that removal is unsound still applies; the follow-up test should cover suspension between open and acquisition as well.

PR #1150, opened 2026-09-10
feat(config, workspace): Add the Anthropic credential chain

JeanMertz:
`providers.llm.anthropic.auth` names the credentials a request may
authenticate with, in fallback order. Each entry selects a source:
`api_key` reads the environment variable `api_key_env` names, `profile`
the sole stored credential profile, and `profile:<name>` a named one.
Entries are tried in order, so listing `api_key` after a subscription
profile is what authorizes continuing on per-token billing once the
allowance is gone.

The default is `["api_key"]`, which is what every existing setup already
does, so nothing changes for anyone who does not opt in. The list
replaces rather than appends across config layers: merging two chains
element-wise would produce an order nobody wrote, while replacement
keeps every effective chain one that some layer spelled out in full.

`AuthEntry` parses from and renders back to the string form users write,
and rejects anything else at parse time, so an unrecognized entry fails
where the value enters rather than deep inside resolution. Validation
covers what parsing cannot see: an empty chain, and duplicates.

`user_data_dir` moves from `jp_workspace` into `jp_config::fs`, beside
the other user-global path resolution. The credential store needs it and
must not depend on the workspace, since credentials are user-global and
the auth commands run with no workspace at all. `jp_workspace` keeps a
thin wrapper that maps the missing-directory case onto its own error.


crates/jp_config/src/providers/llm/anthropic.rs
```diff
@@ -1,19 +1,51 @@
...
+    ///
+    /// This list replaces the one from earlier config layers as a whole; it
+    /// never appends.
+    ///
+    /// ```toml
+    /// [providers.llm.anthropic]
+    /// auth = ["profile:personal", "profile:work", "api_key"]
+    /// ```
+    #[setting(default = vec![AuthEntry::ApiKey])]
+    pub auth: Vec<AuthEntry>,
```
JeanMertz: blocking: Adding `auth` to the generated `PartialAnthropicConfig` leaves the `Self { ... }` initializer in `delta_with_unsets` (lines 139 to 153) missing this field. A normal build of `jp_config`, and consequently JP, fails with a missing-field error (E0063), irrespective of the user's configuration. Add `auth` to that initializer using `delta_opt`, since this list merges by replacement. Verify with `cargo check -p jp_config`; this finding is from source inspection, not a compiler run.

crates/jp_config/src/providers/llm/anthropic.rs
```diff
@@ -60,6 +125,7 @@ impl AssignKeyValue for PartialAnthropicConfig {
 impl PartialConfigDelta for PartialAnthropicConfig {
     fn delta(&self, next: Self) -> Self {
         Self {
+            auth: delta_opt_vec(self.auth.as_ref(), next.auth),
```
JeanMertz: blocking: With workspace auth `["profile:work", "api_key"]`, a continuing conversation invoked with `--cfg=WORKSPACE --cfg providers.llm.anthropic.auth=profile:work` should persist the removal of API-key fallback. `delta_opt_vec` returns `None` for that removal, so the reset payload computed in `jp_cli::lib` retains the workspace's API-key fallback. Reordering the same entries is also lost, and adding an entry emits only the addition even though merging replaces the whole list. Use `delta_opt(self.auth.as_ref(), next.auth)` and add regression tests that merge the resulting delta back into the baseline for removal, reordering, and addition.

crates/jp_config/src/providers/llm/anthropic.rs
```diff
@@ -42,14 +74,47 @@ pub struct AnthropicConfig {
...
 impl AssignKeyValue for PartialAnthropicConfig {
-    fn assign(&mut self, kv: KvAssignment) -> AssignResult {
+    fn assign(&mut self, mut kv: KvAssignment) -> AssignResult {
         match kv.key_string().as_str() {
             "" => kv.try_merge_object(self)?,
             "api_key_env" => self.api_key_env = kv.try_some_string()?,
             "base_url" => self.base_url = kv.try_some_string()?,
             "chain_on_max_tokens" => self.chain_on_max_tokens = kv.try_some_bool()?,
-            "beta_headers" => kv.try_some_vec_of_strings(&mut self.beta_headers)?,
+            _ if kv.p("auth") => kv.try_vec(self.auth.get_or_insert_default(), |kv| {
```
JeanMertz: nit: `--cfg 'providers.llm.anthropic.auth:=null'` should unset the optional partial field, but `get_or_insert_default()` followed by `try_vec` leaves `Some([])`. Resolution then rejects the empty chain rather than restoring the default. This also violates `PartialAppConfig::unset`'s contract for this field. Use the existing `kv.try_some_vec(&mut self.auth, parser)` helper, which preserves the distinction between `None` and an explicit empty list, and test the null assignment.

crates/jp_config/src/providers/llm/anthropic.rs
```diff
@@ -1,19 +1,51 @@
...
+    ///   `api_key_env`.
+    /// - `profile`: The sole stored credential profile.
+    ///   Log in with `jp provider auth login llm.anthropic`.
+    /// - `profile:<name>`: The stored credential profile `<name>`.
+    ///   Profile names are case-sensitive.
+    ///
+    /// Entries are tried in order: when one cannot produce a usable credential,
+    /// JP continues with the next.
+    /// Listing `api_key` after subscription profiles authorizes continuing on
+    /// per-token API billing when the subscription allowance is exhausted.
```
JeanMertz: question: Is this configuration-only PR intended to ship together with the credential resolver? For a user who sets `auth = ["profile:work"]` while `ANTHROPIC_API_KEY` remains in their environment, `Anthropic::try_from` in `jp_llm/src/provider/anthropic.rs:1974` still reads that key unconditionally and builds an API-key client. The accepted configuration therefore does not enforce the billing opt-in described here. If this can ship independently, reject unsupported chains until the resolver is wired in; otherwise make the coordinated-shipping requirement explicit.

crates/jp_config/src/providers/llm/anthropic.rs
JeanMertz: Confirmed and fixed. `cargo check -p jp_config` reproduced it exactly as described: `error[E0063]: missing field 'auth' in initializer of 'PartialAnthropicConfig'` at `anthropic.rs:140`, failing both the `lib` and `lib test` targets.

`delta_opt` is the right helper here, as you say: `auth` carries no `merge = …`, and `schematic_macros/src/config/field_value.rs:342` generates `if next.key.is_some() { self.key = next.key }` for an optional collection without one — replacement. A replacing field never needs a path in `unsets`, which is what the `delta_with_unsets` doc at `delta.rs:34` describes.

The added line carries a comment saying so, to keep the next reader from reaching for `delta_opt_vec_at` the way `beta_headers` does.

crates/jp_config/src/providers/llm/anthropic.rs
JeanMertz: Correct, and fixed by switching the arm to `kv.try_some_vec(&mut self.auth, …)`.

Traced it through: `try_vec` (`assignment.rs:926`) checks `clears_collection()` *after* the caller has already run `get_or_insert_default`, so it clears the vec in place and leaves `Some([])`. `try_some_vec` (`assignment.rs:1007`) checks the same guard first and sets `*vec = None` — the same shape `try_some_vec_of_strings` uses for `beta_headers` on the next line, with the comment at `assignment.rs:1043` explaining why the distinction matters.

You are also right about the `unset` contract: `PartialAppConfig::unset`'s doc at `lib.rs:709` and `unset_clears_a_list_to_none` (`unset_tests.rs:43`) both fix it as `None`. Worth noting the existing sweep, `every_schema_field_is_reachable_by_path` (`unset_tests.rs:97`), would not have caught this: it only asserts `unset` returns `Ok`, never that the field ends up `None`. Added `test_assign_auth_chain_null_clears_to_none`, which fails with `Some([])` against the old arm.

crates/jp_config/src/providers/llm/anthropic.rs
JeanMertz: Right on all three shapes, and the addition case is worse than "emits only the addition". Switched both `delta` and `delta_with_unsets` to `delta_opt`.

`delta_opt_vec`'s own doc at `delta.rs:164` already says this: *"Use `delta_opt` instead for a vector field that merges by replacement: there the whole of `next` is the delta."*

Added four tests modelled on `delta_law_tests.rs` (`fold(prev, delta(prev, next)) == next`), and ran them against the unfixed code first. Observed failures:

- removal `[profile:work, api_key] -> [profile:work]`: folded to `Some([Profile(Some("work")), ApiKey])`. Change lost.
- reorder `[profile:work, api_key] -> [api_key, profile:work]`: folded to `Some([Profile(Some("work")), ApiKey])`. Change lost.
- addition `[profile:work] -> [profile:work, api_key]`: folded to `Some([ApiKey])`. Not a lost change — a chain nobody wrote, with the profile silently dropped and only the paid fallback left.

The fourth asserts an unchanged chain still deltas to `None`.

crates/jp_config/src/providers/llm/anthropic.rs
JeanMertz: Yes — coordinated. This is one PR in a stack that lands together; the credential store and chain resolution follow immediately behind it, so there is no window in which a release carries an inert `auth` key.

Your reading of the current state is accurate: `AuthEntry` is referenced nowhere outside `jp_config` (plus the `app_config_fields` snapshot), `Anthropic::try_from` at `jp_llm/src/provider/anthropic.rs:1975` reads `api_key_env` unconditionally, and no `jp auth` / `jp provider auth` command exists under `crates/jp_cli/src/cmd` yet.

Not taking the "reject unsupported chains until the resolver is wired in" suggestion, though. It would mean shipping a config key whose only accepted value is its default, along with the validation, its test, and a schema entry that all get deleted in the next PR — a temporary lie in place of a temporary gap, and the key lands in `AppConfig::fields()` either way. Given the stack merges as a unit, the gap closes before anyone can reach it.

One thing worth a separate follow-up: the entry spelling here (`profile:<name>`) diverges from RFD 090, which writes `oauth:<name>`, and the command in the doc comment (`jp provider auth login llm.anthropic`) diverges from the RFD's `jp auth login anthropic`. Whichever spelling wins, the RFD and the code should agree before the resolver PR documents it further.

crates/jp_llm/tests/fixtures/anthropic/test_chat_completion_stream__conversation_stream.snap.new
```diff
@@ -0,0 +1,245 @@
+---
+source: crates/jp_test/src/mock.rs
+assertion_line: 135
```
JeanMertz: nit: Remove the committed `.snap.new` companions under `crates/jp_llm/tests/fixtures`, keeping the updated `.snap` files. I compared the companions with their accepted snapshots: they match after excluding `assertion_line` metadata. These are stale pending results, not additional test coverage, and contributors opening `cargo insta review` inherit them as review work on a fresh checkout.

PR #1151, opened 2026-09-10
feat(llm, cli, anthropic): Add Anthropic subscription auth with fallback

JeanMertz:
JP can now authenticate Anthropic requests with a Claude Pro/Max
subscription alongside an API key, and move between them on its own. A
spent allowance no longer ends a turn: the provider records what was
spent, switches to the next credential in the chain, and carries on.

The Anthropic provider owns its credential policy. It reads the `auth`
chain, walks it before every request, refreshes an expired OAuth token
in place, and switches at request admission when a credential is spent
or refused. Nothing above it knows what a credential is: every provider
is built from its config alone, and the turn loop, the tasks, and every
other call site are credential-unaware. What core owns is the store —
`jp_credentials` holds one versioned document, and every mutation runs
under a lock that re-reads and rechecks its precondition, because
refresh tokens rotate and two processes racing a refresh would otherwise
invalidate each other.

Bearer requests carry the Claude Code request fingerprint, which
Anthropic enforces: a bearer request whose system prompt does not begin
with their identity line is refused. JP sends it and immediately follows
it with a block naming it as a transport artifact, so a foreign identity
does not stand as an instruction ahead of JP's own prompt.

A subscription's usage state arrives in response headers rather than
error bodies, and on every response rather than only rejections. That
distinction matters twice. It is the only way to tell a spent window
from ordinary throttling — both are `429 rate_limit_error` with an
opaque body — and it is the only warning JP gets when an account with
extra usage enabled is quietly billing past its allowance. Both are read
and acted on: a scoped cooldown routes the next resolution off the
profile, and a crossed warning threshold is surfaced while the user can
still choose what to spend the rest on.

Reading those headers required the streaming transport to change. The
vendored `async-anthropic` fork now issues the request itself and parses
the body with `eventsource-stream`, which makes a successful response's
headers reachable. SSE-level reconnection goes with it, deliberately:
Anthropic sends no event ids, so a reconnect cannot resume and instead
issues a fresh completion from the beginning. Nothing consumed it, while
the orphaned attempt still fired fifteen seconds later and paid for
output nobody read. Recovery stays in the turn loop, which flushes
partial content and continues from it.

`Event::Notice` carries provider decisions the user must see — a skipped
credential, a switch, a nearly-spent window — as chrome on stderr,
without becoming part of the conversation.

`jp provider auth login|list|logout` manages stored profiles. Login
takes a `claude setup-token` value today; the PKCE, exchange, and
refresh machinery for browser login is in place and the command that
drives it follows.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2468,7 +2468,6 @@ dependencies = [
 name = "jp_credentials"
 version = "0.1.0"
 dependencies = [
- "assert_matches",
  "camino",
  "camino-tempfile",
  "chrono",
@@ -2479,7 +2478,6 @@ dependencies = [
  "serde_json",
  "test-log",
  "thiserror 2.0.20",
- "tracing",
 ]
 
 [[package]]
@@ -4134,7 +4132,6 @@ dependencies = [
 name = "schematic"
 version = "0.19.4"
 dependencies = [
- "convert_case 0.11.0",
  "indexmap",
  "json-strip-comments",
  "schematic",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -171,19 +171,23 @@ checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
 [[package]]
 name = "async-anthropic"
 version = "0.6.0"
-source = "git+https://github.com/JeanMertz/async-anthropic#8c12222078fe90f9e8f15fb078790a6d86ea42c9"
 dependencies = [
+ "async-trait",
  "backon",
  "derive_builder",
+ "eventsource-stream",
+ "futures",
  "reqwest",
- "reqwest-eventsource",
  "secrecy",
  "serde",
  "serde_json",
+ "test-log",
  "thiserror 2.0.20",
  "tokio",
  "tokio-stream",
  "tracing",
+ "uuid",
+ "wiremock",
 ]
 
 [[package]]
@@ -949,6 +953,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "deadpool"
+version = "0.12.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0be2b1d1d6ec8d846f05e137292d0b89133caf95ef33695424c09568bdd39b1b"
+dependencies = [
+ "deadpool-runtime",
+ "lazy_static",
+ "num_cpus",
+ "tokio",
+]
+
+[[package]]
+name = "deadpool-runtime"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "092966b41edc516079bdf31ec78a2e0588d1d0c08f78b91d8307215928642b2b"
+
+[[package]]
 name = "derive_builder"
 version = "0.20.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1484,9 +1506,20 @@ dependencies = [
  "cfg-if",
  "js-sys",
  "libc",
- "r-efi",
+ "r-efi 5.3.0",
  "wasip2",
  "wasm-bindgen",
+]
+
+[[package]]
+name = "getrandom"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "300e883d756b2e4ec94e02791f39b04b522276138852cfc41d9fb7e904106099"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "r-efi 6.0.0",
 ]
 
 [[package]]
@@ -1638,6 +1671,12 @@ name = "heck"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
+
+[[package]]
+name = "hermit-abi"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc0fef456e4baa96da950455cd02c081ca953b141298e41db3fc7e36b1da849c"
 
 [[package]]
 name = "htmd"
@@ -2299,6 +2338,7 @@ dependencies = [
  "comfy-table",
  "comrak",
  "crossterm",
+ "datetime_literal",
  "dhat",
  "duct",
  "fancy-regex 0.19.0",
@@ -2319,6 +2359,7 @@ dependencies = [
  "jp_attachment_mcp_resources",
  "jp_config",
  "jp_conversation",
+ "jp_credentials",
  "jp_editor",
  "jp_id",
  "jp_inquire",
@@ -2424,6 +2465,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "jp_credentials"
+version = "0.1.0"
+dependencies = [
+ "assert_matches",
+ "camino",
+ "camino-tempfile",
+ "chrono",
+ "datetime_literal",
+ "jp_config",
+ "jp_storage",
+ "serde",
+ "serde_json",
+ "test-log",
+ "thiserror 2.0.20",
+ "tracing",
+]
+
+[[package]]
 name = "jp_editor"
 version = "0.1.0"
 dependencies = [
@@ -2513,6 +2572,7 @@ dependencies = [
  "base64",
  "camino",
  "chrono",
+ "datetime_literal",
  "eventsource-stream",
  "futures",
  "gemini_client_rs",
@@ -2523,8 +2583,10 @@ dependencies = [
  "jp_attachment",
  "jp_config",
  "jp_conversation",
+ "jp_credentials",
  "jp_mcp",
  "jp_openrouter",
+ "jp_storage",
  "jp_test",
  "jp_tool",
  "minijinja",
@@ -2537,12 +2599,14 @@ dependencies = [
  "reqwest-eventsource",
  "serde",
  "serde_json",
+ "sha2",
  "test-log",
  "thiserror 2.0.20",
  "tokio",
  "tokio-util",
  "tracing",
  "url",
+ "uuid",
 ]
 
 [[package]]
@@ -3098,6 +3162,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "num_cpus"
+version = "1.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "91df4bbde75afed763b708b7eee1e8e7651e02d97f6d5dd763e89367e957b23b"
+dependencies = [
+ "hermit-abi",
+ "libc",
+]
+
+[[package]]
 name = "object"
 version = "0.36.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3550,6 +3624,12 @@ name = "r-efi"
 version = "5.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "69cdb34c158ceb288df11e18b4bd39de994f6657d83847bdffdbd7f346754b0f"
+
+[[package]]
+name = "r-efi"
+version = "6.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f8dcc9c7d52a811697d2151c701e0d08956f92b0e24136cf4cf27b57a6a0d9bf"
 
 [[package]]
 name = "ra-ap-rustc_lexer"
@@ -5306,6 +5386,7 @@ version = "1.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a68d3c8f01c0cfa54a75291d83601161799e4a89a39e0929f4b0354d88757a37"
 dependencies = [
+ "getrandom 0.4.3",
  "js-sys",
  "wasm-bindgen",
 ]
@@ -5863,6 +5944,29 @@ name = "winsafe"
 version = "0.0.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d135d17ab770252ad95e9a872d365cf3090e3be864a34ab46f48555993efc904"
+
+[[package]]
+name = "wiremock"
+version = "0.6.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08db1edfb05d9b3c1542e521aea074442088292f00b5f28e435c714a98f85031"
+dependencies = [
+ "assert-json-diff",
+ "base64",
+ "deadpool",
+ "futures",
+ "http",
+ "http-body-util",
+ "hyper",
+ "hyper-util",
+ "log",
+ "once_cell",
+ "regex",
+ "serde",
+ "serde_json",
+ "tokio",
+ "url",
+]
 
 [[package]]
 name = "wit-bindgen"
--- a/crates/contrib/async-anthropic/src/client.rs
+++ b/crates/contrib/async-anthropic/src/client.rs
@@ -1,0 +1,460 @@
+use std::{pin::Pin, time::Duration};
+
+use backon::{ExponentialBuilder, Retryable as _};
+use derive_builder::Builder;
+use eventsource_stream::Eventsource as _;
+use futures::StreamExt as _;
+use reqwest::StatusCode;
+use secrecy::ExposeSecret;
+use serde::{Serialize, de::DeserializeOwned};
+use tokio_stream::Stream;
+
+use crate::{
+    bearer,
+    errors::{
+        AnthropicError, ApiError, ApiErrorEnvelope, UnifiedRateLimit, map_deserialization_error,
+    },
+    messages::Messages,
+    models::Models,
+};
+
+const BASE_URL: &str = "https://api.anthropic.com";
+
+/// Main entry point for the Anthropic API
+///
+/// By default will use the `ANTHROPIC_API_KEY` environment variable
+///
+/// # Example
+///
+/// ```no_run
+/// # use async_anthropic::types::*;
+/// # async fn run() {
+/// let client = async_anthropic::Client::default();
+///
+/// let request = CreateMessagesRequestBuilder::default()
+///     .model("claude-3.5-sonnet")
+///     .messages(vec![
+///         MessageBuilder::default()
+///             .role(MessageRole::User)
+///             .content("Hello world!")
+///             .build()
+///             .unwrap(),
+///     ])
+///     .build()
+///     .unwrap();
+///
+/// client.messages().create(request).await.unwrap();
+/// # }
+/// ```
+#[derive(Clone, Debug, Builder)]
+#[builder(setter(into, strip_option))]
+pub struct Client {
+    #[builder(default)]
+    http_client: reqwest::Client,
+    #[builder(default)]
+    base_url: String,
+    #[builder(default = default_api_key())]
+    api_key: secrecy::SecretString,
+    /// OAuth bearer token for subscription (Claude Pro/Max) accounts.
+    ///
+    /// When set, requests authenticate with `Authorization: Bearer <token>` and
+    /// carry the Claude Code request fingerprint (see [`bearer`]); `api_key` is
+    /// ignored and `x-api-key` is never sent.
+    /// User-configured `beta` values merge after the fingerprint's own beta set
+    /// and can neither remove nor duplicate its entries.
+    #[builder(default)]
+    auth_token: Option<secrecy::SecretString>,
+    #[builder(default)]
+    version: String,
+    #[builder(default)]
+    beta: Option<String>,
+    #[builder(default)]
+    backoff: ExponentialBuilder,
+}
+
+impl Default for Client {
+    fn default() -> Self {
+        // Load backoff settings from configuration
+        let backoff = ExponentialBuilder::default()
+            .with_min_delay(Duration::from_secs(15))
+            .with_factor(2.0)
+            .with_jitter()
+            .with_max_delay(Duration::from_mins(2));
+
+        Self {
+            http_client: reqwest::Client::new(),
+            api_key: default_api_key(), // Default env?
+            auth_token: None,
+            version: "2023-06-01".to_string(),
+            beta: None,
+            base_url: BASE_URL.to_string(),
+            backoff,
+        }
+    }
+}
+
+fn default_api_key() -> secrecy::SecretString {
+    if cfg!(test) {
+        return "test".into();
+    }
+    std::env::var("ANTHROPIC_API_KEY")
+        .unwrap_or_else(|_| {
+            tracing::warn!("Default Anthropic client initialized without api key");
+            String::new()
+        })
+        .into()
+}
+
+impl Client {
+    /// Build a new client from an API key
+    pub fn from_api_key(api_key: impl Into<secrecy::SecretString>) -> Self {
+        Self {
+            api_key: api_key.into(),
+            ..Default::default()
+        }
+    }
+
+    /// Create a new client builder
+    #[must_use]
+    pub fn builder() -> ClientBuilder {
+        ClientBuilder::default()
+    }
+
+    /// Set a custom backoff strategy
+    #[must_use]
+    pub fn with_backoff(mut self, backoff: ExponentialBuilder) -> Self {
+        self.backoff = backoff;
+        self
+    }
+
+    /// Call the messages api
+    #[must_use]
+    pub fn messages(&self) -> Messages<'_> {
+        Messages::new(self)
+    }
+
+    #[must_use]
+    pub fn models(&self) -> Models<'_> {
+        Models::new(self)
+    }
+
+    /// Build the request headers, folding `betas` into the client's own set.
+    ///
+    /// `betas` are the extras a single request asked for; they join the
+    /// client-wide `beta` value in order, without duplicates.
+    fn headers(&self, betas: &[String]) -> reqwest::header::HeaderMap {
+        let mut extras: Vec<&str> = self
+            .beta
+            .iter()
+            .flat_map(|beta| beta.split(','))
+            .map(str::trim)
+            .filter(|beta| !beta.is_empty())
+            .collect();
+
+        for beta in betas {
+            let beta = beta.trim();
+            if !beta.is_empty() && !extras.contains(&beta) {
+                extras.push(beta);
+            }
+        }
+
+        let extras = extras.join(",");
+        let extras = (!extras.is_empty()).then_some(extras);
+
+        if let Some(token) = &self.auth_token {
+            // The fingerprint's own betas lead and can be neither removed nor
+            // duplicated by these; see `bearer::merge_betas`.
+            return bearer::headers(token.expose_secret(), &self.version, extras.as_deref());
+        }
+
+        let mut headers = reqwest::header::HeaderMap::new();
+        headers.insert("x-api-key", self.api_key.expose_secret().parse().unwrap());
+        headers.insert("anthropic-version", self.version.parse().unwrap());
+        if let Some(extras) = extras {
+            headers.insert("anthropic-beta", extras.parse().unwrap());
+        }
+
+        headers
+    }
+
+    fn format_url(&self, path: &str) -> String {
+        format!(
+            "{}/{}",
+            &self.base_url.trim_end_matches('/'),
+            &path.trim_start_matches('/')
+        )
+    }
+
+    pub async fn get<O>(&self, path: &str) -> Result<O, AnthropicError>
+    where
+        O: DeserializeOwned,
+    {
+        let request = || async {
+            let response = self
+                .http_client
+                .get(self.format_url(path))
+                .headers(self.headers(&[]))
+                .send()
+                .await
+                .map_err(AnthropicError::Network)?;
+
+            handle_response(response).await
+        };
+
+        request
+            .retry(self.backoff)
+            .sleep(tokio::time::sleep)
+            // A spent quota is not throttling: waiting cannot help, and the
+            // caller may have another credential that can serve the request
+            // now. Only capacity throttling is retried in place.
+            .when(|e| {
+                matches!(e, AnthropicError::RateLimit { limits, .. } if !limits.is_quota_rejection())
+            })
+            .adjust(|err, dur| match err {
+                AnthropicError::RateLimit { retry_after, .. } => {
+                    retry_after.map(Duration::from_secs).or(dur)
+                }
+                _ => dur,
+            })
+            .await
+    }
+
+    /// Make post request to the API
+    ///
+    /// This includes all headers and error handling
+    pub async fn post<I, O>(
+        &self,
+        path: &str,
+        request: I,
+        betas: &[String],
+    ) -> Result<O, AnthropicError>
+    where
+        I: Serialize,
+        O: DeserializeOwned,
+    {
+        let request = || async {
+            // `headers()` already carries the `anthropic-beta` value.
+            let request = self
+                .http_client
+                .post(self.format_url(path))
+                .headers(self.headers(betas))
+                .json(&request);
+
+            let response = request.send().await.map_err(AnthropicError::Network)?;
+
+            handle_response(response).await
+        };
+
+        request
+            .retry(self.backoff)
+            .sleep(tokio::time::sleep)
+            // A spent quota is not throttling: waiting cannot help, and the
+            // caller may have another credential that can serve the request
+            // now. Only capacity throttling is retried in place.
+            .when(|e| {
+                matches!(e, AnthropicError::RateLimit { limits, .. } if !limits.is_quota_rejection())
+            })
+            .adjust(|err, dur| match err {
+                AnthropicError::RateLimit { retry_after, .. } => {
+                    retry_after.map(Duration::from_secs).or(dur)
+                }
+                _ => dur,
+            })
+            .await
+    }
+
+    /// Open a server-sent-event stream.
+    ///
+    /// Returns the response's quota headers alongside the event stream: the
+    /// unified rate-limit headers ride on every response, including successful
+    /// ones, and are the only place a subscription's usage state is reported.
+    ///
+    /// The stream is not reconnected on failure.
+    /// Anthropic's message endpoint sends no SSE event ids, so a reconnect
+    /// cannot resume: it would issue a fresh completion whose content restarts
+    /// from the beginning.
+    /// Recovery belongs to the caller, which can rebuild the request from
+    /// whatever it already received.
+    pub(crate) async fn post_stream<I, O, const N: usize>(
+        &self,
+        path: &str,
+        request: I,
+        event_types: [&'static str; N],
+        betas: &[String],
+    ) -> Result<StreamResponse<O>, AnthropicError>
+    where
+        I: Serialize,
+        O: DeserializeOwned + Send + 'static,
+    {
+        let response = self
+            .http_client
+            .post(self.format_url(path))
+            .headers(self.headers(betas))
+            .header(reqwest::header::ACCEPT, "text/event-stream")
+            .json(&request)
+            .send()
+            .await
+            .map_err(AnthropicError::Network)?;
+
+        // Read the quota headers before the body is consumed, on success as
+        // well as failure.
+        let limits = UnifiedRateLimit::from_headers(response.headers());
+
+        if response.status() != StatusCode::OK {
+            return Err(response_error(response).await);
+        }
+
+        Ok(StreamResponse {
+            limits,
+            stream: stream(response, event_types),
+        })
+    }
+}
+
+/// A server-sent-event stream plus the quota state its response reported.
+pub struct StreamResponse<O> {
+    /// The unified rate-limit headers the response carried.
+    pub limits: UnifiedRateLimit,
+
+    /// The parsed events.
+    pub stream: Pin<Box<dyn Stream<Item = Result<O, AnthropicError>> + Send>>,
+}
+
+async fn handle_response<O>(response: reqwest::Response) -> Result<O, AnthropicError>
+where
+    O: DeserializeOwned,
+{
+    if response.status() == StatusCode::OK {
+        return response.json::<O>().await.map_err(AnthropicError::Network);
+    }
+
+    Err(response_error(response).await)
+}
+
+/// Classify a non-success response.
+///
+/// Shared by the buffered and streaming paths so one place decides what a
+/// status means.
+async fn response_error(response: reqwest::Response) -> AnthropicError {
+    let status = response.status();
+
+    // 529 is the status code for overloaded requests
+    let overloaded_status = StatusCode::from_u16(529).expect("529 is a valid status code");
+
+    if status == StatusCode::TOO_MANY_REQUESTS || status == overloaded_status {
+        let retry_after = response
+            .headers()
+            .get("Retry-After")
+            .and_then(|h| h.to_str().ok())
+            .and_then(|s| s.parse::<u64>().ok());
+
+        // Read the quota headers before consuming the body: they are
+        // the only thing that distinguishes a spent usage window from
+        // capacity throttling, and the body often says nothing.
+        let limits = UnifiedRateLimit::from_headers(response.headers());
+
+        let text = response.text().await.unwrap_or_default();
+        tracing::warn!(?limits, "Rate limited: {text}");
+
+        return AnthropicError::RateLimit {
+            retry_after,
+            limits,
+        };
+    }
+
+    // The credential was refused. The status is what makes this actionable,
+    // and it is lost once the body is parsed into the generic error envelope.
+    if status == StatusCode::UNAUTHORIZED || status == StatusCode::FORBIDDEN {
+        let text = response.text().await.unwrap_or_default();
+        tracing::warn!(%status, "Authentication rejected: {text}");
+
+        let error = serde_json::from_str::<ApiErrorEnvelope>(&text).map_or_else(
+            |_| ApiError {
+                error_type: "authentication_error".to_owned(),
+                message: Some(text),
+            },
+            |envelope| envelope.error,
+        );
+
+        return AnthropicError::Auth {
+            status: status.as_u16(),
+            error,
+        };
+    }
+
+    let text = response.text().await.unwrap_or_default();
+    match serde_json::from_str::<ApiErrorEnvelope>(&text) {
+        Ok(envelope) => AnthropicError::Api(envelope.error),
+        Err(_) => AnthropicError::Unknown(text),
+    }
+}
+
+/// Parse an SSE response body into typed events.
+///
+/// The stream is driven by the caller rather than by a background task, so a
+/// dropped stream stops reading the body instead of continuing to consume it.
+fn stream<O, const N: usize>(
+    response: reqwest::Response,
+    event_types: [&'static str; N],
+) -> Pin<Box<dyn Stream<Item = Result<O, AnthropicError>> + Send>>
+where
+    O: DeserializeOwned + Send + 'static,
+{
+    let events = response
+        .bytes_stream()
+        .eventsource()
+        .map(move |event| {
+            tracing::trace!("Streaming event: {event:?}");
+            match event {
+                Ok(message) => decode_event(&message, &event_types),
+                Err(error) => Err(AnthropicError::StreamTransport(error.to_string())),
+            }
+        })
+        // A failure ends the stream: whatever follows belongs to a response
+        // the caller has already been told is broken.
+        .scan(false, |finished, item| {
+            if *finished {
+                return futures::future::ready(None);
+            }
+
+            *finished = item.is_err();
+            futures::future::ready(Some(item))
+        });
+
+    Box::pin(events)
+}
+
+/// Decode one server-sent event into `O`.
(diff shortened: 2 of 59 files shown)


crates/jp_config/src/providers/llm/anthropic.rs
```diff
@@ -72,6 +138,7 @@ impl PartialConfigDelta for PartialAnthropicConfig {
 
     fn delta_with_unsets(&self, next: Self, prefix: &str, unsets: &mut Vec<String>) -> Self {
         Self {
+            auth: delta_opt_vec_at(&path(prefix, "auth"), self.auth.as_ref(), next.auth, unsets),
```
JeanMertz: blocking: Persist the whole replacement auth chain in config deltas. For a conversation with `auth = ["profile:personal"]`, adding API fallback via `--cfg providers.llm.anthropic.auth=profile:personal,api_key` makes `delta_opt_vec_at` emit only `["api_key"]`. That helper assumes append merging, but `auth` merges by replacement. `turn_config_delta` records this delta, so the persisted conversation silently loses its subscription entry and uses API billing directly on subsequent turns.

Use `delta_opt` for `auth` in both `delta` and `delta_with_unsets`. Add a delta/apply round-trip test for extending a subscription-only chain with API fallback; the resolved chain must retain the profile first.

crates/jp_llm/src/provider/anthropic/resolve.rs
```diff
@@ -0,0 +1,706 @@
...
+            Landing::Stale {
+                profile,
+                refresh_token,
+            } => (profile, refresh_token),
+        };
+
+        let (profile, refresh_token) = stale;
+        debug!(profile, "Access token expired; refreshing.");
+
+        match oauth::refresh(&refresh_token).await {
```
JeanMertz: blocking: Serialize the refresh request itself, not just persistence. With an expired stored `oauth` profile used by concurrent JP requests, both callers can read the same refresh token and submit it here before either enters `store.mutate`. The compare-and-set in `persist_rotation` happens after the token endpoint has already processed both requests, so it cannot prevent reuse of a rotated token or the resulting grant invalidation. A refused competing refresh can also retire the profile before the successful refresh is persisted.

Acquire a cross-process refresh lock, reload and recheck expiry under it, and hold it through the token exchange and persistence. Keep blocking lock acquisition off the async executor. Add a concurrent-refresh test that asserts only one exchange consumes the stale token and both callers resolve the persisted replacement.

crates/jp_llm/src/provider/anthropic/oauth.rs
```diff
@@ -0,0 +1,314 @@
...
+        .post(TOKEN_URL)
+        .json(&body)
+        .send()
+        .await?;
+
+    let status = response.status();
+    let text = response.text().await?;
+
+    if !status.is_success() {
+        return Err(OauthError::Rejected {
```
JeanMertz: blocking: A `503` or `429` from the token endpoint while refreshing an expired OAuth profile enters `OauthError::Rejected` here. `is_rejection()` accepts every instance of that variant, and `resolve` consequently persists `needs_relogin = true`. A temporary outage therefore disables a valid profile across future invocations and can move the user onto API billing after the endpoint recovers.

Distinguish transient HTTP failures from an actual refused grant. Only definitive credential/grant rejection should retire the profile; preserve its state for throttling and server errors. Add a refresh test for HTTP 503 asserting that the stored profile is not marked for re-login.

crates/jp_llm/src/provider/anthropic/resolve.rs
```diff
@@ -0,0 +1,706 @@
...
+    // Only a stored profile has state to record against.
+    let AuthEntry::Profile(Some(profile)) = spent else {
+        return;
+    };
+    let Some(store) = store else {
+        return;
+    };
+
+    let result = if error.is_auth_rejected() {
+        store.mark_needs_relogin(CATEGORY_LLM, PROVIDER_ANTHROPIC, profile)
```
JeanMertz: blocking: Check that the stored credential still matches the failed attempt before marking it for re-login. A user can replace a revoked setup token with `jp provider auth login ... --profile personal` while another JP request using the old token is in flight. When that old request returns 401, this name-only update marks the freshly logged-in credential unusable. The same issue lets a delayed quota response cool down a different account installed under that profile name.

Carry the attempted credential or a generation identifier into outcome recording and compare it with the current profile inside the locked mutation. Apply the same guard to `QuotaWatch` writes. Test replacing a profile between request resolution and outcome recording; the replacement must remain untouched.

crates/jp_llm/src/provider/anthropic.rs
```diff
@@ -363,6 +716,7 @@ fn call(
                         chain_events,
                         is_structured,
                         chains_remaining - 1,
+                        watch.clone(),
```
JeanMertz: blocking: Re-resolve credentials before the continuation request. For a subscription account with extra usage enabled, a successful response can report `status: rejected`, record a cooldown in `watch.observe`, and end with `max_tokens`. This branch still passes the same authenticated client into `chain`, which recursively calls `call` without resolution. The continuation therefore bills extra usage against the profile JP just marked spent, even when another subscription profile is available. The forced-tool retry paths also reuse that client.

Route these internal requests through credential resolution and auth-dependent request construction, preserving their continuation/tool-choice state. Add a wire-level test where a 200 response reports a spent window and triggers a continuation; the next request must use the next configured credential, not the cooled-down one.

crates/jp_llm/src/event_builder.rs
```diff
@@ -65,7 +65,8 @@ pub fn structured_data(events: Vec<Event>) -> Option<Value> {
                 flushed.extend(builder.handle_flush(index, metadata));
             }
             Event::Finished(_) => flushed.extend(builder.drain()),
-            Event::Patch(_) | Event::KeepAlive => {}
+            // A notice is chrome, and carries no structured payload.
+            Event::Patch(_) | Event::KeepAlive | Event::Notice(_) => {}
```
JeanMertz: blocking: Preserve a delivery path for notices from collected requests. When title generation encounters an exhausted subscription and falls back to `api_key`, `title::generate` passes the collected events here and the switch notice is discarded. The title succeeds on paid billing without the promised stderr notice. Summarization sends notices only to `tracing::warn`, whose stderr layer is disabled unless requested, and only after collection completes.

Keep this helper free of rendering, but carry notices to the caller's chrome sink as the stream is consumed, including notices from attempts that later fail. Add a title-generation fallback test asserting that the credential switch reaches stderr without enabling verbose logging.

crates/jp_llm/src/provider/anthropic.rs
```diff
@@ -314,10 +655,22 @@ fn call(
...
         let mut tool_names_called: Vec<String> = vec![];
 
-        let stream = client
+        // An admission rejection (a spent quota, a refused credential)
+        // surfaces here, before any event exists.
+        let response = client
             .messages()
             .create_stream(request.clone())
             .await
+            .map_err(StreamError::from)?;
```
JeanMertz: nit: Admission errors bypass thinking-history repair. For a resumed conversation whose signed thinking block receives HTTP 400 `invalid_request_error` with `Invalid signature ... thinking block`, `create_stream` returns `Err` and this `?` exits before `classify_thinking_rejection` in the event loop runs. The caller gets a terminal error rather than the `Patch` and `Finished(Retry)` events that repair the history. This also affects fallback when another account rejects a stored signature.

Route admission failures through the same thinking-rejection handling as stream failures. Add a mock HTTP 400 test through `call`, asserting the patch/retry events; the existing classifier tests exercise the helpers but not this admission path.

crates/jp_llm/src/provider/anthropic/resolve.rs
```diff
@@ -0,0 +1,706 @@
...
+            return None;
+        }
+    };
+
+    // Resolution landing on the same entry means nothing was recorded that
+    // could move it: `api_key` has no store entry, so an exhausted key falls
+    // through instead of being retried forever.
+    if attempt.selected.as_ref() == Some(spent) {
+        debug!("Credential chain did not advance; treating the failure as terminal.");
+        return None;
```
JeanMertz: nit: An API key cannot fall through to a later profile. With `auth = ["api_key", "profile:personal"]` and a revoked API key, the 401 reaches `advance`, but `record_outcome` has no store entry to update. Re-resolution chooses `api_key` again and this branch ends the request without trying the valid subscription profile. The config accepts this ordering and describes fallback for entries that cannot serve the request.

Track failed entries for the current request independently of persisted profile state, so resolution can skip a rejected API key without persisting its failure. That also lets fallback work when a best-effort cooldown write fails. Add a two-entry test for API-key rejection followed by a usable profile.

crates/jp_config/src/providers/llm/anthropic.rs
JeanMertz: Already done, which is why GitHub has marked this outdated.

`delta` uses `delta_opt` for `auth` (`anthropic.rs:134`), and `delta_with_unsets` no longer exists — it was removed in favour of the default plain diff, with a comment making your own argument: *"`auth` merges by replacement and `beta_headers` carries its own strategy, so every field here is reachable by merging."*

The round-trip test exists too: `assert_auth_delta_law` in `anthropic_tests.rs:150`, with `test_auth_delta_law_holds_for_an_added_entry` covering exactly the "extend a subscription-only chain with API fallback" case, plus reorder and removal.

`git log -S` puts the fix in `2e54fbba`, which landed via #1150 (this PR's base), so the hunk left the diff on rebase. The analysis was correct when written.

crates/jp_llm/src/provider/anthropic.rs
JeanMertz: Fixed in `e47d1ba` — and this is a regression rather than a nit, so thanks for catching it.

`git diff` of `adbbb3ef` shows the old code was `create_stream(...).await.map_err(StreamError::from).map_ok(…)` — `TryStreamExt` combinators, because the external crate returned a *stream*. Admission errors arrived as stream items and reached `classify_thinking_rejection`. The vendored fork's `post_stream` checks the status and returns `Err`, so the `?` now short-circuits. A stale signature is a request *validation* failure, so HTTP 400 at admission is its primary delivery path: the repair was dead for its main trigger.

The classify/build pair moved into `thinking_repair`, called from both the admission error and the stream loop.

`a_rejected_thinking_block_is_repaired_at_admission` drives a real 400 through `call`. Verified red first — without the fix it fails with `StreamError { kind: Other, message: "api error: invalid_request_error: messages.1.content.0: Invalid \`signature\` in \`thinking\` block" }` reaching the caller as terminal.

crates/jp_llm/src/provider/anthropic.rs
JeanMertz: Real defect, fixed in `36672e3c` — but not by re-resolving, because that fix has a cost worth naming.

Confirmed mechanically: `chain(client, …)` forwards to `call(client, …)`, `chain_on_max_tokens` defaults to `true`, and `watch.observe` records the cooldown off a **200**. So the continuation bills extra usage against a profile JP just marked spent.

The problem with routing continuations through resolution: it switches accounts mid-chain, and `chain_events` carries thinking signatures minted under the old account. RFD 090's own Risks section flags cross-account signature rejection, so that trades a billing leak for a likely 400.

Instead, a spent window zeroes `chains_remaining`. The turn ends at `max_tokens`, the caller asks for the rest, and the next turn re-resolves onto a fresh credential — signature invariant intact, cost is one truncated continuation.

The forced-tool retry keeps its request deliberately, noted in a comment: a truncated answer is something the caller can continue, a required tool that never ran has no such recovery.

`a_spent_window_stops_the_continuation` pins it. Without the fix: 6 requests instead of 1.

Worth noting the pre-existing partial mitigation you may not have seen: if the continuation *fails*, `content_seen` is true, so it surfaces as transient and the turn loop retries with fresh resolution. The leak was specifically the continuation succeeding.

crates/jp_llm/src/provider/anthropic/oauth.rs
JeanMertz: Fixed in `9edcc106`. Your reading is exact: `post_token` mapped every non-2xx to `Rejected`, and `is_rejection` was `matches!(self, Self::Rejected { .. })` — the status was captured and never consulted.

Named input is any Anthropic token-endpoint outage or throttle during a refresh. The damage spreads (persisted state outliving the process) and misleads (the notice reads *"session expired; run `jp provider auth login`"*, diagnosing a credential problem that doesn't exist) while routing the user onto per-token billing.

`is_rejection` is status-aware now: `400` (RFC 6749's `invalid_grant`), `401`, `403`. Everything else flows to the `ResolveError::Refresh` arm, which already had the right semantics.

The variant is renamed `Status`, since it carries any non-success answer and only `is_rejection` decides what one means — leaving it called `Rejected` would have made a `503` read as a refusal at every use site.

`test_a_throttle_or_outage_is_not_a_rejection` covers 408, 429, 500, 502, 503, 504.

crates/jp_llm/src/provider/anthropic/resolve.rs
JeanMertz: Fixed in `d66ba915`, and it's reachable more easily than the comment suggests.

You frame it as concurrent JP processes. It's also a **single-process** race: `jp query` spawns the title-generation task, which builds its own `Anthropic` and resolves independently of the main turn. Two resolutions of the same expired profile, milliseconds apart, in one invocation.

`CredentialStore::lock()` returns a `StoreGuard` that can be held across an `await`. `refresh_under_lock` takes it, re-reads, and returns early when `still_stale` finds the token already rotated or the expiry moved out — so the common case is a lock acquisition and no exchange. Acquisition goes through `spawn_blocking`, since the lock is now held across a ≤30s `TOKEN_TIMEOUT` and parking an executor thread on that would be worse than the race.

The compare-and-set in `persist_rotation` is gone rather than kept alongside: under the lock it was dead weight, and `rotate`/`retire` are now plain document edits. That's most of why the diff is larger than the fix sounds.

One thing I did not do: the concurrent-refresh test you asked for. Forcing two resolutions to interleave around a real lock acquisition needs a fixture that blocks inside the exchange, and a test that can't force the window would pass whether or not the lock is held — which is the vacuous shape I'd rather not add. The double-check itself (`still_stale`) is covered indirectly by the existing refresh tests. Happy to be argued into building the fixture if you think the guarantee needs pinning.

crates/jp_llm/src/provider/anthropic/resolve.rs
JeanMertz: Fixed in `d66ba915`. Confirmed: `record_outcome` returns immediately for anything that isn't `Profile(Some(_))`, so `resolve` re-walked, `env::var` still succeeded, `attempt.selected == Some(spent)`, and `advance` returned `None`.

Your second argument is the stronger one and is what moved this off `nit` for me. By severity alone it's contained and visible — one request, terminal error naming the chain. But the same branch swallows the case where `record_outcome`'s store write *fails*: it only `warn!`s, resolution re-lands on the same profile, and a legitimately switchable failure becomes terminal. That one has no workaround.

`Attempt` now carries a `HashSet<AuthEntry>` of entries tried this request, and `walk_chain` skips them. That replaces the `selected == spent` sentinel, which terminated for the wrong reason — it inferred "nothing moved" from "resolution agreed with itself".

Two tests: `test_a_refused_api_key_falls_through_to_a_profile` (red before: `left: Some(ApiKey), right: Some(Profile("personal"))` — your exact case), and `test_an_entry_is_tried_once_per_request`, which drives both failures with a generation that no longer matches so neither records anything, and still expects the chain to advance and then terminate.

crates/jp_llm/src/provider/anthropic/resolve.rs
JeanMertz: Fixed in `d66ba915`, though not by comparing the credential.

The inconsistency you spotted is internal to one file: `persist_rotation` and `retire` both recheck `holds_refresh_token` inside the locked mutation, while `record_outcome` and `QuotaWatch::record` address the profile by name alone. Same store, same race, two standards.

One correction to the framing. The two paths you group have very different windows. A `401` makes `is_retryable()` false, so `advance` runs immediately — milliseconds. A quota rejection on a long streaming turn is minutes, and that's the case actually worth guarding.

On carrying "the attempted credential": I started there, since `Attempt` already holds it, and backed out. A legitimate mid-turn refresh rotates `access_token` on the *same* profile, so a secret-compare would skip recording a real cooldown — trading your rare false positive for a routine false negative.

So it's your generation counter. `StoredCredential.generation` is raised by `insert_profile` (a login replaced it) and left alone by `rotate` (a refresh did not). `Selected` carries the generation it resolved at, and `update_profile` compares under the lock, returning `Stale` rather than writing. Applied to `record_outcome` and `QuotaWatch::record` both, which closes the second half you asked for.

`test_a_replaced_profile_is_not_marked_by_an_in_flight_request` is your scenario: resolve, log in again, then let the old request return 401. Red before the fix (`needs_relogin` true on the fresh credential), green after. `test_a_stale_generation_does_not_cool_down_a_replaced_account` covers the quota half, and `test_a_refresh_does_not_invalidate_an_in_flight_request` pins the inverse — a rotation must *not* look like a replacement.

The field defaults to `0`, so an existing store reads without migration.

crates/jp_credentials/src/store.rs
```diff
@@ -0,0 +1,685 @@
...
+        let profiles = self
+            .credentials
+            .entry(category.to_owned())
+            .or_default()
+            .entry(provider.to_owned())
+            .or_default();
+
+        credential.generation = profiles
+            .get(profile)
+            .map_or(0, |replaced| replaced.generation.saturating_add(1));
```
JeanMertz: blocking: Preserve credential identity across logout followed by login. A request can resolve `personal` at generation 0, then the user runs `jp provider auth logout llm.anthropic --profile personal` and logs in under the same name while that request is pending. `remove_profile` deletes the counter, so this insertion assigns generation 0 again. The delayed request's cooldown or re-login write then passes `update_profile`'s equality check and disables the fresh credential across subsequent invocations.

Use an incarnation identifier that is not reused after deletion, or allocate generations from a persisted document-wide counter. Add a remove/reinsert regression test alongside the replacement test: outcome writes carrying the removed credential's generation must return `Superseded` and leave the new profile untouched.

crates/jp_llm/src/provider/anthropic/resolve.rs
```diff
@@ -0,0 +1,848 @@
...
+        debug!(
+            profile,
+            "Access token was refreshed while this request waited for the lock."
+        );
+        return Ok(());
+    };
+
+    debug!(profile, "Access token expired; refreshing.");
+
+    match oauth::refresh(&refresh_token).await {
```
JeanMertz: theoretical: Keep an in-flight refresh transaction alive when its caller is cancelled. `TitleGeneratorTask::run` selects cancellation against `update_title`, which can be awaiting this exchange through `model_details`. If cancellation wins after the endpoint rotates the refresh token but before its response is consumed, this future and its guard are dropped without persisting the replacement. The next invocation submits the retired token and can mark the profile for re-login.

This is not reachable through a normal login in this PR, which only creates static `Token` records; it matters before browser OAuth is exposed. Run the lock/recheck/exchange/persist transaction in an independently owned task that graceful cancellation drains, rather than tying its lifetime to a title or stream consumer. A mock exchange paused after accepting the refresh can test cancellation without relying on timing.

crates/jp_llm/src/provider/anthropic/resolve.rs
```diff
@@ -0,0 +1,848 @@
...
+    config: &AnthropicConfig,
+    store: Option<&CredentialStore>,
+    spent: &Attempt,
+    error: &StreamError,
+    model: &str,
+    now: DateTime<Utc>,
+) -> Option<Attempt> {
+    let selected = spent.selected.as_ref()?;
+
+    record_outcome(store, selected, error, now);
```
JeanMertz: theoretical: Offload outcome writes as well as refresh lock acquisition. `record_outcome` calls the synchronous `CredentialStore::mutate`, which blocks in `ResourceLocker::lock`; `QuotaWatch::observe` reaches the same lock directly from stream polling. A concurrent refresh holds this global lock across its network await. On a runtime whose worker threads are occupied by these blocking writers, the refresh cannot receive its response or run its timeout, so neither side releases the other.

The long-held-lock scenario needs an OAuth record, which the setup-token login shipped here does not create, so this does not gate that login path. Before enabling browser OAuth, put these mutations on `spawn_blocking` too, or make them await the offloaded `acquire` path. Test a pending refresh competing with an outcome write on a single-worker runtime.

crates/jp_llm/src/event_builder.rs
JeanMertz: blocking: Retaining notices across successful retries fixes part of this, and `structured_data` should remain a payload extractor. Delivery still needs to be addressed before treating the thread as resolved.

The later main-turn notice is not guaranteed: `title::resolve_model` permits an Anthropic title model alongside a non-Anthropic assistant model. In that setup, a title request falling back to `api_key` is silently billed, and later main turns never resolve the Anthropic chain. Summarization and inquiries also have this gap, so the exposure is not limited to a short title. Additionally, `collect_with_retry` still drops its accumulated notices when it ultimately returns `Err`.

Threading a reporting sink has a real cost, but recording a ticket does not make this hidden behavior visible. Deliver notices as they are consumed, including on terminal failure; keep terminal rendering outside `jp_llm`.

crates/jp_credentials/src/store.rs
JeanMertz: Fixed, and your reading is exact. `insert_profile` derived the generation from the profile it replaced, so after `remove_profile` it started again at 0.

`StoreDocument` now has a document-wide `next_generation` counter, and `insert_profile` assigns `max(replaced + 1, next_generation)`. The replaced profile's own generation stays as a floor, so a document whose counter lags its profiles still moves forward. The field is `#[serde(default)]`, and the store is new in this PR, so there's nothing to migrate.

`test_a_removed_and_restored_profile_does_not_inherit_the_old_credentials_state` is your scenario: resolve at gen 0, `remove_profile`, `insert_profile`, then `mark_needs_relogin` and `record_cooldown` at gen 0. Both return `Superseded` and the new profile is untouched. I checked it fails without the counter: `left: Updated, right: Superseded`.

crates/jp_llm/src/event_builder.rs
JeanMertz: You're right, and the title-model point is the one that undoes my "contained" argument. An Anthropic title model beside a non-Anthropic assistant never walks the Anthropic chain on the main turn, so the user never learns about the switch. The `Err` path dropping the accumulated notices was a real hole too.

The notices are now delivered as they're consumed:

- `jp_llm::event::NoticeSink` wraps an `Arc<dyn Fn(&str) + Send + Sync>`. `collect_with_retry` takes one and hands it each `Event::Notice` the moment it's read. Notices no longer appear in the returned events, so the accumulate-and-prepend logic is gone. Failed attempts and a request that ends in `Err` are covered because delivery happens before the outcome is known.
- `jp_llm` does no rendering. `jp_cli::output::notice_sink(&printer)` builds a sink that `eprintln`s as chrome.
- Callers wired: title generation (the background `TitleGeneratorTask` and `jp c title`), inquiries (on `LlmInquiryBackend`), and summarization (`jp c compact`, `jp q --compact`, `jp c fork --compact`). Summarize's `tracing::warn!` arm is gone.
- Teardown drains tasks before `Ctx` drops the printer, so the title task can still print.

Tests:
- `collect_with_retry_delivers_notices_when_the_request_fails` covers the terminal `Err`.
- `collect_with_retry_delivers_notices_from_an_attempt_that_failed` checks the sink's contents and that the returned events are exactly the successful attempt's content.
- `generate_delivers_provider_notices` and `llm_backend_delivers_provider_notices` cover title and inquiry.
- `notice_sink_prints_to_stderr` asserts the exact stderr line and an empty stdout, with no logging enabled.

T-0qch0c6 is closed.

crates/jp_llm/src/provider/anthropic.rs
JeanMertz: Agreed. I took your second option: explicit failure rather than re-resolving, because re-resolving runs into the cross-account signature problem.

It's worse than one request, too. `EscalatingNudge` allows up to `SOFT_FORCE_MAX_RETRIES = 3` follow-ups, all on the spent credential.

`call` now records `window_spent` from the admission headers, and it gates both follow-ups. When the model finishes without the required tool on a spent window, it yields `Event::Notice("the model did not call the required tool; not retrying, since the subscription window is spent and a retry would be billed as extra usage")` and then `Finished(Completed)`. The comment that justified keeping the retry is rewritten.

`a_spent_window_stops_the_forced_tool_retry` covers it: 200 with `anthropic-ratelimit-unified-status: rejected`, the body ends without a tool call, and an `EscalatingNudge` fallback is set. It asserts exactly one request and the notice followed by `Finished(Completed)`. I checked it fails with the guard disabled: 4 requests.

crates/jp_llm/src/provider/anthropic/resolve.rs
JeanMertz: Agreed on the mechanism, and on it being unreachable here: the only production path to a `CredentialSecret::Oauth` record is `rotate`, which needs one to exist already. Setup-token login stores `Token`.

Deferred to T-0qe01vw, which is marked as blocking RFD 090 Phase 3. The fix is the one you describe: the lock, recheck, exchange and persist steps run in a `tokio::spawn`ed task, and the caller awaits its `JoinHandle`, so cancelling the caller doesn't cancel the transaction. The ticket also records the test: a mock exchange paused after accepting the refresh.

crates/jp_llm/src/provider/anthropic/resolve.rs
JeanMertz: Correct, and `build_runtime` honours a user-set thread count, so `--threads 1` is enough to reach it once OAuth records exist. Same gating as the cancellation issue: the lock is only held for long during a refresh, and no shipped login creates a record that refreshes.

Tracked in the same ticket, T-0qe01vw, which blocks Phase 3. The plan: `record_outcome` goes through `spawn_blocking` like `acquire` already does, `QuotaWatch::observe` becomes async to do the same, and a single-worker-runtime test has a pending refresh compete with an outcome write.

crates/jp_llm/src/provider/anthropic/resolve.rs
JeanMertz: A correction to my earlier reply on this thread, which got the names wrong:

- The outcome is `UpdateOutcome::Superseded`, not `Stale`.
- The tests are `test_a_replaced_profile_does_not_inherit_the_old_credentials_state`, which covers both the re-login and the cooldown half, and `test_only_a_replacement_raises_the_generation`, which covers "a rotation is not a replacement". The three names I cited don't exist.

The logout-then-login hole you found separately (`4081028389`) is now closed with a document-wide generation counter, and `test_a_removed_and_restored_profile_does_not_inherit_the_old_credentials_state` pins it.

crates/contrib/async-anthropic/src/errors.rs
```diff
@@ -0,0 +1,359 @@
...
+
+    /// Whether this rejection is a quota limit rather than throttling.
+    ///
+    /// A rejection that names the exhausted window, or reports on paid
+    /// spillover, is about an allowance being spent.
+    /// One that carries neither is not a quota limit at all, however much its
+    /// status code resembles one.
+    #[must_use]
+    pub fn is_quota_rejection(&self) -> bool {
+        self.representative_claim.is_some() || self.overage_status.is_some()
```
JeanMertz: blocking: This predicate never looks at `status`, and `response_error` sends both `429` and `529` through it (`client.rs:342-362`). So a `529 overloaded_error` is classified as `SubscriptionExhausted` whenever it carries a `representative-claim` or `overage-status` header.

Named input: a subscription request answered with `529` during an Anthropic overload. The PR's premise is that the unified headers "ride on every response", and Claude Code's `computeNewLimitsFromHeaders` reads `representative-claim` and `overage-status` whatever the status is, including `allowed`. If a subscription `529` carries them too (I have not seen one captured, so this is the assumption), the same path:

- skips the in-place retry (`client.rs:211/255` and the turn loop, via `needs_credential_switch`),
- records a cooldown until `unified-reset`, which is the window reset and can be days away (capped at 7 days),
- moves the user to the next entry, which is often `api_key`, with the notice "Subscription limit reached".

The damage outlives the process (persisted cooldown) and the notice gives the wrong reason (it says the limit was reached; the server was overloaded). The user has no way to tell a billing switch happened for the wrong reason.

Claude Code applies its header check only to `429`, and only to pick a message. JP acts on it and spends money. Suggested fix:

- require `self.is_rejected()` as well as a claim or overage status. `QuotaWatch::observe` already treats `status == "rejected"` as the signal.
- never quota-classify a `529`. Keep it on the plain `RateLimit`/transient path.

`test_only_quota_headers_make_a_rate_limit_exhaustion` would need `status: Some("rejected")` on its positive cases. Add a negative case: `status: allowed` plus a claim on a `529` stays `RateLimit`.

crates/contrib/async-anthropic/src/client.rs
```diff
@@ -0,0 +1,460 @@
...
+
+        return AnthropicError::RateLimit {
+            retry_after,
+            limits,
+        };
+    }
+
+    // The credential was refused. The status is what makes this actionable,
+    // and it is lost once the body is parsed into the generic error envelope.
+    if status == StatusCode::UNAUTHORIZED || status == StatusCode::FORBIDDEN {
```
JeanMertz: question: Every `403` becomes `AnthropicError::Auth`. `StreamError::from` turns that into `AuthRejected`, and `record_outcome` then persists `needs_relogin` for all models, so the profile stays skipped until the user runs `jp provider auth login` again.

Anthropic also uses `403 permission_error` for scope problems. The bootstrap measurement in the RFD is one: `scope requirement any_of(user:ccr_inference, user:profile)`. A setup token only has `user:inference`. Any endpoint JP reaches with that token that needs a wider scope would therefore retire a profile that works fine for inference. Two examples: `model_details`/`models()` go through `advance_or_fail`, and a beta feature could return a scope-gated `403`.

Have you confirmed that `/v1/models/{id}`, `/v1/models`, and every beta in `CLAUDE_CODE_BETAS` accept a `user:inference`-only token? If you haven't, I'd limit `needs_relogin` to the cases that clearly mean a dead credential: `401`, `OAuth token has been revoked`, and `OAuth authentication is currently not allowed`. Other `403`s would then skip for this request (the `tried` set already handles that) without persisting anything.

crates/jp_credentials/src/store.rs
```diff
@@ -0,0 +1,703 @@
...
+/// A window naming a model family covers only that family, whether it is
+/// recorded as the provider reports it (`seven_day_opus`) or as the bare family
+/// name (`opus`).
+///
+/// An unrecognized scope covers only models whose name contains it, so a window
+/// this build does not know about cannot silently take a whole account out of
+/// use.
+fn scope_covers(scope: &str, model: &str) -> bool {
+    match scope {
+        SCOPE_ACCOUNT | "five_hour" | "seven_day" => true,
```
JeanMertz: nit: `overage` is a `representative-claim` value in Claude Code (`RateLimitType` includes it, and the UI shows it as "extra usage limit"), but it isn't listed here. It falls into the unknown-scope branch, and `model.contains("overage")` is false for every model, so the cooldown gets recorded but never covers any model.

Named input: a subscription account with both its weekly window and its extra-usage allowance spent. The `429` names `overage`, the cooldown doesn't apply, and every later invocation tries this profile first, gets rejected at admission, and only then switches, printing a notice each time. That's contained (admission rejections aren't billed) and visible, but it makes the persisted cooldown useless for exactly the "everything is spent" case.

Suggest adding `"overage"` to the account-wide arm, plus one line in `test_active_cooldown_scoping`.

crates/jp_llm/src/provider/anthropic/resolve.rs
```diff
@@ -0,0 +1,848 @@
...
+            return vec![format!(
+                "{} spent; this request was billed as extra usage",
+                window_name(scope)
+            )];
+        }
+
+        limits
+            .warning()
+            .map(|window| vec![warning_notice(window)])
+            .unwrap_or_default()
```
JeanMertz: nit: Nothing dedupes these notices, so they repeat on every request.

- Once an account crosses a threshold, `surpassed-threshold` comes back on every response until the window resets, which can be days. Each request prints `weekly limit 82% used, resets …` again: every tool-call cycle in a turn, every continuation `call`, title generation, and inquiries.
- Skip notices from `walk_chain` (`skipping profile:personal cooling down until …`, `skipping api_key: environment variable … is not set`) are re-yielded at the top of every `chat_completion_stream`. A chain like `["api_key", "profile"]` with no key set prints the skip line on every request, permanently.

Named input: a 10-tool-call turn on an account at 80% of its weekly window prints 10+ identical stderr lines. Contained and visible, but the notices cover billing decisions, and repeating them trains users to ignore them.

One option: a process-wide `Mutex<HashSet<String>>` (or keyed on window + threshold) inside `QuotaWatch`/the provider, so each distinct notice prints once per invocation. The switch notice should stay unconditional.

crates/jp_cli/src/cmd/provider.rs
```diff
@@ -0,0 +1,484 @@
...
+fn read_setup_token(printer: &Printer, hint: &str) -> Result<String, Error> {
+    if io::stdin().is_terminal() {
+        printer.eprintln(format!("Paste the setup token and press Enter.\n{hint}"));
+    }
+
+    let mut raw = String::new();
+    io::stdin()
+        .lock()
+        .read_line(&mut raw)
+        .map_err(|error| Error::from(format!("failed to read setup token from stdin: {error}")))?;
```
JeanMertz: nit: On a TTY this reads the token with terminal echo on, so the pasted `sk-ant-oat01-…` stays in scrollback, tmux history, and any screen recording or screen share. The doc comment names shell history and `ps` as the leaks being avoided, and this is the same kind of leak.

`jp_inquire::prompt::PromptBackend::password` already exists (`TerminalPromptBackend` wraps `inquire::Password`). Use it when `stdin().is_terminal()`, and keep `read_line` for the piped case, which has no echo anyway.

docs/rfd/090-anthropic-subscription-auth-with-credential-fallback.md
```diff
@@ -56,135 +57,383 @@ account under two profiles, keyed on the account UUID — never email alone.
...
+The lock-mutate-persist cycle lives in the `jp_credentials` crate:
+`CredentialStore` owns the semantics every backend shares — the document
+encoding, the schema-version check, and the mutation cycle — while storage
+backends implement the `keyring-core` store interface, holding the document as a
+single entry.
+A file-based store is the baseline backend, and the macOS keyring (Phase 4)
+swaps in behind the same interface.
+A file-based backend must replace its file atomically (temp file and rename) and
+create it with `0600` permissions; if `keyring-core`'s bundled file store does
+not provide both, JP's file persistence implements the store interface itself.
```
JeanMertz: nit: This says, in the present tense, that storage backends implement the `keyring-core` store interface, and Phase 2c says the store "adopts" it. The code doesn't: `jp_credentials` defines its own `CredentialBackend` trait (`describe`/`load`/`persist`), and no crate depends on `keyring-core`.

Either design is fine, but the RFD should match the code. Either say the `keyring-core` adoption is deferred to Phase 4, or describe `CredentialBackend` as the interface and note that Phase 4 adapts `keyring-core` to it.

crates/contrib/async-anthropic/src/errors.rs
JeanMertz: Fixed, but only half as you suggested.

**529:** taken as proposed. `response_error` still parses the unified headers on a `529` and logs them, but hands back an empty `UnifiedRateLimit`, so the error stays a plain retryable `RateLimit`. All three consumers read the classification from there (the `.when` retry filter, `is_switchable`, and `StreamError::from`), so none of them can switch credentials or record a cooldown over an overload any more.

**Requiring `is_rejected()`:** not taken, because that header can't be relied on. Claude Code's 429 handler (`src/services/api/errors.ts`) never reads `anthropic-ratelimit-unified-status` on a rejection. It hardcodes `status: 'rejected'` and keys only on claim/overage. JP has never captured a real exhaustion 429, so if that header turns out to be missing there, requiring it would turn every real exhaustion into an in-place retry and fallback would stop working.

Your concern still applies to 429s, so the guard runs the other way: `is_quota_rejection` returns false when the status *explicitly* says the allowance is available (`allowed`, `allowed_warning`), and a missing status doesn't count against it.

Tests:
- `test_overloaded_with_quota_headers_is_retried` (wiremock): a `529` carrying `representative-claim` and `overage-status`, then a `200`. It fails without the fix: the first call ends the request with `RateLimit { limits: { representative_claim: Some("five_hour"), … } }`.
- `test_available_allowance_is_not_a_quota_rejection`: both available statuses combined with a claim, plus the no-status case.
- A negative case in `test_only_quota_headers_make_a_rate_limit_exhaustion`: `status: allowed` plus a claim stays `RateLimit` and doesn't switch.

RFD Phase 1 documents both refinements.

crates/contrib/async-anthropic/src/client.rs
JeanMertz: Mostly yes, and I'd keep the current behaviour. Endpoint by endpoint:

- **`/v1/models/{id}`:** confirmed. `run_turn` calls `model_details` (`query.rs:1134`) before the turn loop starts, and the RFD's Phase 1 measurement is a setup-token turn that completed. That can't happen if this endpoint answers 403 to a `user:inference` token.
- **`CLAUDE_CODE_BETAS`:** confirmed by the same measurement, which sent the complete fingerprint including that beta set.
- **`/v1/models` (list):** unmeasured, but Anthropic's `models()` has no production caller. The only uses are `mock_tests.rs` and `test.rs`.
- **User-added `beta_headers`:** unmeasured. The one plan-gated beta I can find in Claude Code (afk-mode) is rejected with a `400`, not a `403`.

Claude Code also treats every other 401/403 as "Please run /login", and the RFD adopts that on purpose ("Any other `401`/`403` is treated the same way").

If a scope-gated 403 does turn up, that profile shows "needs re-login" in `jp provider auth list`, with a notice naming it. That's visible and limited to the one profile.

What would change my mind: a real 403 from something other than the credential. At that point I'd persist `needs_relogin` only for 401 and the two known messages, and treat other 403s as switch-for-this-request, which the `tried` set already supports.

crates/jp_cli/src/cmd/provider.rs
JeanMertz: Fixed. On a TTY, `read_setup_token` prints the provider hint and reads the token through `TerminalPromptBackend::password` (`inquire::Password`, no confirmation), with the printer's prompt writer. Piped input still goes through `read_line`. Both paths feed the same `sanitize_setup_token`.

I didn't add a test: the TTY branch needs a pty to exercise, and the sanitizer it hands off to is already covered.

crates/jp_credentials/src/store.rs
JeanMertz: Fixed. Confirmed in Claude Code's `claudeAiLimits.ts`: `RateLimitType` includes `overage`, displayed as "extra usage limit".

- `scope_covers` treats `overage` as account-wide, alongside `five_hour` and `seven_day`.
- `window_name` maps it to "extra usage limit", so the notice uses the same words as Anthropic's UI.
- The `representative_claim` doc comment and the RFD's claim lists now include it.
- `test_active_cooldown_scoping` gained a case where an `overage` cooldown covers `claude-haiku-4-5`.

crates/jp_llm/src/provider/anthropic/resolve.rs
JeanMertz: Fixed, but tracked per provider rather than process-wide. A process-wide static would carry state from one test into the next, and a turn reuses one `Arc<dyn Provider>` across its tool-call cycles, so a per-provider record covers the case you describe.

- `Anthropic` holds a `SeenNotices` (a shared `HashSet`) next to `client_cache`.
- Skip notices go through it, so each distinct skip prints once per provider.
- The switch notice now lives in its own `Attempt::switch` field and is always shown, since each switch is a new event.
- `QuotaWatch` shares the same set and dedupes warnings by `(claim, surpassed_threshold)` rather than by text, because the text changes as the percentage climbs. A higher threshold is still reported. "Billed as extra usage" is reported on every request, because every one of those requests is billed.

Tests:
- `test_notices_for_a_standing_condition_are_reported_once`: a repeated threshold is silent, a new threshold is reported, and a spent window is reported both times.
- `test_skip_notices_are_reported_once_and_switches_always`.

Both fail with the dedupe disabled.

Title generation and inquiries build their own providers, so each can print a given notice at most once more.

docs/rfd/090-anthropic-subscription-auth-with-credential-fallback.md
JeanMertz: Fixed, and I went with the RFD matching the code. `CredentialBackend` is described as the interface in the Credential store section and in Phase 2c. Phase 4 and the Non-Goals bullet now say Phase 4 adapts `keyring-core`'s Apple store to that trait. I also dropped the sentence hedging on whether `keyring-core`'s file store provides atomic writes and `0600`, since JP's own file backend does both.

While in there I fixed a stale line in Phase 1 that said the classifier "keys on the error body". It keys on the quota headers.

Cargo.lock
```diff
@@ -171,19 +171,23 @@ checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
...
  "secrecy",
  "serde",
  "serde_json",
+ "test-log",
  "thiserror 2.0.20",
  "tokio",
  "tokio-stream",
  "tracing",
+ "uuid",
+ "wiremock",
```
JeanMertz: Why did moving this in-tree change the list of dependencies? I'd want this to stay the same or even ideally *reduce* the number of required dependencies.

Cargo.lock
```diff
@@ -5871,6 +5950,29 @@ version = "0.0.19"
...
+ "hyper-util",
+ "log",
+ "once_cell",
+ "regex",
+ "serde",
+ "serde_json",
+ "tokio",
+ "url",
+]
+
```
JeanMertz: We already have an http recording library, why do we need this one as well?

Cargo.lock
JeanMertz: Three reasons, and two of them were fixable.

1. **Dev-dependencies now show up.** Cargo doesn't resolve a git dependency's dev-dependencies, so the old lockfile never listed them. The upstream `Cargo.toml` at `8c12222` already declared `wiremock`, `async-trait` and `test-log`. Now that the crate is a workspace member, they get resolved.
2. **The transport change** swapped `reqwest-eventsource` for `eventsource-stream` + `futures`, which is how a successful response's headers became readable. Both were already in the graph through `jp_llm`, and `reqwest-eventsource` stays there too, so no packages were added.
3. **`uuid` with `v4`** was the one real addition. It pulled in a third `getrandom` (0.4.3) and a second `r-efi` (6.0.0). T-0jrnyy1 already tracks the duplicate `getrandom` versions.

What I changed:
- `uuid` is gone from both crates. The `x-client-request-id` is built from 16 bytes of `getrandom` 0.3 (already a workspace dependency) with the v4 version and variant bits set, so the header keeps its shape. `random_token` fills its 32 bytes directly, which also makes its "256 bits" doc comment true: two v4 UUIDs only carry 244.
- `tokio-stream` → `futures::Stream`, since `futures` was already a dependency. The `MockApp` wrapper trait and the `async-trait` dev-dependency are gone.
- `wiremock` → `httpmock`; see your other comment.

Net effect: every package this PR added to `Cargo.lock` is gone except `jp_credentials`. That covers `deadpool`, `deadpool-runtime`, `getrandom` 0.4.3, `hermit-abi`, `num_cpus`, `r-efi` 6.0.0 and `wiremock`, and `uuid`'s entry matches `main` again. The crate's normal dependencies went from 11 to 12 edges (`+eventsource-stream`, `+futures`, `+getrandom`, `-reqwest-eventsource`, `-tokio-stream`), all to crates that were already in the graph.

Tests: `test_bearer_mode_headers` now checks the request id is a well-formed v4 UUID (it fails with the version nibble set wrong), and `test_bearer_mode_request_ids_differ` checks two requests get different ids.

Cargo.lock
JeanMertz: No good reason. I brought the upstream tests over as they were. They now use the workspace's `httpmock`, directly rather than through `jp_test`, so a `contrib/` crate doesn't depend on a `jp_*` one.

- `tests/bearer.rs` captures the request headers with `then.respond_with` and keeps the exact per-header assertions.
- The retry sequences in `tests/messages.rs` (3×429 then 200, 529 then 200) use a counting `respond_with`. It runs once per request, so the counts are exact, unlike a matcher closure that httpmock may evaluate more than once.
- Where wiremock verified call counts on drop, the tests now call `assert_calls_async(n)`. `test_with_backoff_basic` was `expect(1..)` and is now exactly 4 (the first attempt plus the backoff's three retries).
- The 529 test still fails with the overload guard disabled.

`wiremock` is removed from the workspace, along with the `cargo vet` audits and imports that existed only for it and for the `uuid` → `getrandom` 0.4 chain: `deadpool`, `deadpool-runtime`, `wiremock`, `num_cpus`, `hermit-abi`, `getrandom` 0.4 and `r-efi` 6.0.

PR #1152, opened 2026-09-10
feat(llm, config, cli): Add ChatGPT subscription auth for OpenAI

JeanMertz:
A ChatGPT Plus or Pro subscription can now serve a turn. `jp provider llm
auth login openai` stores the credential; an `auth` chain decides which
credential a request bills to, falling through to the next when one is
spent.

Auth chain entries name a kind and a credential: `api_key`, `api_key:work`,
`subscription`, `subscription:personal`, or a bare name resolved against
both. `api` and `sub` are accepted for the kinds. `api_key_env` takes a map
as well as a string, so a provider can hold several named keys, and every
provider has a chain rather than only the two selling a subscription.

Subscription requests reach ChatGPT's Codex host, which accepts a narrower
request than the platform API: the system prompt moves into `instructions`,
seven parameters it rejects are dropped, and explicit cache breakpoints are
removed. It answers a 200 carrying an event stream with no content type,
which the client now reads. When a plan's usage window is spent, JP redeems
one of the account's reset credits before falling through to per-token
billing.

Credential commands move from `jp provider auth` to `jp provider llm auth`,
so a provider is named plainly rather than by category, and `--profile`
becomes `--name`. `auth list` covers every provider, marks a fault in
place, and reports an expiry as a duration; its JSON payload carries values
rather than the table's prose.

A provider billed two ways records each scenario twice, and the two
recordings are compared: what JP sent, and what JP recorded. Neither
comparison includes the model's wording, its chosen tool arguments, its
reasoning, or the ids the host mints. Normalizing a wire dialect is the
provider's own job, and the default panics rather than passing the body
through, so a provider gaining a second route fails until it supplies a
projection.


crates/jp_llm/src/provider/openai/resolve.rs
```diff
@@ -0,0 +1,748 @@
...
+                profile,
+                refresh_token,
+                generation,
+            } => (profile, refresh_token, generation),
+        };
+
+        debug!(profile, "Access token expired; refreshing.");
+
+        match oauth::refresh(&refresh_token).await {
+            Ok(tokens) => persist_rotation(store, &profile, &refresh_token, &tokens),
```
JeanMertz: blocking: Hold the credential-store lock across the refresh exchange. Two JP invocations using the same expired subscription can both read the same refresh token and reach this call before either persists its replacement. Checking `holds_refresh_token` inside `persist_rotation` is too late: both exchanges have already happened. If the second exchange is rejected after the first persists, `retire` still matches the unchanged generation and marks the refreshed credential as needing re-login.

Follow the existing Anthropic `refresh_under_lock` pattern: acquire off the executor, reload and recheck staleness under the lock, exchange, and persist before releasing it. Add a concurrent-refresh regression test that verifies only one exchange occurs.

crates/jp_llm/src/provider/openai/oauth.rs
```diff
@@ -0,0 +1,551 @@
...
+
+impl OauthError {
+    /// Whether the grant itself was refused, as opposed to the request not
+    /// getting through.
+    ///
+    /// Only a refusal means the stored credential is dead; a transport failure
+    /// is worth another attempt later.
+    #[must_use]
+    pub fn is_rejection(&self) -> bool {
+        matches!(self, Self::Rejected { .. })
```
JeanMertz: blocking: A `429` or `503` from `/oauth/token` must not retire the stored credential. `post_token` wraps every non-success HTTP status in `Rejected`, so this predicate treats rate limiting and an OAuth service outage as a dead refresh token. `resolve` then persists `needs_relogin`, and subsequent invocations skip a valid subscription even after the outage ends, potentially using paid API fallback.

Restrict this predicate to grant/authentication rejection statuses or recognized OAuth error codes. Keep transient statuses recoverable without changing credential state, and cover `429` and `503` alongside a genuine invalid-grant response.

crates/jp_llm/src/provider/openai/resolve.rs
```diff
@@ -0,0 +1,748 @@
...
+            stored.email = tokens.identity.email.clone();
+        }
+
+        Ok(true)
+    });
+
+    match result {
+        Ok(true) => {}
+        Ok(false) => debug!(profile, "Rotated tokens were not persisted."),
+        Err(error) => warn!(%error, profile, "Could not persist rotated tokens."),
```
JeanMertz: blocking: Propagate a failed rotation write rather than continuing resolution. If the refresh succeeds but persistence fails, for example because the credential filesystem is full, this logs a warning and discards the new token pair. The next loop iteration reloads the expired pair and presents the already-consumed refresh token again. That can trigger a reuse rejection and leave later invocations with an unusable session; the reported chain failure also hides the original storage error.

Return a `Result` from `persist_rotation` and stop resolution on persistence failure. Any persistence retry should reuse the token pair already obtained, not perform another refresh. Add a failing-backend test asserting that the consumed refresh token is not submitted twice.

crates/jp_llm/src/cross_route_tests.rs
```diff
@@ -0,0 +1,232 @@
...
+        "InquiryResponse",
+    ] {
+        if trimmed.starts_with(marker) {
+            return Some(marker.to_owned());
+        }
+    }
+
+    // JP derives these from the schema it sent.
+    for field in ["name:", "id:", "type:"] {
+        if let Some(value) = trimmed.strip_prefix(field) {
```
JeanMertz: blocking: Parse the snapshot's JSON body before projecting events. The checked-in `test_tool_call_auto__conversation_stream.snap` files contain lines such as `"type": "tool_call_request"` and `"name": "run_me"`, not `ToolCallRequest` or `name:`. None of these match this function, so the inspected API and subscription snapshots both project to `[]`. Deleting a tool result or changing a tool name therefore still passes `outcome_parity`.

Strip the complete Insta header, deserialize the JSON, and project its `events` array with explicit normalization for host-generated IDs and model-authored content. Assert the projection is nonempty for these scenarios and add a test where removing a tool response makes the comparison fail.

crates/jp_llm/src/provider/openai/usage.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    if !response.status().is_success() {
+        debug!(status = %response.status(), "Redeeming a reset credit was refused.");
+        return false;
+    }
+
+    match response.json::<ConsumeResponse>().await {
+        // A body that parses but reports nothing is taken at its word: the
+        // request succeeded, so the window is open.
+        Ok(consumed) => consumed.success.unwrap_or(true),
```
JeanMertz: blocking: Decode the endpoint's `code` rather than assuming an omitted `success` means redemption succeeded. The Codex backend client models this response as `{ "code": "reset" | "nothing_to_reset" | "no_credit" | "already_redeemed", "windows_reset": ... }` ([upstream types](https://github.com/openai/codex/blob/main/codex-rs/backend-client/src/types.rs)). For example, if another client consumes the credit between JP's list and consume calls, `{ "code": "no_credit", "windows_reset": 0 }` deserializes here with `success: None` and returns `true`.

JP consequently announces that a reset was redeemed even though no window reopened. Model the actual response codes, return success only for an outcome that establishes recovery, and test both successful and unsuccessful HTTP-200 payloads.

crates/jp_llm/src/provider/openai.rs
```diff
@@ -85,16 +116,347 @@ const TOOL_CALL_KEEPALIVE_INTERVAL: Duration = Duration::from_secs(5);
...
+    /// credit to spend or the redemption was refused — in which case the
+    /// caller falls through to the next credential as it would have anyway.
+    async fn redeem_reset_credit(
+        &self,
+        attempt: &resolve::Attempt,
+        session_id: &str,
+    ) -> Option<String> {
+        let (_, http) = self.clients(attempt, session_id).ok()?;
+        let credits = usage::reset_credits(&http, &self.codex_base_url).await?;
+        let credit = credits.credits.first()?;
```
JeanMertz: blocking: Select an available reset credit, not the first history entry. The backend's credit list includes a `status` (`available`, `redeeming`, or `redeemed`), which `ResetCredit` currently discards. For an account whose list starts with a redeemed credit followed by an available one, this repeatedly requests the spent credit and eventually falls through to paid API billing while a usable reset remains.

Preserve the credit status and select an available, applicable credit. Use the backend's available count for the notice rather than the total list size. Add a response fixture with a redeemed entry before an available entry; Codex's account processor explicitly handles these different statuses.

crates/jp_llm/src/provider/openai.rs
```diff
@@ -85,16 +116,347 @@ const TOOL_CALL_KEEPALIVE_INTERVAL: Duration = Duration::from_secs(5);
...
+
+        // The session identifies the redemption, so a retried turn spends the
+        // same credit once rather than one per attempt.
+        let redeemed = usage::consume_reset_credit(
+            &http,
+            &self.codex_base_url,
+            session_id,
+            credit.id.as_deref(),
```
JeanMertz: blocking: Give each logical redemption its own idempotency key. `session_id` is derived solely from the conversation's creation time, so every turn in a conversation sends the same `redeem_request_id`. A long-running conversation that consumes one reset and later exhausts another window will reuse the first redemption's ID, even when selecting a different credit. The endpoint can then replay or reject the earlier redemption rather than reopen the window, causing API fallback while reset credits remain.

Keep the conversation-stable session header for caching, but use a separate redemption identity that stays stable across retries of that redemption and differs for a later redemption. Test two exhaustion/recovery cycles in the same conversation.

crates/jp_llm/src/provider/openai/resolve.rs
```diff
@@ -0,0 +1,748 @@
...
+            return None;
+        }
+    };
+
+    // Resolution landing on the same entry means nothing was recorded that
+    // could move it: `api_key` has no store entry, so an exhausted key falls
+    // through instead of being retried forever.
+    if attempt.selected.as_ref() == Some(spent) {
+        debug!("Credential chain did not advance; treating the failure as terminal.");
+        return None;
```
JeanMertz: nit: A rejected API key cannot advance to a later credential here. With `auth = ["api_key:work", "api_key:personal"]`, both variables set, and the work key returning `401` or insufficient quota, `record_outcome` records nothing for the API key. Re-resolution selects work again and this branch terminates the request without trying personal. The same happens with a subscription behind a rejected key.

Carry a request-local set of attempted credentials through resolution, as Anthropic's resolver does. This also permits fallback when a best-effort cooldown write fails. Add a regression test where a refused API key advances to the configured next entry. The failure is visible and contained to the request, so this is non-blocking under the review severity rules.

crates/jp_config/src/types/api_key_env.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+}
+
+impl Schematic for ApiKeyEnv {
+    fn schema_name() -> Option<String> {
+        Some("ApiKeyEnv".into())
+    }
+
+    fn build_schema(mut schema: SchemaBuilder) -> Schema {
+        schema.string_default()
```
JeanMertz: nit: Include the named-map form in the schema. A documented value such as `api_key_env = { work = "WORK_OPENAI_KEY" }` deserializes successfully, but this schema advertises only a string. Editor integrations and schema validators consequently reject a valid configuration, and generators cannot discover the named-key form.

Emit a union of string and string-valued map, and add a schema assertion covering both accepted forms.

crates/jp_llm/src/provider/openai/rate_limits.rs
```diff
@@ -0,0 +1,261 @@
...
+/// The headers say which window, and when it reopens, which is the difference
+/// between a cooldown that expires on its own and one guessed from a default.
+///
+/// Headers that report no spent window leave the error alone: a request refused
+/// for another reason should not retire the credential that sent it.
+pub fn apply(error: &mut StreamError, headers: &HeaderMap) {
+    let snapshots = parse_all(headers);
+    let Some((snapshot, window)) = snapshots
+        .iter()
+        .find_map(|snapshot| snapshot.spent().map(|window| (snapshot, window)))
```
JeanMertz: nit: Choose a spent limit that applies to the requested model. The header parser sorts families by ID, so `bengalfox` precedes `codex`. If a `gpt-5.6-luna` request receives a quota rejection while both the Spark family and account-wide family report 100%, this selects the Spark family and records a `gpt-5.3-codex-spark` cooldown. Re-resolution for Luna then selects the same subscription, and `advance` aborts without reaching the API fallback.

Pass the requested model or the provider's active-limit attribution into this decision, and select the relevant account/model scope rather than the first spent family. Add a mixed-family rejection test using the header layout already present in `LIVE_HEADERS`.

crates/jp_llm/src/provider/openai/resolve.rs
```diff
@@ -0,0 +1,748 @@
...
+    #[error(
+        "providers.llm.openai.auth entry `profile` matches no stored profile; run `jp provider \
+         llm auth login openai` to create one"
+    )]
+    NoProfiles,
+
+    /// A bare `profile` entry with multiple stored profiles.
+    #[error(
+        "providers.llm.openai.auth entry `profile` is ambiguous: multiple profiles are stored \
+         ({}); name one with `profile:<name>`",
```
JeanMertz: nit: Suggest `subscription:<name>` in this diagnostic. A user with two stored subscriptions and `auth = ["subscription"]` is told to use `profile:<name>`, but `AuthEntry::from_str` explicitly rejects that spelling. Following the suggested fix therefore produces another configuration error.

Update the user-facing `profile` entry spellings in both subscription resolvers, and check the config-field login examples for the corresponding `jp provider llm auth login <provider>` command. Assert that the entry suggested by this ambiguity error actually parses.

crates/jp_llm/src/provider/openai/resolve.rs
JeanMertz: Fixed. The OpenAI resolver now uses the same approach as Anthropic's `refresh_under_lock`: the lock is taken off the executor (`spawn_blocking`), the document is re-read and `still_stale` rechecked under the lock, and the exchange and persist both happen before the lock is released. `persist_rotation`, `holds_refresh_token`, and the generation-guarded `retire` are gone. Retiring now edits the locked document, so it can no longer hit a credential another process just rotated.

The refresh is passed in as a parameter so tests can count exchanges. `test_concurrent_resolutions_refresh_once` runs two resolutions against one stale profile and asserts a single exchange. I checked that it fails (2 exchanges) with the under-lock recheck disabled.

crates/jp_llm/src/provider/openai/resolve.rs
JeanMertz: Fixed as part of the lock change above. `refresh_under_lock` ends with `guard.persist(&document)?`, so a failed write of the rotated pair returns `ResolveError::Store` and resolution stops. The spent refresh token is never presented again. I didn't add a persist retry: returning the error is enough to avoid the second send.

`test_a_failed_rotation_write_stops_resolution` uses a backend whose writes fail, and asserts a `Store` error and exactly one exchange.

crates/jp_llm/src/provider/openai/oauth.rs
JeanMertz: Fixed. `OauthError::Rejected` is renamed to `Status`, and `is_rejection` now matches only `400 | 401 | 403`, the same filter `anthropic/oauth.rs` uses. A 429 or 5xx from `/oauth/token` comes back as `ResolveError::Refresh` and leaves the profile untouched.

Tests: `test_only_a_refusal_is_a_rejection` covers 400/401/403, `test_a_throttle_or_outage_is_not_a_rejection` covers 408/429/500/502/503/504, and two resolver tests check the store side: a 400 sets `needs_relogin`, a 429 doesn't.

crates/jp_llm/src/provider/openai/resolve.rs
JeanMertz: Fixed. `Attempt` carries a request-local `tried: HashSet<AuthEntry>`, following Anthropic's resolver. `advance` adds the spent entry, and `walk_chain` skips tried entries under both the bare and the resolved spelling. The old "landed on the same entry, so terminal" check is gone.

Tests: `test_a_refused_api_key_falls_through_to_a_subscription` (a quota-exhausted `api_key` moves to the subscription behind it) and `test_an_entry_is_tried_once_per_request` (`api_key:work` → `api_key:personal` → terminal, with no retry of `work`).

crates/jp_llm/src/provider/openai.rs
JeanMertz: Fixed, but not by selecting a credit. The redemption now sends no `credit_id` at all, and the backend picks an available one. Codex does the same when it has no specific choice (`consume_rate_limit_reset_credit` without an id). That way JP never has to tell an available credit from a redeemed one.

`ResetCredits` now reads `available_count`. Redemption is skipped when it's 0, and the notice is built from `available_count - 1` rather than from the list length.

Tests: `the_available_count_ignores_redeemed_credits` (a fixture with a redeemed entry before an available one) and `a_redemption_leaves_the_choice_of_credit_to_the_backend` (asserts the exact request body).

crates/jp_llm/src/provider/openai.rs
JeanMertz: Fixed. Each redemption now gets its own key from `oauth::random_token()`, as Codex does with a fresh UUID per redemption. The session id stays on the session header for caching and is no longer used as `redeem_request_id`. I also removed the comment claiming the session id made retries idempotent.

`test_two_redemptions_in_one_conversation_use_distinct_keys` runs two redemptions with the same session against a local stand-in for the `wham` endpoints. It asserts two different keys, neither equal to the session id, and no `credit_id`.

crates/jp_llm/src/provider/openai/usage.rs
JeanMertz: Fixed. The response is now decoded as `{ code }` into a `ConsumeCode` enum (`reset`, `nothing_to_reset`, `no_credit`, `already_redeemed`, plus `#[serde(other)] Unknown`), matching the upstream type. Only `reset` and `already_redeemed` count as a reopened window. That's the same pair Codex treats as success, and with per-redemption keys `already_redeemed` can only mean a retry of this redemption. An unparseable body counts as no reset.

Tests: `a_reset_reopens_the_window` and `a_200_that_redeemed_nothing_does_not_reopen_the_window` (`no_credit`, `nothing_to_reset`, an unknown code, and `{}`).

crates/jp_llm/src/provider/openai/rate_limits.rs
JeanMertz: Fixed. `apply` now takes the requested model, which is passed through `start_streaming` into `map_error`. It only looks at the model's own family (matched on `limit_name`) and the account-wide `codex` family, preferring the narrower one. A spent family for another model is ignored.

While there I also found that `apply` would relabel any status error, including a 400 or 500, as `SubscriptionExhausted` whenever a relevant window happened to be spent. It now only acts on errors already classified as `RateLimit`, `InsufficientQuota` or `SubscriptionExhausted`.

Tests, using the `LIVE_HEADERS` layout with both families at 100%:
- Luna records `account`.
- Spark records `gpt-5.3-codex-spark`.
- With only Spark spent, a Luna error is left alone.
- A 400 or 500 keeps its kind.

crates/jp_llm/src/cross_route_tests.rs
JeanMertz: Fixed. `recorded_conversation` now strips the whole insta header (up to the second `---`; the old code stopped at the first) and deserializes the JSON body. It panics on an unreadable snapshot rather than skipping it. `project_conversation` then works on the `events` array:

- **Kept:** event order and type, `chat_request` content and schema, tool names, tool-response content and `is_error`, and the response variant.
- **Dropped:** timestamps, metadata, model prose, reasoning, and tool arguments.
- **Ids:** tool call ids go through `number_ids`, so call/result pairing is still compared.

One more thing had to change for a correct projection to pass: the subscription snapshots contain two `config_delta` events that the API ones don't. The harness writes them when it points the subscription route at its own model. Those are dropped too, since they describe the harness rather than the conversation. Consecutive identical `chat_response` variants are also collapsed, because the model chooses how many items to split its answer into.

`outcome_parity` now fails on an empty projection. The new `cross_route/outcome_tests.rs` checks that a dropped tool result, a renamed tool, and a flipped `is_error` are all caught. As a sanity check, I temporarily kept tool `arguments` in the projection: `outcome_parity` went red on 7 real scenarios across both providers.

crates/jp_llm/src/provider/openai/resolve.rs
JeanMertz: Fixed in both resolvers. The `UnknownProfile`, `NoProfiles` and `AmbiguousProfile` messages and the skip notices now say `subscription` / `subscription:<name>`. I fixed other stale command strings along the way:
- the `auth` field docs in `jp_config` (now `jp provider llm auth login <provider>`)
- the duplicate-account error in `jp_cli` (it was missing `llm`)
- two doc comments in `jp_credentials`

`test_every_suggested_entry_parses` (in both resolvers) pulls every `subscription…` spelling out of each error message and asserts that it parses as `AuthEntry::Subscription`.

crates/jp_config/src/types/api_key_env.rs
JeanMertz: Fixed. `ApiKeyEnv::build_schema` now emits a union of `string` and a `string → string` map. The `app_config_schema_shape` snapshot shows both variants for every provider that has `api_key_env`. `the_schema_accepts_a_variable_or_a_map_of_variables` asserts both variants and the map's key/value types.

crates/jp_config/src/providers/llm.rs
```diff
@@ -99,7 +102,188 @@ pub struct LlmProviderConfig {
...
+
+        match (kind, name) {
+            ("api_key" | "api", name) => Ok(Self::ApiKey(name)),
+            ("subscription" | "sub", name) => Ok(Self::Subscription(name)),
+
+            // A bare word that is not a kind names a credential. A typo lands
+            // here and is reported at resolution, which knows what exists.
+            (name, None) if !name.is_empty() => Ok(Self::Named(name.to_owned())),
+
+            _ => Err(AuthEntryParseError(s.to_owned())),
```
JeanMertz: blocking: Migrate stored Anthropic auth entries before rejecting `profile:<name>`. An existing conversation can contain `auth = ["profile:personal", "api_key"]` in its base config or a config delta. `jp_conversation::compat::deserialize_partial_config` handles this parse failure by removing the failing array element and retrying, so that conversation loads with `["api_key"]`. It then bills the API even when the subscription is usable, without a credential-switch notice explaining the lost preference. Updating the user's TOML does not repair the stored delta.

Either accept the legacy spelling and serialize it canonically, or migrate `profile` and `profile:<name>` in the stored-config compatibility pass. Add a conversation-load regression test that preserves the subscription-first ordering through load and save.

crates/jp_llm/src/provider/openai/usage.rs
```diff
@@ -0,0 +1,176 @@
...
+    if !response.status().is_success() {
+        debug!(status = %response.status(), "Redeeming a reset credit was refused.");
+        return false;
+    }
+
+    match response.text().await {
+        Ok(body) => window_reopened(&body),
+        Err(error) => {
+            debug!(%error, "Could not read the redemption result.");
+            false
```
JeanMertz: blocking: Distinguish an unknown redemption outcome from a confirmed refusal. If `/consume` spends the credit and reopens the window, but the connection drops while JP reads its response, this returns `false`. `chat_completion_stream` then advances using the original exhaustion error and persists its cooldown, excluding the recovered subscription from later requests and potentially billing the API while its reset allowance sits unused. The same ambiguity exists in the `.send()` error branch.

Use the redemption's existing idempotency key for a bounded retry, or recheck admission on the same subscription before recording exhaustion. If recovery remains uncertain, report that explicitly rather than persisting the pre-redemption quota state. Add a server test that commits the redemption and truncates its response.

crates/jp_llm/src/provider/openai.rs
```diff
@@ -1388,13 +2271,35 @@ async fn skip_unknown_events(
...
+fn map_error(error: OpenaiStreamError, model: &str) -> StreamError {
+    match error {
+        OpenaiStreamError::Status {
+            status,
+            headers,
+            body,
+        } => {
+            let mut error = StreamError::from_http(status, &headers, &body);
+            // The body says a limit was hit; the headers say which one and
+            // when it reopens, which is what a cooldown needs to expire on its
```
JeanMertz: nit: Classify an actual HTTP authentication rejection as `AuthRejected` before testing whether to advance. With a cached model and `auth = ["api_key:work", "api_key:personal"]`, a revoked work key returning HTTP 401 reaches `StreamError::from_http`, which classifies it as `Other`. `rate_limits::apply` leaves that unchanged, so `is_switchable` is false and personal is never attempted. An expired pasted subscription token has the same problem.

The resolver tests construct `StreamError::auth_rejected` directly, so they do not exercise this missing transport mapping. Add a local HTTP-401 test through the provider and map genuine credential refusals in the OpenAI adapter. The request fails visibly, so this is non-blocking.

crates/jp_llm/src/provider/openai.rs
```diff
@@ -85,16 +120,340 @@ const TOOL_CALL_KEEPALIVE_INTERVAL: Duration = Duration::from_secs(5);
...
+            // Refused here rather than at the host, which answers a model it
+            // does not serve with the same 404 it gives a typo. A model the
+            // table has never heard of is let through: the catalog ships with
+            // the binary and the plan's model set does not.
+            if details.subscription == Some(false) {
+                return Err(Error::UnsupportedForCredential(format!(
+                    "{name} is not served by a ChatGPT subscription; pick one of {}, or add \
+                     `api_key` to providers.llm.openai.auth",
+                    SUBSCRIPTION_MODELS.join(", ")
+                )));
```
JeanMertz: nit: Adding the API fallback suggested here does not make an API-only model reachable. For `auth = ["subscription:personal", "api_key"]`, a usable subscription, and `gpt-4.1`, resolution selects the subscription and this returns immediately even though the configured API key can serve the model. The chat path also resolves without considering model availability, so bypassing this metadata check alone would still send to the subscription host.

Make credential selection account for known model availability in both metadata lookup and chat admission, without marking the subscription as spent. Alternatively, if capability-based fallback is intentionally unsupported, tell the user to select an API-key-only chain rather than merely add an API entry. Cover the subscription-first chain with an API-only model.

crates/jp_config/src/providers/llm.rs
JeanMertz: Fixed by accepting the legacy spelling rather than adding a migration step. `AuthEntry::from_str` now parses `profile` and `profile:<name>` as `Subscription`, and `Display` writes them back as `subscription` / `subscription:<name>`, the same way `sub` is handled. The compat pass no longer has an unparseable entry to drop, and a TOML file still using the old spelling keeps working too.

A bare `profile` is covered as well. Before this it parsed as `Named("profile")` and failed at resolution with `no credential named 'profile'`.

Tests:
- `legacy_profile_auth_entries_survive_load_and_save` in `compat_tests.rs`: a stored `["profile:personal", "profile", "api_key"]` goes through `deserialize_partial_config`, keeps its order, and serializes as `["subscription:personal", "subscription", "api_key"]`. With the parser change disabled it fails with `["profile", "api_key"]`.
- The `from_str` and shorthand round-trip cases in `anthropic_tests.rs` now include both legacy forms.

crates/jp_llm/src/provider/openai/usage.rs
JeanMertz: Fixed, using your second option: check admission on the same subscription instead of retrying `/consume`.

`consume_reset_credit` now returns `Redemption::{Reopened, Refused, Unknown}`:
- `Reopened`: `reset` or `already_redeemed`.
- `Refused`: `no_credit`, `nothing_to_reset`, any non-5xx error status, or a connection that never opened (so no request was sent).
- `Unknown`: the request failed after it was sent, a 5xx, a body that couldn't be read, or an unrecognized code.

`redeem_reset_credit` treats `Unknown` like `Reopened`. The turn retries the same subscription and shows an explicit notice: "subscription limit reached, redeeming a usage reset went unconfirmed; retrying the subscription". If the window really reopened, the retry succeeds. If not, it hits exhaustion again, and since `redeemed` is already set it goes down the normal `advance` path. The cooldown is only recorded when a second exhaustion confirms it, never on the pre-redemption error.

Tests:
- `test_an_unanswered_redemption_retries_the_subscription`: the `wham` stand-in records the consume request and closes the socket without answering. The test asserts the request was sent and the unconfirmed notice comes back. With the post-send error mapped to `Refused` instead, it fails with `None`.
- `an_unreadable_outcome_is_unknown` and `a_200_that_redeemed_nothing_is_a_refusal` cover the body-level outcomes.

crates/jp_llm/src/provider/openai.rs
JeanMertz: Fixed. This went further than the comment suggested: nothing in the OpenAI adapter produced `AuthRejected` at all, so the refusal branches in `is_switchable` and `record_outcome` never ran for this provider.

Both request paths now map HTTP 401 to `StreamError::auth_rejected`:
- streaming: in `map_error`'s `Status` arm, before `rate_limits::apply`
- buffered: in `start_buffered`, where `create` reports the status through `error_for_status`

403 is left alone on purpose. OpenAI also returns it for region and organization restrictions, which no other credential on the same account would get past.

`auth_rejected_tests.rs` runs the provider against a local server that answers every request with 401, using `auth = ["api_key:work", "api_key:personal"]`. It asserts that both keys were sent, in order and only once each, and that the final error is `AuthRejected`. There are two variants, one per path (a model with `STREAMING_UNSUPPORTED` covers the buffered one). With either mapping disabled, only the `work` key is ever sent.

crates/jp_llm/src/provider/openai.rs
JeanMertz: Fixed with the first option: credential selection now takes the model into account. In `openai/resolve.rs::walk_chain`, a `subscription` entry is skipped when the catalog marks the requested model `subscription: Some(false)`. The skip notice reads "`subscription:<name>` does not serve `<model>`; only an `api_key` entry reaches it". Nothing is written to the store, so the same subscription still serves the next request for another model. This applies to metadata lookup and chat admission alike, because both go through `resolve`.

A model the catalog doesn't list is not skipped, since the plan's model set can change without a JP release. That includes the empty name preflight uses. To make the lookup quiet, the catalog match moved into `catalog_entry`, which returns `None` for an unknown id. `map_model` keeps its "Missing model details" warning.

With `["subscription:personal", "api_key"]`, `gpt-4.1` now resolves to the key. With a subscription-only chain it fails as `ChainExhausted` with the reason above, which makes the existing "add `api_key`" advice accurate.

Tests: `test_an_api_only_model_skips_the_subscription` (also checks that a plan model still lands on the subscription) and `test_an_api_only_model_on_a_subscription_only_chain_names_the_fix` (also checks that no cooldown or re-login marker is written). One existing cooldown test used `gpt-5.4-mini` as its "other model on the same subscription". That model is API-only, so the test now uses `gpt-5.5`.

crates/jp_llm/src/provider/openai.rs
```diff
@@ -1388,13 +2319,44 @@ async fn skip_unknown_events(
...
+            // it for region and organization restrictions that no other
+            // credential of the same account would clear.
+            if status == reqwest::StatusCode::UNAUTHORIZED {
+                return StreamError::auth_rejected(error.message().to_owned());
+            }
+
+            // The body says a limit was hit; the headers say which one and
+            // when it reopens, which is what a cooldown needs to expire on its
+            // own instead of on a guess.
+            rate_limits::apply(&mut error, &headers, model);
```
JeanMertz: nit: Recognize the subscription's typed exhaustion response even when it lacks window-utilization headers. For HTTP 429 with `{"error":{"type":"usage_limit_reached","plan_type":"pro","resets_at":1789571252}}` and no `x-codex-*-used-percent` headers, `from_http` produces `RateLimit` and `rate_limits::apply` leaves it unchanged. `is_switchable` then rejects it, so JP retries the spent subscription rather than attempting reset-credit redemption or the configured next credential. The eventual failure is visible and contained to this request, so this is non-blocking.

Codex's own adapter recognizes `error.type == "usage_limit_reached"` independently of its optional header snapshot ([upstream adapter](https://github.com/openai/codex/blob/main/codex-rs/codex-api/src/api_bridge.rs)). Classify that typed body in the OpenAI adapter before enriching it with header scope/reset information, and use the body's reset timestamp when header timing is absent. Add a transport-mapping test without utilization headers, while retaining ordinary rate-limit responses as retryable.

JeanMertz: Fixed. In `map_error`'s `Status` arm, a 429 whose body has `error.type == "usage_limit_reached"` now becomes `StreamError::subscription_exhausted`, with `quota_reset` taken from the body's `resets_at`. This happens before `rate_limits::apply` runs. When the `x-codex-*` headers report a spent window, `apply` still adds the scope and the header's reset time on top. When they don't, the body's reset time stands. So the error is switchable either way, and reaches reset-credit redemption and the next chain entry.

Only `type` and `resets_at` are read, through a small local `Deserialize`; any other 429 body falls through unchanged. I only changed the streaming path. The buffered path gets its error from `error_for_status`, which discards the body, and the only model on that path (`gpt-5.5-pro`) is API-only, so it never reaches the subscription host.

`usage_limit_tests.rs`:
- the typed body with no headers → `SubscriptionExhausted`, `quota_reset` = `resets_at`, no scope. With the new branch disabled this fails with `RateLimit`.
- the typed body plus a spent account window in the headers → scope `account`.
- a plain `rate_limit_exceeded` 429 → still `RateLimit`, so it stays retryable in place.

PR #1153, opened 2026-09-10
feat(cli): Add --auth to bill a turn to a chosen credential

JeanMertz:
`jp q --auth personal` bills the turn to that credential, `--auth api_key`
to per-token billing, and `--auth sub,api_key` tries the subscription
first. It takes the same entries as the `auth` chain in configuration.

The flag applies to the provider the turn's model belongs to and records a
config delta, so the rest of the conversation bills the same way until
another `--auth` changes it.

### DIFF
--- a/crates/jp_cli/src/cmd/query.rs
+++ b/crates/jp_cli/src/cmd/query.rs
@@ -86,9 +86,13 @@ use jp_config::{
         },
     },
     fs::{expand_tilde, load_partial},
-    model::parameters::{
-        PartialCustomReasoningConfig, PartialReasoningConfig, ReasoningConfig, ServiceTier,
+    model::{
+        id::{PartialModelIdOrAliasConfig, ProviderId},
+        parameters::{
+            PartialCustomReasoningConfig, PartialReasoningConfig, ReasoningConfig, ServiceTier,
+        },
     },
+    providers::llm::AuthEntry,
     style::{mcp_startup::McpStartupConfig, reasoning::ReasoningDisplayConfig},
 };
 use jp_conversation::{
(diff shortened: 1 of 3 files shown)


crates/jp_cli/src/cmd/query.rs
```diff
@@ -2621,6 +2640,92 @@ fn build_thread(
...
+/// reports the cycle properly later.
+fn provider_of(
+    id: &PartialModelIdOrAliasConfig,
+    aliases: &IndexMap<String, PartialModelIdOrAliasConfig>,
+    depth: u8,
+) -> Option<ProviderId> {
+    match id {
+        PartialModelIdOrAliasConfig::Id(id) => id.provider,
+        PartialModelIdOrAliasConfig::Alias(alias) if depth > 0 => {
+            provider_of(aliases.get(alias.as_str())?, aliases, depth - 1)
```
JeanMertz: nit: Reuse `PartialModelIdOrAliasConfig::finalize(aliases)` from `jp_config/src/model/id.rs` rather than maintaining another alias resolver here. It already follows partial alias chains and detects cycles, so the provider lookup can use `id.finalize(aliases).ok().and_then(|id| id.provider)`. This keeps `--model fast --auth personal` on the same resolution rules as configuration and removes the CLI-specific eight-hop limit. The CLI should choose where to write the auth chain, while `jp_config` owns how model aliases resolve.

JeanMertz: Done. `provider_of` is gone, and `active_provider` now calls `PartialModelIdOrAliasConfig::finalize(aliases)` and uses the resulting provider. That puts `--auth` on the same alias resolution as configuration: chains of any length, with cycles reported by `resolve_partial_alias_chain`.

The call uses path syntax (`PartialModelIdOrAliasConfig::finalize(&id, aliases)`) on purpose. `schematic::PartialConfig` is in scope in `query.rs`, and its `finalize(self, &Context)` takes `self` by value, so method resolution picks it over the inherent `&self` method. The path form also lets the `MergeableMap` deref-coerce to `&IndexMap`.

The unmapped-alias fallback in `finalize` (parsing the alias as `provider/name`) changes nothing here: `FromStr`/`From<&str>` already turn such strings into `Id`. All ten `test_auth_*` tests pass unchanged, including the cycle case.

PR #1159, opened 2026-09-10
fix(config): Align the schema with accepted inputs

JeanMertz:
Setting `conversation.default_id` to an explicit conversation ID no
longer breaks conversation persistence. `jp init` excludes the internal
`test` provider from its picker, and `assistant.model.id` accepts that
provider consistently in string and table form.

The schema lists enum values and aliases, constrains tool sources, and
describes both bare values and merge wrappers. It includes the boolean
and `"inherit"` forms of `dedup` and marks fields with defaults as
optional. Assignable keys, `--cfg` arguments, and `JP_CFG_*` names are
unchanged. Invalid compaction tool-call policies report
`Unknown enum variant x.` rather than an unknown tool_calls mode.

`ConfigEnum` keeps accepted spellings and schema definitions together,
with optional string-based serialization through the same parser.
This reduces the risk of advertising values that config loading rejects.
Stored-config repair uses the wrapper schema rather than special-casing
wrapped arrays.

Closes: 04hn6c8
Closes: 0dkct9c
Closes: 07vn4tn

### DIFF
--- a/crates/contrib/schematic_macros/src/common/field.rs
+++ b/crates/contrib/schematic_macros/src/common/field.rs
@@ -365,7 +365,11 @@ impl Field<'_> {
         let deprecated = map_option_field_quote("deprecated", extract_deprecated(&self.attrs));
         let env_var = map_option_field_quote("env_var", self.get_env_var());
 
-        let value = self.value;
+        // A `partial_via` field is written in the shape of its via type, not of
+        // the type it resolves to: `attachments` holds a `Vec<AttachmentConfig>`
+        // once resolved, but a document writes either a bare list or the merge
+        // wrapper carrying one, and the via type is what describes both.
+        let value = self.partial_via_ty.as_ref().unwrap_or(self.value);
         let mut inner_schema = if self.is_nested() {
             quote! { schema.infer_as_nested::<#value>() }
         } else {
--- a/crates/contrib/schematic_macros/src/common/field.rs
+++ b/crates/contrib/schematic_macros/src/common/field.rs
@@ -7,7 +7,7 @@ use syn::{Attribute, Expr, ExprPath, Field as NativeField, Type};
 
 use crate::{
     common::{FieldValue, PartialAttr, TypeInfo, extract_inner_type, macros::ContainerSerdeArgs},
-    utils::{extract_common_attrs, format_case, parse_default},
+    utils::{DefaultAttr, extract_common_attrs, format_case, parse_default},
 };
 
 // #[serde()]
@@ -43,7 +43,7 @@ pub struct FieldArgs {
 
     // config
     #[darling(with = parse_default)]
-    pub default: Option<Expr>,
+    pub default: DefaultAttr,
     #[cfg(feature = "env")]
     pub env: Option<String>,
     pub merge: Option<ExprPath>,
@@ -180,7 +180,7 @@ impl Field<'_> {
 
     #[cfg(feature = "schema")]
     pub fn is_optional(&self) -> bool {
-        self.serde_args.default || self.args.default.is_some()
+        self.serde_args.default || self.args.default.is_declared()
     }
 
     pub fn is_required(&self) -> bool {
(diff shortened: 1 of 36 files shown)


crates/contrib/schematic_macros/src/common/field.rs
```diff
@@ -357,14 +377,22 @@ impl Field<'_> {
...
         let env_var = map_option_field_quote("env_var", self.get_env_var());
 
-        let value = self.value;
+        // A `partial_via` field is written in the shape of its via type, not of
+        // the type it resolves to: `attachments` holds a `Vec<AttachmentConfig>`
+        // once resolved, but a document writes either a bare list or the merge
+        // wrapper carrying one, and the via type is what describes both.
+        let value = self.partial_via_ty.as_ref().unwrap_or(self.value);
         let mut inner_schema = if self.is_nested() {
             quote! { schema.infer_as_nested::<#value>() }
```
JeanMertz: nit: Preserve partialization of collection elements when inferring the via type. For example, `{"assistant":{"instructions":[{"title":"Review"}]}}` is accepted by `PartialAppConfig`, but its schema leaves `items` and `examples` on the instruction element non-optional. `infer_as_nested::<MergeableVec<InstructionsConfig>>()` marks the union as partial, not its array elements; `partialize_schema` subsequently reaches those elements with `force_partial = false`. The wrapper's `value` array has the same problem. The snapshot already shows this under the per-question `PartialAssistantConfig`, where the elements remain `InstructionsConfig` rather than `PartialInstructionsConfig`.

This is a schema-metadata issue, not a config-loading failure. Preserve the nested-element marker in both array forms and add an assertion on their optional/nullable field flags.

crates/contrib/schematic_macros/src/common/field.rs
```diff
@@ -357,14 +377,22 @@ impl Field<'_> {
...
         } else {
             quote! { schema.infer::<#value>() }
         };
 
-        if let Some(Expr::Lit(lit)) = &self.args.default {
+        if let Some(path) = &self.args.schema_union_with {
+            inner_schema = union_with_declared_shapes(path, &inner_schema);
```
JeanMertz: theoretical: A field combining `schema_union_with` with a literal `#[setting(default = ...)]` loses its additional input shapes. This block constructs the union, but the literal-default branch below replaces `inner_schema` with `infer_with_default::<#value>(...)` at line 416. For example, a boolean field with `default = true` and a custom deserializer accepting `"inherit"` would advertise only the boolean.

I have not identified a current JP input that triggers this: the dedup fields added here do not use literal defaults. Apply the union expansion after default inference so these supported field attributes compose, and cover that combination with a derive test.

crates/contrib/schematic_macros/src/common/field.rs
JeanMertz: Accepted and implemented locally on the follow-up branch. `Field::generate_schema_type` reuses the partial collection type from `FieldValue` for nested `partial_via` collections, so both the bare array and the wrapper's `value` array infer partial elements. This does not change global union partialization or deserialization.

Added a regression test that deserializes title-only instructions in both forms and checks the instruction fields' optional/nullable flags and partial type name in `PartialAppConfig::schema()`. It failed on the original metadata and passes with the fix. Refreshed the schema snapshot. The `jp_config`, `schematic`, and `jp_conversation` suites, plus CLI schema and init tests, pass.

The changes are not yet committed or pushed.

crates/contrib/schematic_macros/src/common/field.rs
JeanMertz: Accepted as a localized, non-blocking fix despite having no identified current JP input. Moved the `schema_union_with` expansion after literal-default inference, so the default cannot overwrite the additional shapes.

Added a `Config` derive regression test combining `default = true` with an additional `"inherit"` shape. It asserts the exact resulting schema, including the boolean default and expanded index, and checks that the field stays optional. The test failed before the move and passes afterward. Implemented locally on the follow-up branch; not yet committed or pushed.

PR #1160, opened 2026-09-10
fix(cli): Preserve conversation config clears

JeanMertz:
Fields cleared with `jp query --cfg assistant.name:=null` stay cleared on subsequent invocations, even when workspace config supplies a value. The next query no longer records a delta restoring that value.

Carry active conversation clears through base-layer filling before applying `--cfg` overrides. Explicit values still win, later sets retire clears, and resets discard earlier clears. Existing stored clears work without a data migration.

Fixes: T-0gqh7z7

### DIFF
--- a/crates/jp_cli/src/cmd/query.rs
+++ b/crates/jp_cli/src/cmd/query.rs
@@ -2431,10 +2431,15 @@ impl IntoPartialAppConfig for Query {
         partial: PartialAppConfig,
         _: Option<&PartialAppConfig>,
         handle: &ConversationHandle,
-    ) -> std::result::Result<PartialAppConfig, Box<dyn std::error::Error + Send + Sync>> {
-        let config = workspace.events(handle)?.config().map(|c| c.to_partial())?;
+    ) -> std::result::Result<
+        (PartialAppConfig, Vec<String>),
+        Box<dyn std::error::Error + Send + Sync>,
+    > {
+        let events = workspace.events(handle)?;
+        let config = events.config()?.to_partial();
+        let unsets = events.config_unsets()?;
 
-        load_partial(partial, config).map_err(Into::into)
+        Ok((load_partial(partial, config)?, unsets))
     }
 }
 
--- a/crates/jp_cli/src/config_pipeline.rs
+++ b/crates/jp_cli/src/config_pipeline.rs
@@ -24,7 +24,7 @@ use jp_config::{
 use jp_storage::backend::FsStorageBackend;
 use jp_workspace::Workspace;
 use relative_path::RelativePath;
-use tracing::{debug, error};
+use tracing::{debug, error, warn};
 
 use super::{CfgKeyword, KeyValueOrPath};
 use crate::error::{Error, Result};
@@ -314,12 +314,20 @@ impl ConfigPipeline {
     /// base only what it does not, which is what the snapshot already meant for
     /// every field that merges by replacement.
     ///
-    /// `--cfg` still merges on top, where combining is what the user asked for.
+    /// `unsets` carries the conversation's explicitly cleared paths.
+    /// These are cleared after filling so the base cannot restore them.
+    /// `--cfg` merges on top and can set them again.
     pub fn partial_with_conversation(
         &self,
         conversation: PartialAppConfig,
+        unsets: &[String],
     ) -> Result<PartialAppConfig> {
-        let partial = conversation.fill_from(self.base.clone());
+        let mut partial = conversation.fill_from(self.base.clone());
+        for path in unsets {
+            if let Err(error) = partial.unset(path) {
+                warn!(%path, %error, "Ignoring a config delta unset for an unknown field.");
+            }
+        }
         apply_cfg_args(partial, &self.cfg_args)
     }
 
--- a/crates/jp_cli/src/ctx.rs
+++ b/crates/jp_cli/src/ctx.rs
@@ -293,6 +293,7 @@ pub(crate) trait IntoPartialAppConfig {
         merged_config: Option<&PartialAppConfig>,
     ) -> std::result::Result<PartialAppConfig, Box<dyn std::error::Error + Send + Sync>>;
 
+    /// Apply the conversation layer and return its explicitly cleared paths.
     #[expect(unused_variables)]
     fn apply_conversation_config(
         &self,
@@ -300,8 +301,11 @@ pub(crate) trait IntoPartialAppConfig {
         partial: PartialAppConfig,
         merged_config: Option<&PartialAppConfig>,
         handle: &jp_workspace::ConversationHandle,
-    ) -> std::result::Result<PartialAppConfig, Box<dyn std::error::Error + Send + Sync>> {
-        Ok(partial)
+    ) -> std::result::Result<
+        (PartialAppConfig, Vec<String>),
+        Box<dyn std::error::Error + Send + Sync>,
+    > {
+        Ok((partial, Vec::new()))
     }
 }
 
--- a/crates/jp_conversation/src/stream.rs
+++ b/crates/jp_conversation/src/stream.rs
@@ -1,6 +1,9 @@
 //! See [`ConversationStream`].
 
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::{BTreeSet, HashMap},
+    sync::Arc,
+};
 
 use chrono::{DateTime, Utc};
 use jp_config::{AppConfig, ConfigError, FillDefaults as _, PartialAppConfig, PartialConfig as _};
@@ -549,6 +552,37 @@ impl ConversationStream {
         }
 
         Ok(partial)
+    }
+
+    /// Dotted config paths explicitly cleared and not subsequently set.
+    ///
+    /// Reset deltas discard earlier clears.
+    /// A clear followed by a replacement in the same apply delta is not
+    /// included.
+    /// Unknown paths are ignored.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if a delta cannot be folded onto the accumulated state.
+    pub fn config_unsets(&self) -> Result<Vec<String>, StreamError> {
+        let mut unsets = BTreeSet::new();
+        for delta in self.config_deltas() {
+            match delta {
+                ConfigDelta::Apply(apply) => unsets.extend(apply.unsets.iter().cloned()),
+                ConfigDelta::Reset(_) => unsets.clear(),
+            }
+        }
+
+        let partial = self.config_partial()?;
+        Ok(unsets
+            .into_iter()
+            .filter(|path| {
+                // Probe the typed path before resolution injects defaults. Clearing
+                // it changes nothing only when the accumulated field is still absent.
+                let mut cleared = partial.clone();
+                cleared.unset(path).is_ok() && cleared == partial
+            })
+            .collect())
     }
 
     /// Removes all events from the end of the stream, until a [`ChatRequest`]
--- a/crates/jp_cli/src/cmd.rs
+++ b/crates/jp_cli/src/cmd.rs
@@ -192,7 +192,7 @@ impl IntoPartialAppConfig for Commands {
         partial: PartialAppConfig,
         merged_config: Option<&PartialAppConfig>,
         handle: &jp_workspace::ConversationHandle,
-    ) -> Result<PartialAppConfig, Box<dyn std::error::Error + Send + Sync>> {
+    ) -> Result<(PartialAppConfig, Vec<String>), Box<dyn std::error::Error + Send + Sync>> {
         match self {
             Commands::Query(args) => {
                 args.apply_conversation_config(workspace, partial, merged_config, handle)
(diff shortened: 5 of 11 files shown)


crates/jp_cli/src/cmd/query.rs
```diff
@@ -2431,10 +2431,15 @@ impl IntoPartialAppConfig for Query {
...
-        let config = workspace.events(handle)?.config().map(|c| c.to_partial())?;
-
-        load_partial(partial, config).map_err(Into::into)
+    ) -> std::result::Result<
+        (PartialAppConfig, Vec<String>),
+        Box<dyn std::error::Error + Send + Sync>,
+    > {
+        let events = workspace.events(handle)?;
+        let config = events.config()?.to_partial();
+        let unsets = events.config_unsets()?;
```
JeanMertz: nit: `config()` and `config_unsets()` each fold the whole delta chain — `config()` via `config_partial()` + `build`, `config_unsets()` via its own `config_partial()` call — so every `jp query` on a conversation with a long delta history pays the fold twice. On top of that, `config_unsets` clones the full `PartialAppConfig` once per candidate path for its equality probe.

More than the cost, the two calls have to agree: `config_unsets` decides a path is "still cleared" by probing `config_partial()`, but the partial the pipeline actually unsets is `config()?.to_partial()`. Nothing here states that they must resolve the same way at those paths, and the invariant that makes it true (only a field whose *resolved* type is optional can reach `unsets`) lives over in `delta_law_tests.rs`.

Consider one stream API returning both, e.g. `fn config_layer(&self) -> Result<(PartialAppConfig, Vec<String>)>`, folding once and deriving the unsets from the same partial it returns. That removes the duplicate fold and makes the coupling structural rather than incidental.

crates/jp_cli/src/ctx.rs
```diff
@@ -293,15 +293,19 @@ pub(crate) trait IntoPartialAppConfig {
...
     fn apply_conversation_config(
         &self,
         workspace: &Workspace,
         partial: PartialAppConfig,
         merged_config: Option<&PartialAppConfig>,
         handle: &jp_workspace::ConversationHandle,
-    ) -> std::result::Result<PartialAppConfig, Box<dyn std::error::Error + Send + Sync>> {
-        Ok(partial)
+    ) -> std::result::Result<
+        (PartialAppConfig, Vec<String>),
```
JeanMertz: nit: The bare tuple makes the callers read `(conversation_config, unsets)` and `(conversation_partial, unsets)` with nothing at the trait naming what the second element is — `Vec<String>` of what, in which order, with what meaning if empty? The doc line above helps, but it's one line away from three destructuring sites in three files.

A small named type would carry that at the type level and leave room for the next thing the conversation layer needs to hand across:

```rust
/// The per-conversation config layer: its accumulated partial, plus the
/// dotted paths it explicitly cleared.
pub(crate) struct ConversationConfigLayer {
    pub partial: PartialAppConfig,
    pub unsets: Vec<String>,
}
```

The default impl then reads `Ok(ConversationConfigLayer { partial, unsets: Vec::new() })`, and `partial_with_conversation` can take it whole instead of two positional arguments that must not be swapped.

crates/jp_cli/src/config_pipeline.rs
```diff
@@ -314,12 +314,20 @@ impl ConfigPipeline {
...
         conversation: PartialAppConfig,
+        unsets: &[String],
     ) -> Result<PartialAppConfig> {
-        let partial = conversation.fill_from(self.base.clone());
+        let mut partial = conversation.fill_from(self.base.clone());
+        for path in unsets {
+            if let Err(error) = partial.unset(path) {
+                warn!(%path, %error, "Ignoring a config delta unset for an unknown field.");
+            }
+        }
```
JeanMertz: nit: This loop silently drops a value that the user's `config.toml` declares, and nothing is logged on the success path. Named input: a user clears `assistant.name` in a conversation on Monday, sets `assistant.name` in `.jp/config.toml` on Friday, then wonders why that conversation ignores it. `RUST_LOG=debug` gives them nothing to go on — the only trace is the `unsets` array inside the stored event stream.

Given the project's "observability is a must" stance, a single `debug!` after the loop covering what was suppressed would close that gap:

```rust
if !unsets.is_empty() {
    debug!(?unsets, "Conversation clears suppressing base-layer values.");
}
```

Separately, the `warn!` arm is unreachable from the production caller: `config_unsets` already filters on `cleared.unset(path).is_ok()`, so every path reaching here names a real field. It's still right to keep it for other callers, but it's worth knowing it isn't the diagnostic anyone will actually see.

crates/jp_cli/src/config_pipeline.rs
```diff
@@ -314,12 +314,20 @@ impl ConfigPipeline {
     /// base only what it does not, which is what the snapshot already meant for
     /// every field that merges by replacement.
     ///
-    /// `--cfg` still merges on top, where combining is what the user asked for.
+    /// `unsets` carries the conversation's explicitly cleared paths.
+    /// These are cleared after filling so the base cannot restore them.
+    /// `--cfg` merges on top and can set them again.
```
JeanMertz: question: This changes what `--cfg <path>:=null` means for a conversation, from "cleared for this turn" to "cleared for the rest of the conversation's life, overriding anything the workspace file says at that path." That's the fix the ticket asks for, and it is the right semantics, but it is also a user-visible contract change that lives nowhere outside this doc comment.

`docs/configuration.md:45-47` hands the loading sequence off to RFD 079, and RFD 079's step 12 is just "Load the conversation's `base_config.json` and event-stream `ConfigDelta`s" — no mention that those deltas' cleared paths now veto the file layer. Is a doc update in scope for this PR, or is it queued separately?

Related, and the part I could not answer from the code: what is the documented escape hatch? From reading `config_unsets`, a clear is retired by recording a value at that path (`--cfg assistant.name=Bot`) or by a reset (`--cfg=WORKSPACE`). Neither is obvious from the user's side, and there is no surface I could find that lists a conversation's active clears — `jp config` doesn't apply the conversation layer at all (`Commands::Config(_)` returns an empty layer in `cmd.rs`). Worth a line in `docs/configuration.md` naming both.

crates/jp_conversation/src/stream.rs
```diff
@@ -551,6 +554,37 @@ impl ConversationStream {
         Ok(partial)
     }
 
+    /// Dotted config paths explicitly cleared and not subsequently set.
+    ///
+    /// Reset deltas discard earlier clears.
+    /// A clear followed by a replacement in the same apply delta is not
+    /// included.
+    /// Unknown paths are ignored.
```
JeanMertz: nit: The second line ("A clear followed by a replacement in the same apply delta is not included") describes the filter more narrowly than it behaves. The implementation probes the *fully accumulated* partial, so a clear is also dropped when a much later delta sets the path — which is what `config_unsets_drops_a_path_when_a_later_delta_sets_it` asserts. The same-delta case is just one instance of that rule.

The summary line already says it correctly ("explicitly cleared and not subsequently set"); the second line reads as an additional, tighter rule and contradicts it. Suggest dropping it, or restating as the general rule:

```
/// A path the accumulated state holds a value at is not included, whether
/// the value arrived in the clearing delta or a later one.
```

crates/jp_cli/src/cmd/query.rs
JeanMertz: Took the cost half, left the API alone.

`config_unsets` now returns early when the delta chain cleared nothing, which is every conversation that never typed `:=null` — so the second fold and the per-path clone are gone from the common path:

```rust
// Nothing cleared means nothing to probe, and the fold below is a full
// merge per delta in the stream.
if unsets.is_empty() {
    return Ok(Vec::new());
}
```

with `config_unsets_is_empty_without_a_clearing_delta` pinning it. I haven't measured either version; n here is the delta count, so this is "don't do useless work", not a performance fix.

On the coupling: I went looking for a case where the two disagree and couldn't construct one that matters. Both partials come from the same fold; the only difference is what `build` adds on top, and `build` injects defaults and resolves aliases — it can add a value at a path, never remove one. So a path the filter *keeps* is absent in `config_partial`, therefore in the resolved partial it is either absent or a default, and unsetting a default is harmless because resolution reinjects it downstream. A path the filter *drops* is present in `config_partial`, therefore present after `build` too. The divergence is one-directional and benign.

That leaves `config_layer()` fixing an invariant that already holds, at the cost of the stream's public API plus three call sites. Not taking it for a bug fix, but I agree it's the shape to reach for if a third thing ever needs to cross that boundary.

crates/jp_cli/src/config_pipeline.rs
JeanMertz: The doc gap is real and now closed; the framing I'd push back on slightly.

This isn't a contract change introduced here. The stream already recorded the clear as durable — `fold_config_delta` applies `unsets`, `ConversationStream::config` resolves the field to `None` — and the pipeline was silently undoing it one layer up. That's what the ticket calls the bug. This PR makes the pipeline honor a contract the storage layer already implemented.

What was genuinely missing is any user-facing mention of clearing at all. `docs/configuration.md` never mentioned `:=null`, before or after this change. Added under "Conversation Configuration": the clearing example, that the clear holds for the rest of the conversation, and both ways to retire it (a value at the same path, or `--cfg=WORKSPACE`). RFD 079 step 12 grew two lines saying the cleared paths are also cleared from the base partial, with an [RFD 070] reference — it's a `Category: Guide` and `configuration.md:45-47` hands the loading sequence off to it, so it needs to stay accurate.

One correction to the escape hatches while writing that: the example had to be `jp query --cfg ...:=null`, not `jp config set`. Only the query path records `unsets` (`turn_config_delta` → `delta_with_unsets`). `jp config set --id` goes through `override_to_record`, which returns a bare `PartialAppConfig` — a partial can't express a clear, so the null merges as "no opinion", no delta is recorded, and it still prints `Set configuration in conversation <id>`. Filed as T-0kwgqkx.

The third part — no surface listing a conversation's active clears — is a feature, not a doc line, and you're right that `jp config` can't answer it (`Commands::Config(_)` returns an empty conversation layer at `cmd.rs:200-207`, and `cmd/config/show.rs` never touches a conversation). Filed as T-0kwgjda, noting that RFD 060 would subsume it.

crates/jp_cli/src/config_pipeline.rs
JeanMertz: Taken. Your Monday/Friday case is the one that convinced me: the damage there is hidden, not just unpleasant — a value the user's own `config.toml` declares never applies, and the only record was the `unsets` array inside the stored event stream.

Added before the loop, so the suppressed set is logged first and any per-path `warn!` reads as a correction to it:

```rust
if !unsets.is_empty() {
    debug!(?unsets, "Conversation clears suppressing base-layer values.");
}
```

`debug` was already imported, so that's the whole change.

On the `warn!` arm: agreed, and I checked why. `unset` only errors on a path that names no field (`unset_tests.rs:130`), an absent map entry is a no-op (`unset_tests.rs:67`), and `config_unsets` already filters on `unset(path).is_ok()` against the same type — so from the production caller it can't fire. The one crack I could construct is a map entry that's absent in the conversation's own partial but filled from the base as a variant lacking the named field. Contrived enough that I'm not writing a test for it; keeping the arm for other callers.

crates/jp_cli/src/ctx.rs
JeanMertz: Leaving the tuple, because both facts holding the suggestion up don't check out.

Destructuring sites are two, in two files: `lib.rs:1053` and `query_tests.rs:281`, one of which is a test. The other three mentions are the trait declaration in `ctx.rs`, its default impl, and `cmd.rs`'s delegation — none of them destructure.

And the swap can't happen: `partial_with_conversation(PartialAppConfig, &[String])` takes two unrelated types, so transposing them is a compile error rather than a bug. That was the argument doing most of the work for the struct.

What's left is naming taste, paid for with a new `pub(crate)` type and churn across four files plus tests, against a doc line at `ctx.rs:296` that already says what the second element is. If the conversation layer ever needs to hand a third thing across, the struct becomes worth it and `ctx.rs` next to the trait is where it goes.

crates/jp_conversation/src/stream.rs
JeanMertz: Right that the comment is wrong, though I'd diagnose it differently — so I fixed both lines rather than dropping one.

The second line doesn't contradict the summary; it covers a case the summary misses. "Not *subsequently* set" says nothing about a value arriving in the clearing delta itself, which is why `config_unsets_excludes_a_same_delta_replacement` exists as its own test. The real problem is that both lines under-describe one rule: the filter probes the fully accumulated partial, so any path holding a value is dropped regardless of which delta supplied it.

Now:

```rust
/// Dotted config paths explicitly cleared and still absent from the
/// accumulated state.
///
/// Reset deltas discard earlier clears.
/// A path the accumulated state holds a value at is not included, whether
/// the value arrived in the clearing delta or a later one.
/// Unknown paths are ignored.
```

Your replacement line verbatim, plus the summary reworded so it isn't narrower than the rule underneath it.

PR #1162, opened 2026-09-11
fix(md): Preserve reference-link definitions

JeanMertz:
Terminal output retains `[label]: URL` definitions, so streamed reference-style links no longer lose their destinations. Multiline definitions and definitions inside lists or blockquotes remain visible with the surrounding terminal styling. Markdown serialization is unchanged.

Block-by-block parsing cannot resolve definitions that arrive later, and comrak consumes definitions without rendering them. Preserve their source as literal text and buffer potential multiline definitions until a block boundary, while ordinary bracket-led prose continues to stream.

### DIFF
--- a/crates/jp_md/src/references.rs
+++ b/crates/jp_md/src/references.rs
@@ -1,0 +1,211 @@
+//! Reference-definition preservation for terminal rendering.
+//!
+//! [`restore`] adds literal paragraphs for definitions consumed by comrak.
+
+use std::ops::Range;
+
+use comrak::{
+    Arena, Node, Options,
+    nodes::{Ast, NodeValue},
+    parse_document,
+};
+
+/// Add literal paragraphs for reference definitions consumed during parsing.
+///
+/// `root` must be an unmodified comrak AST parsed from `source` with `options`.
+/// Existing links keep their resolved destinations; definitions retain their
+/// source order and container nesting.
+pub fn restore<'a>(arena: &'a Arena<'a>, root: Node<'a>, source: &str, options: &Options<'_>) {
+    if !source.contains("]:") {
+        return;
+    }
+
+    // Source positions count CRLF and bare CR as single line endings.
+    let source = source.replace("\r\n", "\n").replace('\r', "\n");
+    let lines: Vec<_> = source.split_inclusive('\n').collect();
+    // Only inspect the parser's nodes, not the literal paragraphs we insert.
+    let nodes: Vec<_> = root.descendants().collect();
+    for node in nodes {
+        let data = node.data();
+        let span = data.sourcepos;
+        match data.value {
+            NodeValue::Document
+            | NodeValue::BlockQuote
+            | NodeValue::Item(_)
+            | NodeValue::TaskItem(_) => {
+                drop(data);
+                // Definition-only paragraphs leave gaps between child spans.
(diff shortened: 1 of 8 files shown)


crates/jp_md/src/references.rs
```diff
@@ -0,0 +1,211 @@
...
+                        && let NodeValue::List(list) = parent.data().value
+                    {
+                        // TaskItem has no padding field. Its checkbox position
+                        // identifies the content column even when numbering
+                        // crosses from 9 to 10.
+                        let opening = lines[data.sourcepos.start.line - 1];
+                        let marker = data.sourcepos.start.column - 1;
+                        let checkbox = task.symbol_sourcepos.start.column - 1;
+                        let start_column = opening[..marker].chars().fold(0, advance_column);
+                        let end_column = opening[marker..checkbox]
```
JeanMertz: blocking: A checklist whose checkbox starts on the line after its list marker can panic during terminal rendering. For example, pass `"-\n  [x] Done.\n\n  [docs]: /docs\n"` to `Formatter::format_terminal`.

Comrak gives the `TaskItem` a start on line 1, but `symbol_sourcepos` refers to the `x` on line 2 (column 4). Here `opening` is therefore `"-\n"`, while `checkbox` is 3, so `opening[marker..checkbox]` slices past the end of that line. This takes down the command rather than returning a formatting error.

Derive the item's indentation from its actual list marker without assuming the checkbox shares that line, and add this input as a regression test.

JeanMertz: Confirmed and fixed. Your input panics with `end byte index 3 is out of bounds for string of length 2`, exactly as described.

The root cause is where you placed it: `process_tasklist` builds `symbol_sourcepos` from the *Text* node's position, then assigns `NodeValue::TaskItem` to the grandparent `Item` — so the two carry different lines whenever the marker line has no content, and comrak never promises otherwise.

Two details worth recording. The definition does not have to sit inside the task item: `restore` only guards on `source.contains("]:")`, so any `]:` anywhere in the document reaches this arm. And `format_terminal` also runs over user-authored markdown through `reformat_quoted_message`, not just model output.

The fix keeps the checkbox derivation for the common case and falls back to the marker line when the checkbox sits elsewhere:

```rust
let start_line = data.sourcepos.start.line;
let opening = lines[start_line - 1];
let marker = data.sourcepos.start.column - 1;
let start_column = opening[..marker].chars().fold(0, advance_column);
let end_column = if task.symbol_sourcepos.start.line == start_line {
    let checkbox = task.symbol_sourcepos.start.column - 1;
    opening[marker..checkbox].chars().fold(start_column, advance_column)
} else {
    // An empty marker line indents content one column past the marker.
    opening[marker..]
        .trim_end_matches(['\r', '\n'])
        .chars()
        .fold(start_column, advance_column)
        + 1
};
```

The `else` branch reproduces comrak's `nl.padding = matched + 1` rule for a content-free marker line, which is the only shape that can push the checkbox onto a later line.

Added `task_items_whose_checkbox_starts_on_a_later_line_remain_visible` with your input, asserting `"- [x] Done.\n  [docs]: /docs\n"`. I watched it panic against the unfixed code before taking the fix, so it pins the crash rather than just passing. All 256 `jp_md` tests are green.

PR #1164, opened 2026-09-11
fix(llm, openai, openrouter, llamacpp): Decode tool call argument nulls

JeanMertz:
Optional tool arguments encoded as null by strict providers reach tools as omitted fields. This lets Bookworm apply its defaults rather than rejecting `kinds: null`. Explicit values, source-accepted nulls, and required arguments retain their meaning.

Pair schema conversion with request-local decoding plans, consumed when streamed arguments become a complete tool call. Cover OpenAI's streaming and non-streaming paths, OpenRouter, and llama.cpp without changing MCP transport or persisted event types.

Keep reference resolution and schema composition behavior unchanged. Decoding through references and general unions remains deferred to T-0h28vbn; shared dispatch validation remains tracked by T-063sw1z.


crates/jp_llm/src/provider/openai.rs
```diff
@@ -1743,16 +1771,27 @@ fn process_strict(schema: &mut Value, root: &Value) {
...
     }
     if let Some(Value::Array(variants)) = map.get_mut("anyOf") {
         for variant in variants.iter_mut() {
-            process_strict(variant, root);
+            let plan = process_strict(variant, root);
+            if !skip_decoding {
+                decoding.omit_null.extend(plan.omit_null);
+                decoding.properties.extend(plan.properties);
+                decoding.items = decoding.items.or(plan.items);
+            }
```
JeanMertz: nit: this `!skip_decoding` guard is dead. When `skip_decoding` is true, `decoding` is never returned — the function ends with `if skip_decoding { return ArgumentDecoding::default(); }` (line 1839), and the only other exit after this point is the `$ref` unravel branch, which also returns `default()`. So anything hoisted here would be discarded regardless.

Keeping it invites the next reader to hunt for a case where hoisting *into a discarded plan* matters, and there isn't one. Dropping the guard leaves the single `skip_decoding` exit as the only place that policy lives:

```rust
if let Some(Value::Array(variants)) = map.get_mut("anyOf") {
    for variant in variants.iter_mut() {
        let plan = process_strict(variant, root);
        decoding.omit_null.extend(plan.omit_null);
        decoding.properties.extend(plan.properties);
        decoding.items = decoding.items.or(plan.items);
    }
}
```

If you'd rather keep it as a guard against a future early `return decoding`, a one-line comment saying so would do.

crates/jp_llm/src/provider/openai.rs
```diff
@@ -1734,7 +1757,12 @@ fn process_strict(schema: &mut Value, root: &Value) {
             if let Some(Value::Object(props)) = map.get_mut("properties") {
                 for key in &newly_required {
                     if let Some(prop_schema) = props.get_mut(key) {
-                        make_schema_nullable(prop_schema);
+                        let can_decode = ["$ref", "allOf", "anyOf", "oneOf"]
+                            .iter()
+                            .all(|key| prop_schema.get(*key).is_none());
```
JeanMertz: nit: there are now two hard-coded lists of composition keywords in this function that have to stay in agreement, and they deliberately disagree on `anyOf`:

- line 1713: `["$ref", "allOf", "oneOf"]` plus `anyOf` **only when it isn't a nullable pair`
- here: `["$ref", "allOf", "anyOf", "oneOf"]`, unconditionally

The reason they differ is real but invisible: at this site an `anyOf` means the source schema already expressed nullability itself, so a null is the author's value and must not be recorded as an omission. At line 1713 a nullable-pair `anyOf` is the wrapper *we* just injected, so descending into it is safe.

Two failure modes from the duplication: someone adds a fourth keyword (`not`, `if`/`then`) to one list and not the other, or someone "fixes" the inconsistency by making the two lists identical. Both silently start deleting nulls the source schema accepts — and the resulting bug is invisible in the tool call, because the argument is just gone.

Suggest a named const plus a short comment on why this site is stricter:

```rust
/// Keywords whose branch selection this traversal cannot follow.
const COMPOSITION_KEYS: &[&str] = &["$ref", "allOf", "oneOf"];

// An `anyOf` here means the source schema declared its own nullability, so a
// null is a value the author accepts — never an omission placeholder.
let can_decode = COMPOSITION_KEYS
    .iter()
    .chain(["anyOf"].iter())
    .all(|key| prop_schema.get(*key).is_none());
```

crates/jp_llm/src/tool/decoding.rs
```diff
@@ -0,0 +1,91 @@
...
+            if arguments.get(name) == Some(&Value::Null) {
+                arguments.remove(name);
+            }
+        }
+        for (name, plan) in &self.properties {
+            if let Some(value) = arguments.get_mut(name) {
+                plan.apply_value(value);
+            }
+        }
+    }
```
JeanMertz: nit: this removes arguments with no trace of having done so.

Named input: a model calls a tool with `{"kinds": null}`, the plan strips it, and the tool runs with defaults. If the defaults are wrong for the user's intent, the user sees a tool result that doesn't match what they watched the model ask for, and nothing in the logs connects the two. The `$JP_LOG=trace` request dump (`trace_to_tmpfile`) shows the outgoing schema, and the persisted `ToolCallRequest` shows the post-decode arguments — neither shows the delta.

Contained and visible (the tool ran, the user can see the result), so this isn't blocking. But observability is a stated must for JP, and this is a silent mutation of data on its way to execution. A single line covers it:

```rust
pub(crate) fn apply(&self, arguments: &mut Map<String, Value>) {
    for name in &self.omit_null {
        if arguments.get(name) == Some(&Value::Null) {
            trace!(argument = name, "Omitting null placeholder from tool arguments.");
            arguments.remove(name);
        }
    }
    ...
}
```

Nested removals go through the same path via `apply_value`, so one call site covers the whole tree — though the name alone won't say *where* in the tree it was. Good enough for a trace line.

crates/jp_llm/src/event_builder.rs
```diff
@@ -217,15 +225,19 @@ impl EventBuilder {
...
                     serde_json::Map::new()
                 } else {
                     serde_json::from_str(&arguments_json).unwrap_or_else(|e| {
                         warn!("Failed to parse tool call arguments JSON: {e}");
                         serde_json::Map::new()
                     })
                 };
+                if let Some(decoding) = decoding {
+                    decoding.apply(&mut arguments);
+                }
```
JeanMertz: question: decoding lands *before* the `ToolCallRequest` is constructed, so the decoded arguments are what gets persisted — and therefore what goes back on the wire on the next turn. `openai.rs:2288` re-serializes them verbatim:

```rust
EventKind::ToolCallRequest(request) => vec![types::InputListItem::Item(
    types::InputItem::FunctionCall(types::FunctionCall {
        call_id: request.id,
        name: request.name,
        arguments: Value::Object(request.arguments).to_string(),
```

So on turn N+1 the assistant history contains `{"query":"parse_document"}` for a tool whose strict schema declares `kinds` in `required`. Did you confirm none of OpenAI / OpenRouter / llama.cpp validate historical `function_call.arguments` against the advertised tool schema? My understanding is that they don't — validation is on generation, not on replay — but I haven't verified it and the fixtures here only cover a single turn.

Not blocking on my read: if a provider did reject it, the failure is loud (a 400 on the next request) rather than silent. But it's the one consequence of decoding-before-persistence that the tests don't reach, and it'd be worth a line in the ticket comment either way.

The alternative shape — persist what the model emitted, decode only on the way into `executor` — costs you the Bookworm fix for anything that reads arguments off the stream rather than off the dispatch path, so I don't think it's obviously better. Mostly asking whether it was weighed.

crates/jp_llm/src/event_builder.rs
```diff
@@ -393,12 +408,14 @@ impl IndexBuffer {
...
                 id: incoming_id,
                 name: incoming_name,
+                decoding: incoming_decoding,
             } => {
                 if id.is_empty() && !incoming_id.is_empty() {
                     *id = incoming_id;
                 }
                 if name.is_empty() && !incoming_name.is_empty() {
                     *name = incoming_name;
+                    *decoding = incoming_decoding;
```
JeanMertz: theoretical: `decoding` is adopted only on the branch that fills an empty `name`. A `Start` that arrives with a name already set on the buffer keeps the original plan — correct, and consistent with the first-non-empty-wins rule documented above.

The gap is the other order: a buffer opened by `ArgumentChunk` gets `decoding: None` (line 181), and if the `Start` that follows carries an *empty* name but a non-empty id, `decoding` stays `None` for the rest of the call and the nulls survive.

I couldn't name a provider that does this. `attach` keys the lookup on the name, so an empty-name `Start` gets `None` anyway — the plan would have to arrive attached by something other than `ArgumentDecoders`, which nothing does today. Recording it because the coupling is implicit: `decoding` is correct here only because it is derived from `name`, and nothing in the type says so. A word on the field doc at line 376 (`Decoding chosen by the first non-empty tool name`) already half-says it; tying the two together in one line at line 416 would close it.

crates/jp_llm/src/event.rs
```diff
@@ -110,6 +114,9 @@ pub enum ToolCallPart {
 
         /// Name of the tool to execute.
         name: String,
+
+        /// Request-local decoding instructions, consumed before persistence.
+        decoding: Option<Arc<ArgumentDecoding>>,
```
JeanMertz: nit: two things about the public surface here.

**The invariant isn't enforceable by anything but convention.** "Consumed before persistence" is true today because `EventBuilder::handle_flush` is the only production path that turns a `ToolCallPart` into a `ToolCallRequest` — I grepped, and every other construction site is in test code. But the doc states the invariant without saying who upholds it, so the next person adding a stream adapter has no way to know they'd break it. Worth naming the enforcer:

```rust
/// Request-local instructions for restoring arguments the provider's
/// schema encoding turned into nulls.
///
/// Set by `ArgumentDecoders::attach` on the provider's stream and applied
/// by `EventBuilder::handle_flush`. It is never persisted: no code path
/// constructs a `ToolCallRequest` from a `ToolCallPart` outside the builder.
decoding: Option<Arc<ArgumentDecoding>>,
```

**`ArgumentDecoding` is `pub` with no public API.** Every field and every method on it is `pub(crate)`, so a downstream crate sees an opaque type it can only `Default`-construct — and, because this is a public enum variant, constructing `ToolCallPart::Start` from outside `jp_llm` now requires naming it. Nothing outside the crate does either today, so this costs nothing right now. If the type is meant to stay internal, `pub(crate) struct` plus a `#[doc(hidden)]`-style escape isn't available for a public variant field, so the honest options are to leave it as-is or to give it a minimal public constructor. Mentioning it so the choice is deliberate rather than incidental.

PR #1166, opened 2026-09-11
feat(conversation, rfd): Add stable event IDs

JeanMertz:
Every entry in `events.json` now has an `event_id` that remains stable across saves and content edits. Scripts can reference stored entries without relying on array positions, and tool-call payload IDs retain their separate meaning.

Legacy entries receive IDs on load and persist them on the next save. For duplicate IDs, the first occurrence is preserved and later ones get new IDs. JP warns for each replacement and retains duplicated values as load-scoped ambiguity for future reference consumers.

Projection preserves the IDs of retained entries. Synthetic summary entries receive ephemeral IDs with no corresponding entry in `events.json`. The CLI help and architecture docs explain this boundary and the manual-edit repair rules.

Implements: RFD 097


crates/jp_cli/src/cmd/conversation/edit_tests.rs
```diff
@@ -2,8 +2,36 @@ use std::{fs, time::Duration};
 
 use camino::Utf8PathBuf;
 use camino_tempfile::tempdir;
+use clap::{Parser as _, error::ErrorKind};
 
 use super::{Edit, ExpirationDuration};
+use crate::Cli;
+
+#[test]
+fn events_help_explains_stable_ids_and_duplicate_repair() {
```
JeanMertz: Drop this

crates/jp_cli/src/cmd/conversation/fork_tests.rs
```diff
@@ -23,6 +23,18 @@ use crate::{
     cmd::{compact_flag::CompactFlag, conversation_id::PositionalIds},
 };
 
+/// Compare stream content without requiring cross-conversation ID equality.
+fn assert_same_stream_content(left: &ConversationStream, right: &ConversationStream) {
```
JeanMertz: This seems brittle. Should we manually implement `PartialEq` on `ConversationStream`? Or else perhaps an explicit method such as ConversationStream::equal_except_event_ids(&self, &Self)`?

crates/jp_conversation/src/stream.rs
```diff
@@ -1407,8 +1466,7 @@ impl ConversationStream {
         IterTurns::new(self.iter())
     }
 
-    /// Returns each event paired with the 0-based index of the turn it belongs
-    /// to.
+    /// Iterate over `(turn_index, event_id, event)` with 0-based turn indices.
```
JeanMertz: Should we still advertise `turn_index` at all? Is it still needed?

docs/rfd/097-stable-event-identifiers.md
```diff
@@ -221,36 +221,67 @@ The event constructors are unchanged: `ConversationEvent::new(kind, ts)`,
...
-wrapper directly with a fixed `EventId`; live and synthetic insertion paths let
-the stream assign one.
-There is no thread-local RNG and no test/prod plumbing inside the event
-constructors.
+
+An ID is retired with the entry that held it: the stream does not hand out an ID
+again after the entry holding it is removed.
+Otherwise a reference to a deleted entry could silently rebind to a later,
+unrelated one — the positional aliasing this RFD exists to remove, reintroduced
+through the ID.
```
JeanMertz: theoretical: Please scope the non-reuse guarantee to the lifetime of the in-memory stream. After deleting an entry, saving, and loading the conversation in another command, `from_parts` rebuilds `event_ids` from the surviving entries only (`stream.rs:1471-1475`); `to_parts` does not persist retired IDs. A subsequent random draw can therefore reuse a deleted entry's ID, contrary to this unconditional retirement claim.

The lifecycle is ordinary, but triggering the rebind also requires a random collision with a deleted ID; I cannot name a plausible input that forces that collision. This does not need to block merging or require persistent tombstones. Qualify this paragraph and the glossary's “never reissues” wording so future reference consumers do not assume a cross-load guarantee.

crates/jp_conversation/src/stream/entry.rs
```diff
@@ -0,0 +1,317 @@
...
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let mut value = Value::deserialize(deserializer)?;
+        let object = value
+            .as_object_mut()
+            .ok_or_else(|| D::Error::custom("stream entry must be a JSON object"))?;
+        let event_id = match object.shift_remove("event_id") {
+            None => None,
+            Some(Value::String(id)) if id.is_empty() => None,
+            Some(id) => Some(from_value(id).map_err(D::Error::custom)?),
+        };
```
JeanMertz: nit: `""` reads as absent but `null` is fatal, and the `--events` help doesn't draw that line.

Named input: a user follows `jp conversation edit --events` and clears an entry's ID. The help text added in this PR says "a missing, empty, or duplicated ID is replaced on the next load", and `null` is the natural JSON spelling of "no value" — but `Some(Value::Null)` falls into the last arm, `EventId::deserialize` rejects a non-string, and the whole conversation fails to load.

The blast radius is small: `Edit::validate_edits` reloads the file and `Edit` reverts from its pre-edit snapshot, so the user gets an error and an untouched conversation rather than a wedged one. That's why this is a nit and not blocking.

Still, the leniency the comment above describes ("leniency the storage boundary owes a legacy file") applies just as well to `null` as to `""`, and the fix is one arm:

```rust
None | Some(Value::Null) => None,
Some(Value::String(id)) if id.is_empty() => None,
```

A non-string, non-null value (a number, an object) staying fatal is the right call and stays intact.

crates/jp_conversation/src/stream.rs
```diff
@@ -807,11 +488,17 @@ impl ConversationStream {
...
+    /// entry.
     /// `other`'s config deltas are appended verbatim rather than recomputed, so
     /// the two streams must share a base config for the result to resolve the
     /// same way.
     pub fn append_stream(&mut self, other: Self) {
-        self.events.extend(other.events);
+        for entry in other.events {
+            self.adopt(entry);
+        }
     }
```
JeanMertz: theoretical: `append_stream` carries the IDs across but drops `other.duplicated_event_ids`, so the ambiguity record does not survive a fork.

Concretely: load a conversation whose `events.json` was hand-edited to repeat an ID. `from_parts` keeps the first entry's ID, reassigns the later one, and records the shared value. `jp conversation fork` then builds a fresh stream and calls `append_stream`, which adopts every entry — including the one that kept the ambiguous ID — into a stream whose `duplicated_event_ids` is empty. A consumer asking the fork whether that ID is ambiguous is told it is not.

`Extend<ConversationEventWithConfig>` has the same shape, and `adopt` is the common funnel.

I'm filing this as theoretical rather than blocking because nothing resolves `EventId` references yet — `duplicated_event_ids` has no production reader (grep finds only tests), so no input produces a wrong resolution today. But the RFD's contract is that a consumer "must do so within this load cycle", and a fork is inside that cycle: this is the first place the contract can be broken, and it will be broken silently.

Two directions, either fine:
- Union `other.duplicated_event_ids` into `self` in `append_stream`, so ambiguity travels with the entries that carry it.
- Or state in the `duplicated_event_ids()` doc that it reports the ambiguity *this stream's own load* found, and that entries adopted from elsewhere carry none — so a future consumer knows to ask the source stream, not the fork.

The second is cheaper and probably the honest one, given `Extend` recomputes rather than copies config deltas anyway.

crates/jp_conversation/src/event_id_tests.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+#[test]
+fn fresh_never_repeats_itself() {
+    let mut ids = EventIds::default();
+    let drawn: Vec<_> = (0..1_000).map(|_| ids.fresh()).collect();
+
+    let unique: std::collections::HashSet<_> = drawn.iter().collect();
+    assert_eq!(unique.len(), drawn.len());
+}
```
JeanMertz: nit: this test can't fail for the reason its name gives, and can fail for a reason it doesn't.

`fresh` would pass this with its retry loop deleted: 1000 draws from a 36^7 space collide with probability ~6e-6, so `EventId::random()` alone satisfies the assertion. What actually pins the retry is `a_draw_retries_past_an_id_the_set_handed_out_earlier` a few lines up, which scripts the generator and is deterministic.

The flip side of that same arithmetic is that this one flakes about once in 160,000 runs — rare enough to never be caught, common enough to eventually waste someone's afternoon.

You annotate exactly this distinction elsewhere in the PR (`a_summarized_entrys_id_does_not_reappear_in_the_projected_view` says outright that it's "the weaker half of the pair" and names the deterministic half). This test is the one place the annotation is missing. Either drop it, or borrow that comment and point at the scripted test.

Same applies to `two_draws_produce_distinct_ids` at line 136, though there the cost is one draw rather than a thousand.

crates/jp_cli/src/cmd/conversation/edit_tests.rs
JeanMertz: Dropped, along with the `clap` and `Cli` imports it needed.

Agreed on the merits: it pinned a doc comment rather than a behaviour. The only way it fails is if someone edits the `///` on `Edit::events`, and that edit is visible in the same diff that breaks the test.

crates/jp_cli/src/cmd/conversation/fork_tests.rs
JeanMertz: Keeping the helper, for two reasons.

`PartialEq` on `ConversationStream` is already hand-rolled (`stream.rs:93-105`). It excludes `event_ids` and `duplicated_event_ids` but keeps `events`, and `InternalEvent` derives `PartialEq` over `event_id`. So entry IDs are part of stream equality by design, and the helper exists *because* of that: the two streams genuinely differ in one entry's ID, since each side synthesized its own leading `TurnStart`.

`equal_except_event_ids` would be a production method with one test caller, and it would throw out the assertion this PR exists to make — that a fork preserves IDs. The comparison that ignores IDs is the one this test must not do.

One real thing did fall out of re-reading it: `assert_eq!(source.created_at, fork.created_at)` at line 60 is dead. Both call sites set `stream.created_at = DateTime::<Utc>::UNIX_EPOCH` on both streams just before calling. Happy to delete that line if you want it gone; left it alone for now since it's outside what the rest of this round touches.

crates/jp_conversation/src/stream.rs
JeanMertz: Still needed, and nothing is called `turn_index` any more — the field is `EventInTurn::turn` (`stream/iter.rs:298`). Grep finds `turn_index` only as two test function names in `turn_iter_tests.rs` and an unrelated comment in `anthropic.rs`.

Three production sites read it:

- `grep.rs:434` — `Some(event.turn + 1)`, the 1-based header matching the `--turn` selector.
- `jp_ffi/display.rs:96` — `let index = in_turn.turn;`
- `stream.rs:1266` — `turn_count()` is `iter_events_by_turn().last().map_or(0, |e| e.turn + 1)`.

Dropping it would push all three onto `iter_turns`, which the comments at the first two call sites explicitly avoid because it resolves and clones a `PartialAppConfig` per event.

crates/jp_conversation/src/event_id_tests.rs
JeanMertz: The arithmetic checks out, and so does the flake rate. Took the second option: renamed to `a_thousand_draws_produce_distinct_ids` and borrowed the annotation.

Kept it rather than dropping it because it covers one thing nothing else does: `a_generated_id_has_the_documented_length` and `a_generated_id_uses_lowercase_base36` both pass for a generator that returns `"aaaaaaa"` forever. What was wrong was the name's promise, not the test.

```rust
#[test]
fn a_thousand_draws_produce_distinct_ids() {
    // The weak half of the pair: a thousand draws from a 36^7 space are almost
    // certainly distinct whether or not `fresh` checks the set, so what this
    // pins is that `EventId::random` varies at all — a generator returning one
    // value forever satisfies every other test in this file.
    // `a_draw_retries_past_an_id_the_set_handed_out_earlier` pins the retry
    // itself, deterministically.
```

Left `two_draws_produce_distinct_ids` alone: one pair collides at 1.3e-11, so it will not flake in the life of this project, and it is the cheapest statement that the generator varies.

While in there: the cross-reference at `event_id_repair_tests.rs:71` named `fresh_skips_an_id_the_set_already_holds`, which does not exist. Pointed it at `a_draw_retries_past_an_id_the_set_already_holds`.

crates/jp_conversation/src/stream/entry.rs
JeanMertz: Taken. `null` now reads as absent:

```rust
let event_id = match object.shift_remove("event_id") {
    None | Some(Value::Null) => None,
    Some(Value::String(id)) if id.is_empty() => None,
    Some(id) => Some(from_value(id).map_err(D::Error::custom)?),
};
```

The deciding argument is the one your comment makes: the comment two lines above already commits to leniency the storage boundary owes a hand-edited file, and `null` is JSON's own spelling of "no value" — unlike `7` or `{}`, which are type errors. Treating the canonical null as corruption while treating `""` as absence was the inconsistency.

Two consequences, since the old behaviour was pinned deliberately:

- `an_entry_with_a_non_string_id_is_rejected` loses `Value::Null` from its list; its comment now names `null` as the exception. Everything else there stays fatal, as you said it should.
- Added `an_entry_with_a_null_id_reads_as_having_none` alongside the missing and empty cases.

Left the `--events` help text as-is: "a missing, empty, or duplicated ID is replaced" now reads correctly for all three accepted spellings, and naming `null` explicitly would be noise for readers who never write it.

crates/jp_conversation/src/stream.rs
JeanMertz: Confirmed on every point, and took the second direction.

`append_stream` (`stream.rs:498`) and `Extend<ConversationEventWithConfig>` (`iter.rs:47`) both funnel through `adopt`, which touches `event_ids` only, and `ConversationStream::new` seeds `duplicated_event_ids` empty. `fork.rs:280` is the single production caller. The "no production reader" half holds too — grep finds the field, the accessor, `event_id_repair_tests.rs`, and two doc files.

Union is the wrong shape for now: it would change the field's meaning from "what this load found" to "what this stream has inherited", and `Extend` does not have copy semantics anyway. Defining inheritance before a consumer exists to say what it needs is a speculative axis.

`duplicated_event_ids()` now says so:

> Scoped to this stream's own load. Entries taken from another stream, through `append_stream` or `Extend`, arrive without that stream's record, so a consumer resolving a reference against a copy asks the stream that read the file.

RFD 097 gains the same paragraph next to the load-cycle contract, including why the union is deliberately not done — so whoever writes the first reference-resolving feature finds the reasoning rather than re-deriving it.

docs/rfd/097-stable-event-identifiers.md
JeanMertz: Correct, and scoped. `from_parts` seeds `EventIds` from `stored.iter().filter_map(|e| e.event_id.clone())` — the IDs the file still carries — and `to_parts` writes no tombstones, so retirement cannot outlive the in-memory stream.

Qualified in four places, since the unconditional claim appeared in all of them:

- RFD 097, the retirement paragraph: retirement lasts the lifetime of that in-memory stream and no longer, with the reason nothing is done about it (the draw has to hit one value out of 36^7, and tombstones would grow `events.json` without bound to guard a collision no input can force), plus a line telling a future reference consumer not to read cross-load non-reuse into it.
- `ubiquitous-language/conversation.md`: "never reissues one" → a removed entry takes its ID with it for as long as that stream is in memory, and the guarantee is per-load.
- The `event_ids` field doc on `ConversationStream`.
- The `EventIds` type doc, which made the same claim about the set.

Agreed on not blocking and not adding tombstones — this was documentation overclaiming, not behaviour.

PR #1170, opened 2026-09-12
feat(mcp)!: Add an in-process tool server

JeanMertz:
JP executes local, built-in, and configured MCP tools through one
in-process MCP server. Ordinary JP queries and third-party MCP clients
use the same loopback Streamable HTTP handlers. Approvals, argument and
result editing, inquiry routing, and conversation recording remain with
the MCP Host through private channels. Tools requesting input run again
with accumulated answers; final delivery waits for recording.

Existing conversations retain their text/error storage format. MCP
resource attachments remain readable and removable, but cannot be
resolved for queries. Typed tool results preserve ordered content,
resources, annotations, structured data, and error details through Host
review. Unchanged reviews preserve that data; text edits replace the
delivered content.

Configured upstream MCP servers remain stdio-only. The HTTP endpoint
validates Host and Origin headers, supports scoped cancellation and
response resumption, and has no authentication. This implements
RFD 109 and provides the tool service required by RFD 110.

BREAKING CHANGE: MCP resource attachments cannot be resolved.

Conversations containing `mcp+<server>+<scheme>://` attachments still load,
list, and print. Before querying them, remove those attachments with
`jp attachment rm '<uri>'` and supply their content through another
supported attachment source.


crates/contrib/async-anthropic/src/bearer.rs
```diff
@@ -0,0 +1,143 @@
+//! Bearer-token (OAuth) authentication mode.
```
JeanMertz: Do we need this code for the ACP flow or only for the direct/old oauth flow?

I wonder how much code we could remove by dropping support for that old flow, given how well the ACP flow works, and is officially supported.

crates/contrib/async-anthropic/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
```
JeanMertz: Why is this crate actually embedded in this repo in THIS particular PR, if at all needed, shouldn't it be pulled in in the actual PR that adds the Anthropic subscription flow, not the MCP server?

crates/jp_attachment_mcp_resources/src/lib.rs
```diff
@@ -1,11 +1,19 @@
-use std::{collections::BTreeSet, error::Error};
+//! The `mcp` attachment scheme, kept readable but no longer resolvable.
```
JeanMertz: I wonder, could we not keep this handler working without relying on jp_mcp by letting it manage its own mcp connections somehow?

crates/jp_mcp/src/server/service.rs
```diff
@@ -0,0 +1,1049 @@
...
+            // race-free: a call finishing in between is still observed.
+            let idle = self.inner.idle.notified();
+            tokio::pin!(idle);
+            idle.as_mut().enable();
+            if self.inner.state().active.is_empty() {
+                break;
+            }
+            idle.await;
+        }
+        self.inner.upstream.shutdown().await;
```
JeanMertz: blocking: With two concurrent plugin queries, finishing one Turn shuts down the upstream MCP servers used by the other. `plugin::dispatch::prepare_turn` deliberately uses `McpServerScope::Shared`; `TurnInputs::collect` clones `ctx.mcp_client`, and `TerminalExecutorSource::start_with_metadata` passes another clone into `Service::new`. Those clones share the same services map, which `Client::shutdown` drains. A surviving Turn between tool calls therefore gets `UnknownServer` on its next upstream invocation. Shutdown can also wait for an unrelated Turn's in-flight call because `Client::call_tool` holds the map's read lock across its await.

Keep shared upstream lifetime with the plugin Host, or give each execution service genuinely independent upstream ownership. Add a concurrent-turn regression where one Turn finishes and the other successfully invokes its MCP tool afterward.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
```diff
@@ -1516,273 +1564,298 @@ impl ToolCoordinator {
...
+                    %error,
+                    tool = %tool.tool_name,
+                    "Tool call could not be completed."
+                );
+                self.set_tool_state(&tool.tool_id, ToolCallState::Completed);
+                *tracked_review = Some(Review::replaced(ToolCallResponse {
+                    id: tool.tool_id.clone(),
+                    result: Err(format!(
+                        "Tool '{}' was not executed: JP could not complete the call. You may \
+                         retry it.",
```
JeanMertz: blocking: An upstream tool can create a GitHub issue and then lose its connection before returning the result. That becomes `ToolError::McpRunToolError`, propagates out of `execute_with_answers`, and reaches this branch as `ExecutorError::Transport`. Execution has already happened, but the persisted response says it did not and explicitly invites the model to retry, risking a duplicate side effect.

Distinguish failures known to precede execution from failures with an uncertain outcome. For the latter, record that execution may have occurred and require checking the external state before retrying. A regression fixture should perform a visible side effect and disconnect before returning its response.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
```diff
@@ -1919,80 +1959,76 @@ impl ToolCoordinator {
...
+    fn handle_edit_result(prompter: &ToolPrompter, response: ToolCallResponse) -> Review {
+        let original = response.result.as_deref().unwrap_or_default();
+        match prompter.edit_result(original) {
+            Ok(Some(edited)) => Review::replaced(ToolCallResponse {
                 id: response.id,
                 result: Ok(edited),
-            },
-            Ok(None) => response,
-            Err(_) => ToolCallResponse {
+            }),
```
JeanMertz: blocking: With `result = "edit"`, pressing Enter on an unchanged non-empty result reaches `Ok(Some(edited))`: `ToolPrompter::edit_result` returns `Some` for every non-empty submission, not just changed text. This marks an unchanged review as replaced, so `approved()` rebuilds a text-only `ToolResult` and silently drops the upstream resource/image blocks, structured content, and metadata before delivery to the agent.

Compare the submitted text with the offered text and return `Review::unchanged(response)` when they match. Test the actual editor-submit path with a rich MCP result; the existing tests construct `Review::unchanged` directly and miss this distinction.

crates/jp_mcp/src/server/service.rs
```diff
@@ -0,0 +1,1049 @@
...
+        if let Some(value) = arguments.get(&name) {
+            validate_value(&name, value, &node)?;
+        }
+    }
+    Ok(())
+}
+
+fn validate_value(path: &str, value: &Value, node: &Node<'_>) -> Result<(), ServiceError> {
+    if !node.permits(value) {
+        return Err(ServiceError::InvalidArgument { path: path.into() });
```
JeanMertz: nit: An OpenAI tool call that omits an optional integer by sending `null` is rejected here. `openai::ensure_strict_schema` explicitly makes optional properties required-but-nullable so the model can use `null` for omission, while this check uses the original non-nullable schema. For example, a tool with optional `count: integer` accepts omission in its advertised OpenAI schema but `{"count":null}` fails before preparation. The failure is contained and reported, but it breaks ordinary calls conforming to the provider's schema.

Normalize provider-added null-as-omitted values back to omission at the provider boundary, preserving genuinely nullable fields, before applying source-schema validation. Add a test spanning strict-schema adaptation and server validation, including a field with a configured default.

crates/jp_llm/src/provider/anthropic.rs
```diff
@@ -380,6 +330,34 @@ fn is_switchable(error: &AnthropicError) -> bool {
...
+    ) -> Result<QueryStream> {
+        let attempt = self.resolve(model.name()).await?;
+        if matches!(attempt.route, resolve::Route::Acp) {
+            acp::inspect().await?;
+            return Ok(QueryStream {
+                events: acp::stream(model, query, context)?,
+                execution: ToolExecution::Agent {
+                    correlation_key: "claudecode/toolUseId",
+                },
+            });
```
JeanMertz: question: Is credential-chain fallback intentionally unavailable for ACP in this delivery? With `auth = ["subscription", "api_key"]` and an exhausted subscription, this returns the ACP stream directly. The ACP protocol maps the runtime's `rate_limit` error to `StreamError::Other`, and the agent branch in `run_turn_loop` returns the error without advancing the chain. The new `a_spent_acp_subscription_falls_through_to_the_api_key` test exercises `resolve::advance` directly, so it does not cover this production path.

RFD 110 says an explicitly listed API entry retains the paid fallback policy. Either wire a safe pre-content exhaustion transition through the provider/Host dispatch contract and test it end to end, or document this as an explicit v0.1 limitation and adjust the test's claim.

crates/jp_mcp/src/server/service.rs
JeanMertz: Fixed. `Service::shutdown` no longer closes upstream connections: they belong to the Host (`Ctx`), which already decides their lifetime through `McpServerScope`. I also removed `Client::shutdown`. Nothing else called it, and keeping it around makes it easy to repeat this mistake. I updated the `Service` doc, the `jp_mcp` README and the shutdown paragraph in RFD 109 to match.

Regression test: `shutting_down_one_service_leaves_shared_upstream_servers_running` builds two services over one shared `Client`, shuts down the first, then runs an upstream tool call through the second.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
JeanMertz: Fixed. `handle_edit_result` now compares the submitted text with the offered text and returns `Review::unchanged` when they match, so the service still delivers the tool's full result (resources, images, structured content, error details).

Tests: `submitting_the_offered_result_unchanged_is_not_an_edit` and `submitting_changed_text_replaces_the_result`. Both drive the real `ToolPrompter::edit_result` path through a mock inline-reply backend. The first fails with the guard removed.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
JeanMertz: Fixed. By the time the error reaches the Host it's flattened into `ExecutorError::Transport`, so the classification is done where the answer is known instead: in `ToolExecutor::execute`, based on whether the release (or inquiry answer) had already been delivered to the service.

- Failed before release: `ExecutorResult::Failed`. The response still says the tool was not executed and may be retried.
- Failed after release: the new `ExecutorResult::OutcomeUnknown`. The recorded response is "Tool '…' may have run, but JP lost the call before its result arrived. Check whether its effects took place before calling it again."

Regression test: `a_call_lost_after_release_is_not_reported_as_unexecuted`. The tool increments a counter (the side effect) and then panics, so the service loses the call after release. The test asserts `OutcomeUnknown` and exactly one attempt. With the release flag ignored, it fails as `Failed(Transport(...))`.

crates/jp_mcp/src/server/service.rs
JeanMertz: Agreed this is a real bug, and I'm treating it as more than a nit. Strict mode makes every optional property nullable, so any tool with an optional non-nullable parameter failed on almost every OpenAI call. Before this PR, only missing and unknown keys were checked, so the `null` got through.

I put the fix in `jp_tool::ToolDefinition::coerce_arguments` rather than at the provider boundary. Coercion is already where provider-neutral repairs of model output happen, and it runs before defaults in both `Service::validate_arguments` and `execute`. Doing it in the provider would mean passing the source schema into every response parser. The rule: a `null` for a property that is optional and whose schema doesn't accept `null` is dropped, so the configured default applies. Nullable fields keep their `null`, and a required field's `null` still fails validation.

Tests in `definition_tests.rs` cover each case: optional with default (coerce, then defaults, then validation passes), nested optional, nullable kept, required kept. One trade-off: a third-party MCP client sending `null` for such a field is also treated as omitting it, the same lenient-input stance coercion already takes.

crates/jp_llm/src/provider/anthropic.rs
JeanMertz: Correct: not in this PR. #1192 (`93a0abe`, "Add ACP account fallback") wires it through. `Anthropic::with_acp_fallback` advances the chain on `SubscriptionExhausted`. The ACP protocol classifies SDK `rate_limit_event` rejections, and only five-hour/seven-day scopes or `credits_required` count as exhaustion; a bare `rate_limit` doesn't switch accounts. `run_turn_loop` restarts the agent turn via `can_restart_agent` instead of returning the error. The resolver test here stays at the `resolve::advance` level; the end-to-end coverage lives in #1192's `fallback_tests.rs` and `agent_turn_tests.rs`.

crates/contrib/async-anthropic/Cargo.toml
JeanMertz: This comment is outdated after the rebase: the crate came in with #1151 (`611fa1c9`), which is already on `main`. The only change to it left in this PR is `src/types.rs`, which adds the cache-usage fields (`cache_creation_input_tokens`, `cache_read_input_tokens`, `cache_creation`) that ACP usage accounting deserializes.

crates/contrib/async-anthropic/src/bearer.rs
JeanMertz: Only the direct flow uses it. `bearer::merge_betas` is called from `anthropic/http.rs` and the OAuth constants from `anthropic/auth.rs`, and ACP never builds a bearer client. Dropping the direct flow would remove `bearer.rs`, `auth.rs`, `oauth.rs`, the `Route::Http(Bearer)` handling and the direct-subscription tests. RFD 110 lists removing it as a Non-Goal and keeps `subscription_flow = "direct"` as the documented opt-out, though, so I'd do that in its own PR with an RFD 110 amendment once ACP has run as the default for a while.

crates/jp_attachment_mcp_resources/src/lib.rs
JeanMertz: Possible, but not as a small change, so I'd leave it out of this PR. `Handler::get(&self, cwd)` can't see the `providers.mcp` config, so the handler would need config passed in, and it would start stdio servers on every resolution (startup, checksum checks, timeouts). RFD 109 already scopes this out ("Plugin-based MCP attachments can be designed separately"). What this PR needed was to take `Client` out of the base `Handler` trait. A self-contained handler could come back later depending on `jp_mcp` with just the `client` feature.

docs/rfd/110-anthropic-subscription-queries-via-acp.md
```diff
@@ -0,0 +1,508 @@
...
+entry.
+
+Honor `assistant.request.cache` through the qualified runtime controls:
+
+| JP policy       | ACP flow mapping                                                                                         |
+| --------------- | -------------------------------------------------------------------------------------------------------- |
+| `off`           | `DISABLE_PROMPT_CACHING=1`.                                                                              |
+| `short`         | `CLAUDE_CODE_PROMPT_CACHE_TTL=5m`.                                                                       |
+| `long`          | `CLAUDE_CODE_PROMPT_CACHE_TTL=1h`.                                                                       |
+| Custom duration | Existing Anthropic mapping: at least 30 minutes selects one hour; shorter durations select five minutes. |
```
JeanMertz: nit: Align this cache-policy table and the following default-retention paragraph with the implemented contract. For `assistant.request.cache = "short"`, `"long"`, or a custom duration, `acp::options::environment` leaves retention to Claude Code; the launcher removes inherited `CLAUDE_CODE_PROMPT_CACHE_TTL`, and `cache_policy_reaches_the_native_sdk_environment` explicitly asserts that no TTL override is sent. The provider README and qualification guide describe that behavior correctly, but this accepted RFD promises explicit five-minute/one-hour selection. Document the runtime-managed mapping here too, so future changes are not judged against a conflicting requirement.

JeanMertz: Fixed. The table now has two rows: `off` → `DISABLE_PROMPT_CACHING=1`, and `short`/`long`/custom duration → no override, so Claude Code's own default and managed policy apply. I also rewrote the paragraph after it, which contradicted the code in two places: it said "JP's default remains `short`" and promised a managed-policy conflict report that doesn't exist. It now says the ACP flow enforces no retention duration, that `off` is the only thing JP sends, that the launcher strips the inherited cache overrides (`DISABLE_PROMPT_CACHING*`, `CLAUDE_CODE_PROMPT_CACHE_TTL`, `FORCE_PROMPT_CACHING_5M`, `ENABLE_PROMPT_CACHING_1H`, matching `acp::removes_variable`), and that the direct HTTP flows keep their duration mapping. This matches the provider README, the qualification guide, and `cache_policy_reaches_the_native_sdk_environment`.

PR #1171, opened 2026-09-13
fix(llm, openai): Stop a redeemed usage reset locking the account out

JeanMertz:
A turn that spends a reset credit now keeps trying for up to a minute
while the redemption reaches the responses host, instead of retrying
once and abandoning the turn. The account confirms a redemption before
the host honours it, so that single immediate retry almost always
landed in the gap: the credit was spent, the reset applied, and the
turn ended anyway with `Rate limited`.

Attempts go out every five seconds until the request is admitted. The
credit is spent by then and the turn is unfinished, so waiting is what
the user paid for; each attempt is an admission-stage rejection that
bills nothing. Exactly one credit is spent however long the wait runs.

The wait is bounded because the other explanation for the refusal — a
usage window the credit does not cover — looks identical from here and
never resolves. When the budget runs out the turn falls through to the
rest of the chain as before.

### DIFF
--- a/crates/jp_llm/src/provider/openai.rs
+++ b/crates/jp_llm/src/provider/openai.rs
@@ -245,11 +245,16 @@ impl Openai {
     ///
     /// Returns `None` when there is no chain to advance or nothing further in
     /// it, which the caller surfaces as the original, now-terminal error.
+    ///
+    /// `after_redemption` reports whether this turn already spent a reset
+    /// credit on the attempt's profile, which is what the recorded cooldown
+    /// depends on.
     async fn advance(
         &self,
         attempt: &resolve::Attempt,
         error: &StreamError,
         model: &str,
+        after_redemption: bool,
     ) -> Option<resolve::Attempt> {
         let spent = attempt.selected.as_ref()?;
 
@@ -260,6 +265,7 @@ impl Openai {
             error,
             model,
             Utc::now(),
+            after_redemption,
         )
         .await
     }
@@ -577,9 +583,14 @@ impl Provider for Openai {
                         // subscription the user already paid for, instead of
                         // falling through to per-token billing with allowance
                         // still on the account.
+                        // One credit reopens one window, so a limit reporting
+                        // more than one spent window stays closed after the
+                        // redemption. Spending a credit there buys nothing and
+                        // the account has few to spend.
                         if subscription
                             && error.kind == StreamErrorKind::SubscriptionExhausted
                             && !redeemed
+                            && error.quota_spent_windows <= 1
                             && let Some(notice) =
                                 this.redeem_reset_credit(&attempt, &session).await
                         {
@@ -591,7 +602,9 @@ impl Provider for Openai {
                             continue;
                         }
 
-                        if let Some(mut next) = this.advance(&attempt, &error, &name).await {
+                        if let Some(mut next) =
+                            this.advance(&attempt, &error, &name, redeemed).await
+                        {
                             next.notices.splice(..0, notices);
                             attempt = next;
                         } else {
--- a/crates/jp_llm/src/provider/openai.rs
+++ b/crates/jp_llm/src/provider/openai.rs
@@ -29,6 +29,7 @@ use openai_responses::{
 use reqwest::header::{self, HeaderMap, HeaderName, HeaderValue};
 use serde::Deserialize;
 use serde_json::{Map, Value};
+use tokio::time::{Instant, sleep};
 use tracing::{debug, trace, warn};
 
 use super::{EventStream, ModelDetails, Provider};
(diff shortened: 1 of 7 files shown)


crates/jp_llm/src/provider/openai.rs
```diff
@@ -599,21 +650,45 @@ impl Provider for Openai {
...
+                        // through to per-token billing or abandoning the turn
+                        // outright.
+                        if subscription
+                            && error.kind == StreamErrorKind::SubscriptionExhausted
+                            && reopened_until.is_some_and(|until| Instant::now() < until)
+                        {
+                            for notice in notices {
+                                yield Ok(Event::Notice(notice));
+                            }
+                            sleep(this.reset_interval).await;
```
JeanMertz: blocking: Emit a liveness event while waiting for reset propagation. With `assistant.request.stream_idle_timeout_secs = 30` and a host that keeps rejecting admission for longer than 30 seconds, the CLI's `with_idle_timeout` cancels this stream before the reset budget expires. After the redemption notice, `notices` is empty, so these successful HTTP exchanges yield no events. The default 60-second idle timeout also competes with the 60-second propagation budget and its final request.

The outer retry rebuilds the stream with `reopened_until = None`, without recording a cooldown or advancing the chain. It can therefore redeem another credit during the same unfinished turn. Yield `Event::KeepAlive` on the repeated admission refusals, and add a regression test with the idle-timeout wrapper that verifies one redemption and eventual fallback.

crates/jp_llm/src/provider/openai.rs
```diff
@@ -599,21 +650,45 @@ impl Provider for Openai {
...
+                            continue;
+                        }
+
+                        if let Some(mut next) = this
+                            .advance(&attempt, &error, &name, reopened_until.is_some())
```
JeanMertz: blocking: Associate `after_redemption` with the profile that redeemed the credit, not merely with an earlier redemption in this stream. For an auth chain `[subscription:first, subscription:second, api_key]`, let `first` redeem a credit and exhaust its propagation budget, then let `second` report an exhausted window that resets in five minutes. `reopened_until` stays `Some` after `attempt = next`, so this passes `true` for `second` too, even though no credit was spent on it.

`record_outcome` then replaces the second profile's valid five-minute reset with `DEFAULT_COOLDOWN` (30 minutes). Subsequent requests skip that recovered subscription and can use paid API fallback unnecessarily. Keep the one-redemption guard for the whole stream, but track the redeemed profile separately for the wait and cooldown policy. Add a chain test that checks the second profile retains its reported reset.

crates/jp_llm/src/provider/openai.rs
JeanMertz: Confirmed, and there was a second way to spend another credit. Any retryable error during the wait that isn't a credential failure (a `503`, a connect error) takes the `!is_switchable` branch, ends the stream, and the rebuild starts with fresh redemption state. So a keep-alive alone wasn't enough.

Fixed in two parts:

- The wait branch yields `Event::KeepAlive` before each sleep. `RESET_RETRY_INTERVAL` (5s) stays below the 10s minimum idle timeout, and its doc comment says so.
- The spent-but-unconfirmed credit lives on the provider (`pending_redemption: Arc<Mutex<Option<PendingRedemption>>>`) instead of in the stream. A rebuilt stream that hits the same limit on the same entry waits out the remaining budget and does not redeem again. The record clears once that entry gets a request admitted, or when the chain moves past it.

Tests in `redeem_tests.rs`:

- `test_the_wait_for_a_redeemed_reset_does_not_trip_the_idle_timeout` wraps the stream in `with_idle_timeout` (100ms idle, 250ms budget, 25ms interval) and asserts that the turn ends on `SubscriptionExhausted` with exactly one credit spent. Without the keep-alive it ends on `Timeout`.
- `test_a_rebuilt_request_waits_on_the_credit_already_spent` returns a `503` on the first retry after redemption. It checks that the error is retryable, builds a second stream on the same provider, and asserts that one credit was spent in total. With stream-local state it spends two.

crates/jp_llm/src/provider/openai.rs
JeanMertz: Confirmed and fixed. The pending redemption records the chain entry it was spent for (`attempt.selected`). The wait applies only to that entry. `after_redemption` is the return value of `settle_redemption(&attempt)`, which is true only when the attempt's entry is the one with a credit pending, and it clears that record. A later profile in the chain keeps its reported reset. The one-credit-per-request guard stays as a stream-local `redeemed` flag, so `second` doesn't get a credit of its own either.

Test: `test_a_redemption_on_one_profile_leaves_the_next_profiles_reset_alone` runs an `Openai` over an in-memory store (new `#[cfg(test)] Openai::with_store`) with a `[subscription:first, subscription:second]` chain. `first` redeems and runs out its budget. `second` reports a reset 300s out. The test asserts one credit spent, `first` at `now + DEFAULT_COOLDOWN`, and `second` at `now + 300s`. With the per-stream flag, `second` came out at `now + 30min`.

PR #1174, opened 2026-09-16
feat(config, llm): Add a `vllm` provider

rxgrant:
For vLLM compatibility, the openai provider has the Bearer token auth, but it speaks a Responses dialect that vLLM does not serve. The llamacpp provider speaks the Chat Completions dialect that vLLM serves, but it sends no auth and it strips the vendor prefix from the model id.  We need a new provider for vLLM.

Users can now point `jp` at a self-hosted vLLM server by setting a model's provider to `vllm`. vLLM speaks the same OpenAI-compatible `/v1/chat/completions` dialect as llama.cpp, so it supports streaming chat, tool calls, structured output, and thinking control the same way. Model listings come from `GET /v1/models`, and the reported `max_model_len` becomes the model's context window; the model name keeps its full id (e.g. `Qwen/Qwen3-8B`) since vLLM only accepts that form in requests.

`providers.llm.vllm` configures the new provider, with `api_key_env` defaulting to `VLLM_API_KEY` and `base_url` defaulting to `http://127.0.0.1:8000`:

```toml
[providers.llm.vllm]
api_key_env = "VLLM_API_KEY"
base_url = "http://127.0.0.1:8000"
```

To support this without duplicating llama.cpp's SSE parsing and event conversion, the shared OpenAI-compatible message/tool conversion and stream assembly logic moves from `provider/llamacpp.rs` into `provider/openai_compat.rs`, where both providers now call it. This is behavior-preserving for llama.cpp.

Closes: #0khs4b1


rxgrant:
rebased

crates/jp_llm/src/provider/openai_compat.rs
```diff
@@ -169,6 +186,427 @@ pub(crate) fn merge_consecutive_assistant_messages(messages: Vec<Value>) -> Vec<
...
+                        .take()
+                        .unwrap_or(FinishReason::Completed),
+                )));
+                state.finished = true;
+                return Ok(events);
+            }
+
+            let Some(chunk) = parse_chunk(&msg.data, state.provider) else {
+                return Ok(vec![]);
+            };
```
JeanMertz: blocking: Propagate vLLM's in-stream errors rather than treating them as empty chunks. When generation fails after HTTP 200, vLLM sends an error payload such as `{"error":{"message":"Internal server error","type":"InternalServerError","param":null,"code":500}}`, followed by `[DONE]`. Its [stream generator explicitly emits this sequence](https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/chat_completion/serving.py#L904-L911).

`parse_chunk` logs and drops the error, then the `[DONE]` handler flushes partial output and emits `Finished(Completed)`. The failed answer is therefore committed as a successful completion, bypassing error handling and retries.

Return a `StreamError` for an error payload and prevent the subsequent sentinel from flushing pending content/tool calls as successful output. Add a regression test with partial content, an error payload, and `[DONE]`.

crates/jp_llm/src/provider/openai_compat.rs
```diff
@@ -169,6 +186,427 @@ pub(crate) fn merge_consecutive_assistant_messages(messages: Vec<Value>) -> Vec<
...
+                if let Some(content) = &delta.content
+                    && !content.is_empty()
+                {
+                    // A chat template puts a separator between the reasoning
+                    // block and the answer. Some servers (vLLM) pass it through
+                    // as content, others (llama.cpp) strip it before it reaches
+                    // the wire; dropping it here spares the answer a pair of
+                    // leading blank lines.
+                    let content = if state.trim_content_prefix {
+                        content.trim_start()
```
JeanMertz: blocking: Preserve meaningful indentation after reasoning. A model asked to return an indented Python function body without fences can emit `"\n\n    return value\n"` as its first content frame. `trim_start()` removes the four spaces as well as the separator, so the recorded response contains `"return value\n"`. This alters the source text, not just its terminal presentation. The flag also stays set across whitespace-only frames, so sending the indentation separately does not preserve it.

This applies to llama.cpp too, even though its separator is already stripped according to the comment above. Keep cosmetic trimming in rendering, or remove only a known provider-specific separator without consuming the answer's indentation. Add regression coverage for an indented answer, including indentation split across frames.

crates/jp_llm/src/provider/vllm.rs
```diff
@@ -0,0 +1,320 @@
...
+        body["temperature"] = json!(temperature);
+    }
+
+    if let Some(top_p) = parameters.top_p {
+        body["top_p"] = json!(top_p);
+    }
+
+    if let Some(max_tokens) = parameters.max_tokens {
+        body["max_tokens"] = json!(max_tokens);
+    }
```
JeanMertz: blocking: Forward the supported sampling parameters rather than silently discarding them. For example, a vLLM model configured with `top_k = 20`, `stop_words = ["<END>"]`, or `min_p = 0.05` produces a request without any of those settings. vLLM accepts `top_k`, `stop`, and `min_p`, but this builder never reads `parameters.top_k`, `parameters.stop_words`, or `parameters.other` (where `min_p` lands). The request succeeds using server defaults, so the user cannot tell that their sampling or stop configuration was ignored.

Map the typed settings to their vLLM fields and handle the extra provider parameters promised by `ParametersConfig::other`, with explicit precedence for fields JP constructs. If a setting cannot be supported, report that rather than dropping it. Add a request-construction test with non-default values.

crates/jp_llm/src/provider/openai_compat.rs
JeanMertz: Accepted, and confirmed upstream: vLLM's own `test_chat_error_stream` yields partial content, then a chunk carrying `"Internal server error"`, then asserts `chunks[-1] == "data: [DONE]\n\n"`. So the sequence is a documented shape rather than a hypothetical one.

The `error` field's doc comment stated the premise this PR falsifies — "every failure these providers are known to produce arrives as an HTTP status before the stream opens" — so fixing it here rather than deferring.

What changed:

- `parse_chunk` returns `Result<Option<StreamChunk>, Value>`, handing the raw error payload back instead of logging and dropping it.
- `StreamState` gains `errored`. The parse site surfaces `StreamError::transient(...)` carrying the payload's `message`, and sets the flag. Transient because the request itself was accepted, so a fresh attempt is worth making, and the retry layer reports the server's own message once attempts run out.
- The `[DONE]` branch returns early when `errored` is set: no content flush, no tool-call index flush, no `Finished`. It still marks the stream finished so the connection close that follows doesn't surface a second error.
- Cerebras keeps its current drop-and-log at its own call site, with a comment recording why that arm is the unreached one there.

Two regression tests, both watched failing against the old behaviour first. The failure output is the defect stated plainly:

```
[DONE] after a failure must emit nothing, got
  [Ok(Flush { index: 1 }), Ok(Flush { index: 2 }), Ok(Finished(Completed))]
```

`Flush { index: 2 }` is the truncated tool call being committed, and `Finished(Completed)` is the failed generation reported as an answer. Both are gone now: `surfaces_an_in_stream_error_instead_of_finishing` pins the retryable error and the absence of a finish, and `an_in_stream_error_drops_pending_tool_calls` pins that `[DONE]` emits nothing at all.

crates/jp_llm/src/provider/openai_compat.rs
JeanMertz: Both mechanics are as you describe, including the part that makes it worse than it first looks: `trim_content_prefix` is only cleared inside `if !content.is_empty()`, so a frame that trims away to nothing leaves the flag armed and the *next* frame — the answer's real first line — gets trimmed too. And it does reach llama.cpp, since the trim runs before the `reasoning_content.is_some()` branch and so applies to the extractor path as well. That's the one spot where this PR isn't behaviour-preserving for llama.cpp.

Taking the finding, with a narrower fix than moving the trimming into rendering.

A chat template's separator is made of newlines (`</think>\n\n`), never spaces. So `content.trim_start()` becomes `content.trim_start_matches('\n')`: the separator still goes, and an answer that opens with an indented line keeps its indentation. That leaves the flag-persistence logic correct as written rather than needing its own fix, and keeps the recorded text identical to what the model produced, which moving the trim to the renderer would not.

Two tests, both in `openai_compat_tests.rs`, and both watched failing against `trim_start()` first:

```
keeps_the_indentation_an_answer_opens_with
  left: "return value\n"
 right: "    return value\n"

keeps_indentation_that_arrives_after_the_separator
  left: "return value\n"
 right: "    return value\n"
```

The second is the split case: `"\n\n"` in one frame, `"    return value\n"` in the next.

crates/jp_llm/src/provider/vllm.rs
JeanMertz: Split this one: `top_k` taken here, the other two filed as T-0px2158.

The vLLM API claims check out — `ChatCompletionRequest` in `vllm/entrypoints/openai/chat_completion/protocol.py` declares `top_k`, `min_p`, and `stop`. But grepping the rest of the workspace changes what kind of finding each part is.

**`top_k` — done.** Forwarded in `create_request`, with `create_request_forwards_sampling_parameters` asserting the whole body against a static expected JSON. A Qwen3 deployment documents a `top_k` alongside its `top_p`, and a locally-served Qwen3 is this provider's typical audience, so dropping it changes what the model was tuned for with no signal to the user.

**`stop_words` — not a vLLM defect.** It reaches *no* provider: grepping `crates/jp_llm/src` for `stop_words` returns nothing, and every hit in the workspace is in `jp_config` or its snapshots. It is dead config everywhere, and wiring it in one provider would leave the other eight unchanged.

**`parameters.other` — not a passthrough anywhere.** The doc comment promises one, but no provider forwards the map; each reads named keys out of it (`verbosity` and `reasoning_mode` in `openai.rs`, `keep_alive` in `ollama.rs`, `context_management` in `anthropic.rs`). Implementing generic passthrough here alone would make vLLM the odd one out, and there's a real decision behind it: a passthrough sends a user's typo to the server as an unknown field, while an allowlist means the documented promise has to be narrowed to match. That belongs in a discussion, not in this diff.

T-0px2158 carries both, plus the related split it turned up: `top_k` reaches `anthropic`, `google`, `ollama`, and now `vllm`, but not `openai`, `openrouter`, `cerebras`, or `llamacpp`. llama.cpp and vLLM serve the same dialect and the same local models, so which of the two is running currently decides whether a configured `top_k` applies — worth closing in that ticket rather than leaving as a new inconsistency from this PR.

crates/jp_llm/src/provider/openai_compat.rs
```diff
@@ -169,6 +195,455 @@ pub(crate) fn merge_consecutive_assistant_messages(messages: Vec<Value>) -> Vec<
...
+                            flush_reasoning_if_needed(&mut events, &mut state.reasoning_flushed);
+
+                            if state.is_structured {
+                                events.push(Ok(Event::structured(1, content.to_owned())));
+                            } else {
+                                events.push(Ok(Event::message(1, content.to_owned())));
+                            }
+                        } else {
+                            // Might contain <think> tags — feed through extractor.
+                            state.extractor.handle(content);
```
JeanMertz: blocking: Remember server-separated reasoning across frames before applying the `<think>` extractor. A plausible input is asking the vLLM model to show a Qwen chat-template example containing a fenced `<think>\nplan\n</think>\nanswer` block. The recorded vLLM fixtures send `delta.reasoning` first and subsequent `delta.content` frames without a reasoning field. Consequently, `delta.reasoning_content.is_some()` is false for the answer, even though the server has already separated it.

`ReasoningExtractor::handle` searches the entire content for `<think>\n`, including inside code fences. It removes those literal tags and moves `plan` into the reasoning event, silently changing the answer stored in the conversation. This branch is inherited from llama.cpp, but also affects the new provider.

Track whether this stream has received server-separated reasoning and bypass tag extraction for its subsequent content. Add a regression test with a reasoning-only frame followed by content containing a literal think block, asserting that the answer stays intact.

crates/jp_llm/src/provider.rs
```diff
@@ -113,6 +116,7 @@ pub(crate) fn build_request_value(
         ProviderId::Openrouter => {
             Openrouter::try_from(&config.openrouter)?.request_value(model, query)
         }
+        ProviderId::Vllm => Vllm::try_from(&config.vllm)?.request_value(model, query),
```
JeanMertz: nit: Add vLLM to `request_for_all_providers!` in `compaction_request_tests.rs` so this request-building arm is exercised by the compaction snapshots. That suite's provider list omits vLLM, and its `provider_config()` does not assign a dummy API-key environment variable for it. The recorded suite added here does not exercise compaction overlays.

Include the provider in that macro and dummy-key setup, then generate its request snapshots. This would cover a user compacting reasoning or tool-call history before sending the next vLLM request, without requiring another live recording.

crates/jp_llm/src/provider/openai_compat.rs
```diff
@@ -169,6 +195,455 @@ pub(crate) fn merge_consecutive_assistant_messages(messages: Vec<Value>) -> Vec<
...
+            })),
+            _ => None,
+        })
+        .collect();
+
+    merge_consecutive_assistant_messages(messages)
+}
+
+/// Convert tool definitions to the OpenAI-compatible JSON format.
+///
```
JeanMertz: nit: Keep the named-tool limitation scoped to llama.cpp rather than describing it as a limitation of Chat Completions. vLLM accepts `tool_choice = {"type":"function","function":{"name":"my_tool"}}`; its `ChatCompletionRequest` includes `ChatCompletionNamedToolChoiceParam` and validates that form. Narrowing the tool list and using `required` can remain the shared implementation, but this explanation would incorrectly steer a future provider away from a supported API capability.

Describe the filtering as the compatibility strategy used by this helper, not something the wire format requires.

crates/jp_llm/src/provider/openai_compat.rs
JeanMertz: Taken, and the recorded cassette settles it rather than leaving it as a reading of the code. In `test_chat_completion_stream.yml` the reasoning arrives as `{"delta":{"reasoning":"We"}}` and then every answer frame is `{"delta":{"content":"..."}}` with no reasoning field at all — so `delta.reasoning_content.is_some()` is false for all of them and the whole answer goes through `ReasoningExtractor::handle`, whose `buffer.find("<think>\n")` is unanchored and happily matches inside a fence.

The comment above that branch already claimed the right invariant ("the server already separated reasoning from content"); the bug is that it was implemented as a per-frame check when the fact is per-stream.

`StreamState` gains a sticky `server_separated_reasoning`, and the branch reads that instead. It is set on the field's *presence*, separately from the existing `!reasoning.is_empty()` emit guard, so no frame that takes the direct branch today changes behaviour — only frames that currently fall through to the extractor after a reasoning field has been seen.

`--reasoning-format none` is untouched, because a stream in that mode never carries the field. That is pinned by its own test rather than left to argument: `still_extracts_think_tags_when_the_server_sends_no_reasoning_field` asserts the tags are still parsed out when no reasoning field ever appears.

Watched failing first, and the failure is the finding stated exactly:

```
keeps_a_literal_think_block_in_an_answer_the_server_separated
  left: "Qwen renders it as:\n\n```\nanswer\n```\n"
 right: "Qwen renders it as:\n\n```\n<think>\nplan\n</think>\nanswer\n```\n"
```

Both tags and the text between them, deleted from inside a code fence.

Two notes. This reaches llama.cpp as well, since the branch is shared — the second caveat to this PR's "behaviour-preserving for llama.cpp" claim, alongside the separator trim. And bypassing the extractor also retires its hold-back tail for server-separated streams, so those can no longer split a word across frames. No fixture churn: no recorded cassette contains a `<think>`.

crates/jp_llm/src/provider/openai_compat.rs
JeanMertz: Correct, and the repo contradicts the sentence two files over. `cerebras.rs:780` sends exactly that shape:

```rust
ToolChoice::Function(name) => json!({
    "type": "function",
    "function": { "name": name },
}),
```

Cerebras being one of the three providers this module's own header names as speaking the dialect, JP already depends on the capability the comment says doesn't exist.

The original llamacpp wording was scoped correctly — "llama.cpp doesn't support calling a specific tool by name" — and generalizing it into a claim about the wire format happened during the move. Restored to a statement about the constraint and the strategy, with no claim about what the dialect can express:

```rust
/// If [`ToolChoice::Function`] is set, only include the named tool. llama.cpp
/// doesn't support naming a tool in `tool_choice`, so the list is narrowed to
/// one tool and paired with `required` mode instead, which reaches the same
/// outcome on every server speaking this dialect.
```

The filtering behaviour is unchanged.

crates/jp_llm/src/provider.rs
JeanMertz: Done, and it turned out sharper than the `nit:` label: `build_request_value` has exactly two consumers, this macro and `openai_tests.rs`. So the `ProviderId::Vllm` arm this PR added at `provider.rs:119` was never executed by anything. Its sibling `unreachable!` says the excluded providers are "not part of the request snapshot suite", so the PR declared vLLM part of that suite and then didn't enrol it.

Added the `mod vllm` line to `request_for_all_providers!` and the dummy key to `provider_config()` — vLLM being the only self-hosted provider whose `TryFrom` reads the environment, where llamacpp and ollama need nothing. The openrouter line needed an `env.clone()`, since it was the one moving the binding.

The suite goes 56 to 64, and the eight new snapshots are not copies of each other: `reasoning_strip` drops the turn-0 reasoning event, `summary_overlap` resolves to Summary A for turn 0 and Summary B for turns 1-2, and the rest carry `reasoning_content` through. Each is a real chat-completions body for `Qwen/Qwen3.8-Flash-Next-NVFP4`.

PR #1175, opened 2026-09-16
feat(tool): Delay rendering until questions are answered

JeanMertz:
Custom tool formatters receive the full execution context and defer display until their questions are answered, so approval previews match the call that runs.

Ticket creation asks the assistant to shorten titles over 60 characters before filing them, preserving the complete title in both the heading and filename.


crates/jp_cli/src/cmd/query/tool/coordinator.rs
```diff
@@ -725,9 +728,40 @@ impl ToolCoordinator {
             // call to announce: a bare header would say otherwise.
             return RenderOutcome::Rendered { content: None };
         };
+        // The formatter needs answers the call has not collected. The call
+        // still runs, and the service formats it again once it has; see
+        // `render_deferred_arguments`.
+        if matches!(formatted, Err(FormatterError::InputRequired)) {
+            return RenderOutcome::Rendered { content: None };
```
JeanMertz: blocking: Preserve a pre-execution argument preview for `run = "ask"`. With the workspace's `ticket_create` configuration and a title over 60 characters, this returns `Rendered { content: None }`, which `pre_render_for_prompt` treats as `Ready(None)`. The permission prompt therefore appears without the ticket's title, body, or labels. After approval, the assistant supplies the shorter title and the ticket is written; `execute_released` only reruns the formatter after that write. The supposed approval preview is therefore a post-execution description.

At minimum, show a raw-argument fallback before prompting when the custom formatter defers. If the approval must cover the final title too, resolve the question and render before the write is released. Add an interactive `run = "ask"` regression test that checks what is visible when permission is requested; the added turn-loop test uses `run = "unattended"`.

JeanMertz: Confirmed. `ticket_create` has no `run` of its own, so it inherits `tools.'*'.run = "ask"` from `.jp/config.toml`. With a title over 60 characters, the formatter's `NeedsInput` became `Some(Err(InputRequired))`, `render_executor` returned `Rendered { content: None }`, and the prompt opened with nothing drawn (only `p` would show the arguments).

Fixed in `pre_render_for_prompt`: when the formatter is waiting for answers, the prompt shows the call's raw arguments as JSON, so title, body and labels are visible when approving. The formatter's own description still follows once the call has run, under its own header, because the prompt sits between the two. The post-approval path (`run = "unattended"`, or arguments edited at the prompt) is unchanged, so an unattended call doesn't print two headers back to back.

Added `an_approval_prompt_shows_the_raw_arguments_of_a_call_whose_formatter_waits`. It runs a `run = "ask"` call through the turn loop, snapshots the chrome when the prompt opens, and asserts it exactly (header plus JSON arguments), then asserts exactly what follows (the deferred description, then the result). With the fallback disabled, the snapshot ends at the assistant header, so the test catches this.

Not done here: making the approval cover the *shortened* title. That means answering the formatter's questions before `Release` and seeding execution with those answers, which reorders the service's Prepare → Release protocol. What remains open: the user approves the body and labels plus the original title, and sees the assistant's replacement title just before the result. I'd track that separately rather than grow this PR.

PR #1178, opened 2026-09-16
fix(cli, workspace): Store a conversation only when a turn starts

JeanMertz:
Quitting the editor without typing anything leaves no trace. `jp q -n`
prints "Query is empty, ignoring." and the conversation it was about to
start is gone: nothing on disk, and the session still points at
whatever it pointed at before. Abandoning a query on an existing
conversation the same way leaves its metadata and `events.json`
untouched.

Every flag that shapes the conversation waits for the request now.
`--tmp` stamped the expiry and flushed it while the conversation was
being created, `--fork` wrote the whole fork before the prompt was
composed, `--compact` wrote its compaction events, and `--title` wrote
the title, all before anyone knew whether there would be a turn. The
same holds for a run that fails on the way to the turn, such as a
missing API key or an editor that will not spawn.

`ConversationMut::discard_on_drop` inverts guard-scoped persistence for
a single scope: `flush` becomes the only thing that writes, and every
other way out leaves storage alone. `jp query` stages its whole
pre-turn phase in one such scope and flushes it once the request is
known to be non-empty. `fork_conversation` hands its scope back
unflushed, so the flush in `jp c fork` is what creates a fork, and a
multi-source run that fails part-way no longer leaves an unreported
conversation behind.


crates/jp_cli/src/cmd/query.rs
```diff
@@ -471,7 +491,14 @@ impl Query {
...
 
-        let conv_title = lock.metadata().title.clone();
+        // The title this run ends with, resolved here because the terminal
+        // title and the draft file's directory below are named after it.
+        // Writing it into metadata waits until the request is known to be
+        // non-empty, so an abandoned query leaves the conversation as it found
+        // it.
+        let stored_title = lock.metadata().title.clone();
+        let conv_title =
+            resolve_title_override(stored_title.clone(), self.title.as_deref(), self.no_title);
```
JeanMertz: blocking: Keep the draft in the existing conversation directory until the title is committed. For an existing conversation titled `Old`, `jp q --edit --title New` builds a separate `New` directory for the editor while metadata still says `Old`. `activate_session_conversation` then persists that old metadata through another scope. `Storage::reconcile_conversation_dir` keeps the `Old` directory and recursively deletes the `New` directory, including the request the user just composed. If attachment loading or the turn subsequently fails, the recovery draft is gone. An existing draft under `Old` also never seeds this editor session.

Resolve the existing user-local directory by ID for draft reads/writes, falling back to a constructed path for a new conversation, and let persistence rename it when the title commits. Add a regression test that renames an existing conversation through `--edit --title`, fails after composition, and verifies that the edited request remains recoverable.

crates/jp_cli/src/cmd/query_tests.rs
```diff
@@ -3462,6 +3445,363 @@ fn run_missing_at_path_query_leaves_conversation_and_session_untouched() {
...
+        source: SessionSource::env("JP_SESSION"),
+    };
+    let (mut ctx, out, err, tmp) = empty_editor_ctx(&session);
+
+    run_query(&mut ctx, &["--new"]).unwrap();
+
+    assert_eq!(out.lock().as_str(), "Query is empty, ignoring.\n");
+    assert_eq!(err.lock().as_str(), "");
+    assert_eq!(projected_conversation_ids(&tmp.path().join(".jp")), []);
+    assert_eq!(ctx.workspace.session_active_conversation(&session), None);
```
JeanMertz: blocking: Check the durable user-local store here too. With ordinary `jp q --new` followed by quitting the editor successfully without typing, `editor::edit_query` disarms its `RevertFileGuard`, and `cleanup_query_message_file` removes only `QUERY_MESSAGE.md`. The conversation directory remains under user-local `conversations/`, without metadata or events. `load_conversation_index` includes that directory by its name, and the next CLI sanitization treats it as a corrupt conversation and moves it into `.trash` with `TRASHED.md`. The abandoned query therefore leaves persistent invalid storage even though this projection-only assertion passes.

Remove the newly created scratch directory when it is empty after cleanup, without deleting an existing conversation or other files. Assert that the backend's full conversation index is empty and that a subsequent sanitization produces no repairs.

crates/jp_cli/src/cmd/query.rs
JeanMertz: Confirmed and fixed in `eaa81654`.

The chain holds at every link. `persist_conversation` always writes the user-local copy, `persist_conversation_to` names the directory from `metadata.title`, and `reconcile_conversation_dir` then `remove_dir_all`s every other directory matching the id prefix (`jp_storage/src/lib.rs:1016-1020`). With `<id>-Old` present as the target there is no rename, so `<id>-New` and the draft inside it are deleted. And it is a regression from this PR: before it, `apply_title_override` wrote the new title at the top of `run_locked`, so the directory and metadata never disagreed.

Worth recording that a *new* conversation survives this by luck rather than design — the target does not exist yet, so `reconcile_conversation_dir` takes the `fs::rename` branch and carries the draft across. Only a conversation that already has a directory loses it.

The draft directory is now resolved by id, falling back to a constructed path for a conversation that has none yet:

```rust
|fs| {
    fs.find_user_local_conversation_dir(&cid).unwrap_or_else(|| {
        fs.build_conversation_dir(&cid, stored_title.as_deref(), true)
    })
},
```

`find_user_local_conversation_dir` was already the established answer to this: `cleanup_query_message_file` and `draft_fingerprint` both re-resolve by id, for the same staleness reason, and say so. The fallback uses `stored_title` rather than `conv_title` so a new conversation's first write names the directory it was already composing in.

`run_renaming_query_keeps_the_composed_draft` covers it: an existing conversation titled `Old`, `--edit --title New` with an editor that types, the turn failing on the test provider, and the composed request read back from the live directory. Verified red before the fix (`NotFound`), green after. It also asserts the rename actually happened, so the draft cannot survive by the title having been quietly dropped.

crates/jp_cli/src/cmd/query_tests.rs
JeanMertz: Confirmed and fixed in `8176e801`.

Verified end to end: `editor::open` creates the directory, `edit_query` disarms the guard on a zero-exit editor even with an empty buffer, and `cleanup_query_message_file` removes only the file. `validate_conversations` scans the user root too (`validate.rs:121`), `validate_entry` fails an empty directory with `MissingMetadata`, and `jp_cli/src/lib.rs:671` runs sanitize on every workspace-loading command. Regression from this PR: the directory used to hold `metadata.json` and `events.json`, because activation wrote them before the editor opened.

Two corrections to the severity, neither of which rescues it:

- The husk never reaches `jp c ls`. Sanitize runs *before* `load_conversation_index`, so it is trashed before anything lists it. What is left is durable litter in `.trash/` plus a `warn!("Trashed corrupt conversation")` that misdescribes what happened — one per abandoned query, and this PR is what makes abandoning one routine.
- `sanitize` already skips locked conversations for precisely this scenario (`backend/fs.rs:294-303`, whose comment names `QUERY_MESSAGE.md` while the editor is open). That only covers a concurrent process; the lock is gone by the next run.

Reclaimed at the end of the run, while the lock is still held so no other process can be mid-compose in the same directory. `remove_dir` refuses a non-empty directory, which is exactly the "without deleting an existing conversation or other files" guard you asked for, for free.

One thing the investigation turned up: the error paths were already clean. `RevertFileGuard::drop` removes the file *and* walks up removing empty parents, so an editor that fails to spawn reclaims its own directory. The husk is specific to the editor closing successfully on an empty buffer, which is why the fix is one call at the drain point rather than a guard around the whole region.

On the test: you are right that `projected_conversation_ids` scans one root and cannot see the user-local store. All five abandoned-query tests now go through `stored_conversation_ids` (the backend's full index across both roots) and `assert_store_needs_no_repair` (a following `sanitize()` reporting nothing). Four of the five went red on the index assertion before the fix.

PR #1180, opened 2026-09-17
feat(config, cli): Add per-tool `joins_reasoning` display style

JeanMertz:
A tool called part-way through a reasoning block has its chrome shaded with the reasoning background, so the reasoning reads as one continuous span. That fights tools whose output carries its own colors: a diff's red and green come out tinted. Setting `joins_reasoning = false` keeps that tool's header, arguments, progress row, and results on the default background:

```toml
[conversation.tools.fs_modify_file.style]
joins_reasoning = false
```

Defaults to `true`, and has no effect when `style.reasoning.background` is unset or `style.reasoning.extend_across_tool_calls` is off, so no existing terminal output changes. Stored conversation streams record the resolved tool config, so every recorded stream gains the field at its default.

The key is deliberately not a per-tool spelling of `style.reasoning.extend_across_tool_calls`. That one ends the region at every tool call, leaving each reasoning block on its own background; this one leaves the region intact and punches a hole in it, so editing five files mid-reasoning reads as one region with five unshaded islands rather than six separate ones. Membership is decided per tool call and keyed by tool-call ID, so two tools in one region can disagree, parallel calls included.


crates/jp_cli/src/cmd/query/tool/coordinator.rs
```diff
@@ -665,6 +665,16 @@ impl ToolCoordinator {
...
     }
 
+    /// Whether this tool's chrome joins the reasoning region it was called
+    /// from, per `conversation.tools.<name>.style.joins_reasoning`.
+    ///
+    /// An unconfigured tool joins, matching the field's default.
+    pub fn joins_reasoning(&self, tool_name: &str) -> bool {
+        self.tools_config
+            .get(tool_name)
+            .is_none_or(|cfg| cfg.style().joins_reasoning)
```
JeanMertz: nit: A plausible input is a model emitting an unknown or misspelled tool name while the user has set `[conversation.tools.'*'.style] joins_reasoning = false`. This `None` branch returns `true`, so the failed call is shaded even though the global tool defaults say otherwise; replay already falls back to `tools_config.defaults.style`. Please use that resolved default here too, for example `map_or(self.tools_config.defaults.style.joins_reasoning, ...)`. The wrong shading is visible and confined to that call.

crates/jp_config/src/conversation/tool/style.rs
```diff
@@ -49,6 +50,19 @@ pub struct DisplayStyleConfig {
     #[setting(default = false)]
     pub hidden: bool,
 
+    /// Whether this tool's chrome joins the reasoning it was called from.
+    ///
+    /// Defaults to `true`: when the assistant calls the tool part-way through a
+    /// reasoning block, the call header, arguments, progress row, and results
+    /// carry the reasoning background, so the reasoning reads as one continuous
```
JeanMertz: nit: A plausible input is two parallel calls in one reasoning block where one tool keeps the default `true` and another sets `false`. Permanent headers and results are keyed by call ID, but the live progress/stderr window is one shared `StatusRegion`: each `set_region`/`complete` overwrites `ToolRenderer::current_region`, and `start_progress()` applies only that last value to the aggregate. The opting-out tool's live stderr can therefore still be shaded when the joining call is current, or the joining call can be unshaded in the opposite order. Please either qualify this user-facing promise with the aggregate policy, or define an explicit mixed-membership rule, such as leaving the shared window unshaded when any running tool opts out, and cover it with a live parallel-call test. The mismatch is visible but confined to the shared progress window.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
JeanMertz: Taken, though the input you name is weaker than the reason to fix it.

`tool_definitions` builds the offered set from `cfg.conversation.tools.iter()` (`query.rs:1507`), so every tool the model is legitimately offered *is* in this map. Reaching the `None` branch needs a name the model was never given. That does land here — the boundary runs at `ToolCallPart::Start` on the raw streamed name, before anything validates it, and `register` applies the captured region straight to the preparing row — so it is observable, but it needs a hallucinated name *and* `'*'.style.joins_reasoning = false`. Contained and visible; `nit:` is the right label.

What makes it worth doing is the part you only half-state: **live and replay disagree inside this PR.** `render/turn.rs:203-207` resolves `map_or(default_style, ToolConfigWithDefaults::style)` and filters on that, so replay already honours `'*'` while live returns a hardcoded `true`. RFD 095 makes uniformity across the two paths an explicit design point, and I introduced the divergence. Your suggested shape also has precedent three lines below — `cancellation_response` (`:681-686`) does exactly this and documents it as "falling back to the global default for unconfigured tools". `is_none_or` copied `is_hidden`, which turns out to be the odd one out in that impl block.

Fixed as suggested, with the doc line corrected (it claimed the opposite) and a test modelled on `test_result_mode_default`:

```rust
#[test]
fn joins_reasoning_falls_back_to_the_defaults_block_for_an_unconfigured_tool() {
    let mut tools = jp_config::AppConfig::new_test().conversation.tools;
    tools.defaults.style.joins_reasoning = false;
    let coordinator = ToolCoordinator::new(tools, empty_executor_source());

    assert!(!coordinator.joins_reasoning("a_tool_the_model_invented"));
}
```

Watched it fail against `is_none_or` before restoring the fix.

Out of scope: `is_hidden` has the identical gap, so `'*'.style.hidden = true` does not hide unconfigured tools live either. Pre-existing, and changing it moves behaviour for existing users — separate ticket if we want it.

crates/jp_config/src/conversation/tool/style.rs
JeanMertz: The accuracy complaint is right and I have fixed it. Both proposed remedies I am declining.

Your mechanics check out exactly: `set_region` writes `current_region` (`tool.rs:200`), `complete` realigns it from `regions.get(id)` (`:628`), and `start_progress` (`:387`), the `register` claim (`:608`) and `refresh_preparing` (`:684`) all apply whatever it currently holds. You are also right that the permanent surfaces are unaffected — `render_result` reads `self.regions.get(&response.id)` (`:446`), so headers and results stay per-call correct under mixed membership. Only the shared row is ambiguous.

**On the mixed-membership rule.** The row is a live aggregate naming several tools at once, so there is no per-tool background available to it by construction. "Unshaded when any running tool opts out" is one arbitrary pick among several, and it degrades the joining tools for the sake of the opt-out. The policy is already deliberate and documented at `tool.rs:678-681`: *"The row is a live aggregate over the tools pending right now, so it follows whichever reasoning region is active rather than the one it was claimed under."* That predates this PR. What changed here is that mixed membership became *reachable*, not that the rule became wrong — and the trade-off was raised and accepted during design.

**On qualifying the promise.** `*Config` doc comments land in the generated `config.toml` and the JSON Schema `description`, so explaining `StatusRegion` aggregation there is implementation leak into user documentation.

The real defect is narrower than either remedy: the sentence enumerated four surfaces, and three are keyed by tool-call ID while the fourth is shared. So I dropped the one that is not per-call:

```rust
/// Defaults to `true`: when the assistant calls the tool part-way through a
/// reasoning block, its call header, arguments, and results carry the
/// reasoning background, so the reasoning reads as one continuous span.
```

Every remaining claim is now true per call, parallel mixed membership included. No behaviour change and no new test — there is no new behaviour to pin, and a parallel-call test asserting today's last-wins aggregate would pin an ordering detail `tool.rs:678` deliberately leaves open.

PR #1182, opened 2026-09-18
feat(config, llm, cli): Add tool call fan-out

JeanMertz:
A tool can now declare `fan_out` in its configuration, after which a single call to it carries several independent operations in an `ops` array. JP runs each one as its own invocation and folds the results into the one response the provider is waiting for. An assistant that already knows it wants fourteen file reads issues one call instead of fourteen, each of which would otherwise re-send the whole accumulated context.

`fan_out = true` runs every operation at once and reports each failure on its own. The table form takes `concurrency`, which set to `1` runs them in order, and `on_error = "stop"`, which starts nothing further once one has failed. The folded result frames each operation with its position and names the ones that never ran, so a model that asked for five and reads three does not assume the other two succeeded.

Nothing changes until a configuration sets the key: a tool without `fan_out` keeps its schema, its arguments, and its response unchanged, and no tool ships with it enabled. Terminal output stays identical either way, because a call carrying three operations renders three headers and prompts three times, so what the user approves is what they were shown. The envelope wraps whatever schema a tool declared, so it reaches MCP servers JP does not control, which no per-tool schema change could do.


crates/jp_llm/src/tool/fan_out.rs
```diff
@@ -0,0 +1,210 @@
...
+pub fn envelope(operation: &Value) -> Value {
+    json!({
+        "type": "object",
+        "properties": {
+            FAN_OUT_KEY: {
+                "type": "array",
+                "minItems": 1,
+                "description": "The operations to perform. Each element is one complete set of \
+                                this tool's arguments.",
+                "items": operation.clone(),
```
JeanMertz: blocking: Preserve same-document references when wrapping an MCP schema. For example, the `crate_search_items` fixture in `provider/ollama_tests.rs` has `items: { "$ref": "#/$defs/EntryType" }` and a root `$defs` block. Nesting that document under `properties.ops.items` moves the definition but leaves the reference pointing at the envelope root, where it no longer exists. Ollama's subsequent `json_schema::inline` cannot resolve it, so the model loses the referenced type/enum; providers that validate references can reject the entire request.

Rebase local references to the nested document (including references to `#`), or otherwise preserve their resolution scope. Add a fan-out variant of the existing referenced-schema provider test; the current envelope tests only cover inline schemas.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
```diff
@@ -1018,37 +1212,16 @@ impl ToolCoordinator {
...
-                tool_id: tool_id.clone(),
-                tool_name: tool_name.clone(),
-                accumulated_answers: accumulated_answers.clone(),
-                stderr: stderr.clone(),
-            });
-
-            self.set_tool_state(&tool_id, ToolCallState::Running);
-
-            Self::spawn_tool_execution(
+        for (index, executor) in schedule.release(&results) {
```
JeanMertz: blocking: Check for completion after the initial `release`, before awaiting an event. With `on_error = "stop"`, a two-operation call whose custom argument formatter fails for the first operation but succeeds for the second still passes `group.has_work()`. `Schedule::new` records the failure, and this initial `release` marks the remaining operation as not run and returns no executors.

The loop then waits on `event_rx.recv()` even though no execution was spawned. The retained senders keep the channel open, and `all_accounted_for` is checked only after receiving an event, so the entire turn stalls until an external interrupt. Guard entry to the receive loop with the schedule's completion check. Add a coordinator/turn-loop timeout test for this case; the schedule-only predecided-failure test does not exercise the wait.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
```diff
@@ -863,7 +993,7 @@ impl ToolCoordinator {
                         .insert(permission_key, true);
                 }
                 executor.set_arguments(arguments);
-                self.set_tool_state(&info.tool_id, ToolCallState::Running);
+                self.set_tool_state(&info.state_key, ToolCallState::Running);
```
JeanMertz: blocking: Use the operation key on both sides of prompt state transitions. For a fanned-out call with `run = "ask"`, `resolve_tool_call_decision` still writes `AwaitingPermission` under `info.tool_id` (line 660), but approving here writes `Running` under `info.state_key` (`call_1#0`). The `call_1` entry remains prompting after approval and completion. `reset_for_execution` does not clear it.

Consequently, pressing Ctrl-C while an approved operation runs makes `handle_tool_interrupt` decline the interrupt as though a prompt were still active, bypassing the tool cancellation menu. The immediate `NeedsInput` and assistant-inquiry branches also write `AwaitingInput` under the shared tool ID while their completion paths use the operation key. Audit those transitions together and test that `is_prompting()` becomes false after approving or answering a fan-out operation.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
```diff
@@ -304,6 +311,73 @@ pub enum PermissionDecision {
...
+                }
+            })
+            .collect();
+
+        // Every operation was skipped or failed the same way, so the call as a
+        // whole did not run. `Ok` matches how a single skipped tool answers
+        // today: the user declining is not an error the assistant should retry.
+        ToolCallResponse {
+            id: self.tool_id,
+            result: Ok(fan_out::fold(&outcomes)),
```
JeanMertz: blocking: Preserve the original response when `self.fan_out` is `None`. A normal tool whose custom argument formatter exits unsuccessfully produces `GroupOp::Resolved` with an `Err`, then reaches this method because nothing can run. This unconditionally converts it to `Ok("[1/1] error\n...\n")`, changing both the body and the persisted error flag even though fan-out was never enabled. Anthropic consequently receives `is_error: false`, and replay selects the success style rather than the error style.

`Schedule::fold` already handles the non-fan-out case by passing its result through; apply the same rule here. Add a regression test for a formatter failure on a tool without fan-out, checking the `Result` variant as well as the text.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
```diff
@@ -1313,7 +1488,35 @@ impl ToolCoordinator {
...
             }
 
-            if results.iter().all(Option::is_some) {
+            // Tell the schedule which operations failed, so a call configured
+            // to stop on error starts none of the ones still queued.
+            // Re-scanning every finished operation each time is cheap at these
+            // sizes, and recording the same failure twice is a no-op.
+            for (index, response) in results.iter().enumerate() {
+                if let Some(response) = response {
+                    schedule.record_outcome(index, response);
```
JeanMertz: blocking: Base `on_error` on the execution outcome, not the response left after result-delivery policy. For a write tool with `fan_out = { concurrency = 1, on_error = "stop" }` and `result = "skip"`, an operation returning `Err` is replaced in `handle_tool_result` with `Ok("Result delivery skipped by configuration.")`. This scan therefore records no failure and releases the next write, despite the configured stop policy. Declining delivery at a `result = "ask"` prompt has the same effect.

Record the execution failure before applying skip/edit/confirmation handling, and retain it independently of the text sent to the assistant. Add a sequential fan-out test where the first operation fails and its result is suppressed, asserting that the second executor is never invoked.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
```diff
@@ -1313,7 +1488,35 @@ impl ToolCoordinator {
...
+                }
+            }
+
+            // Release whatever the finished operations made room for: the next
+            // operation of a call with a concurrency limit, or nothing at all
+            // for a call already running everything it has.
+            if !tools_cancelled {
+                for (index, executor) in schedule.release(&results) {
```
JeanMertz: blocking: Stop releasing queued operations for restart and escalation too. Consider an unattended local write tool with `concurrency = 1`: while its first operation runs, the user cancels the interrupt menu with Ctrl-C. `handle_tool_interrupt` cancels the token and returns `Escalate`, but that branch only upgrades `outcome`; `tools_cancelled` stays false. Once the running operation reports cancellation, this code releases the next queued operation, and continues through the batch before exiting.

A cancelled token is not enough to make invoking those executors harmless: `run_tool_command` spawns the subprocess before checking cancellation, so queued writes can start after the user requested shutdown. The restart branch has the same release behavior. Abandon unstarted operations whenever the execution phase is being cancelled, including restart/escalation, and test that queued executors are not invoked after either action.

crates/jp_llm/src/tool/fan_out.rs
JeanMertz: Confirmed, and this is the worst of the six because it lands on exactly the population fan-out was designed for. `json_schema::with_overrides` preserves the server's document "including any `$defs` block", so MCP tools are the ones carrying references.

Verified the blast radius is every provider, two different ways:

- `json_schema::inline` resolves against the document root, so `follow_pointer` returns `None` and the documented behaviour leaves the unresolvable ref in place. Ollama then drops `$ref` while decoding, and the property reaches the model with no type at all — precisely what `tool_references_are_expanded` exists to prevent.
- anthropic, google (`closed_object_schema`), and openai/openrouter/cerebras/llamacpp (`parameters_with_strict_mode`) all pass references through by design, so they forward a dangling pointer.

Fixed by taking your "or otherwise preserve their resolution scope" branch: `envelope` now lifts `$defs` and `definitions` out of the operation schema onto the envelope's root. Root-anchored pointers stay valid because the block moves to the new root, so no pointer rewriting is needed and the references themselves are untouched.

`$ref: "#"` is still not handled — it would resolve to the envelope rather than the operation. No tool in the tree declares one, and the doc comment now says so rather than leaving it implicit.

Added the fan-out variant of the referenced-schema provider test you asked for (`fan_out_tool_references_are_expanded`), plus two unit tests on `envelope` covering both spellings. I reverted the hoist to check the provider test is not vacuous; it fails with `{"$ref": "#/$defs/EntryType"}` surviving into the request, which is the exact symptom.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
JeanMertz: Confirmed, and traced end to end. `Schedule::new` records a predecided failure into `first_failure`, `has_work()` is still true because a later op is `GroupOp::Run`, so the group reaches `execute_with_prompting`. The initial `release` hits the `stop_on_error` branch, marks every remaining op `NotRun`, and returns nothing. The loop then entered `while let Some(event) = event_rx.recv().await` with the completion check at the bottom, and since no tool task was spawned and `event_tx` is still alive in scope, nothing ever wakes it.

You are also right that my `a_predecided_failure_stops_the_operations_behind_it` test proves nothing here — it exercises `Schedule` in isolation and never drives the loop.

Fixed as you suggested: the `while let` became a `loop` that checks `schedule.all_accounted_for(&results)` before the receive rather than after handling an event.

Added `a_stop_policy_that_rules_out_every_operation_ends_the_turn` in the turn-loop suite, which is the timeout test you asked for. It uses a `ParametersStyle::Custom` formatter that exits non-zero for the first operation's path and succeeds for the second, so one operation is resolved to a failure before the loop starts while the other is still runnable — the reachable shape, not a synthetic one. Reverting the loop restructure makes it fail on the outer timeout with "the turn hung: nothing was spawned, so no event ever arrived to wake the loop".

One note on severity, since it affects how you read this: Ctrl-C does get the user out. The interrupt arrives, `unfinished` returns empty because the not-run ops are already accounted for, and the loop breaks with a correctly folded response. So this stalls rather than wedges. Still worth gating on — a dead terminal with no output is the worst diagnosis experience JP can offer.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
JeanMertz: Correct on both counts, and this one breaks the compatibility promise in the commit message ("a tool without `fan_out` keeps its schema, its arguments, and its response unchanged").

`fold_resolved` ignored `self.fan_out` entirely. `fan_out::fold` returns a lone `Ok` bare but frames a lone `Err` as `[1/1] error\n…\n`, so a formatter failure on a tool with no `fan_out` key changed twice over:

- `anthropic.rs` maps `Ok` → `is_error: false`, so the model is told the tool succeeded and handed the failure text as its output.
- `render/turn.rs` selects `style.error.*` off `result.is_err()`, so replay renders the failure in the success style.

Skips are unaffected (`Ok("Tool skipped by user.")` folds bare), so it is specifically the `Err` paths: `render_failed_response` and the prompt-failure branch. `fs_modify_file` in this repo uses a custom formatter, so the input is ordinary rather than contrived.

Rather than patch the one site, I extracted the rule into `fan_out::fold_call(fans_out, outcomes)` and called it from both `fold_resolved` and `Schedule::fold`. The divergence existed because the rule was written twice, and `Schedule::fold` already had it right — leaving two copies would just set up the next drift.

Three regression tests in `coordinator_tests.rs` covering the `Result` variant as well as the text: a formatter failure on a tool without fan-out stays `Err` with no framing, a skip stays unframed, and a fully-skipped fan-out call still frames each operation so the assistant can tell which of them was refused.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
JeanMertz: Confirmed, and I verified both paths you name.

The bottom-of-loop scan read `results`, which is written after result-mode policy:

- `ResultMode::Skip` replaces any response with `Ok("Result delivery skipped by configuration.")`.
- `ResultMode::Ask` declined gives `Ok("Result delivery skipped by user.")`, and a prompt error gives `Ok("Result delivery cancelled.")`.

`record_outcome` returns early on `Ok`, so `stop` never fired and the next write was released. Hidden, too — nothing on screen says the configured stop policy was skipped.

Fixed as you asked: `handle_tool_result` now returns whether the *tool* reported a failure, read off `response.result.is_err()` before any of the result-mode branches can replace it, and the caller records that through a new `Schedule::record_failure`. The doc comment on `handle_tool_result` states the distinction explicitly, since "returns whether it failed" is otherwise ambiguous with what `tracked_response` ends up holding.

I kept the response-reading scan as a backstop rather than deleting it. It is purely additive — it can miss a failure but never invent one — and it still covers the inquiry-failure and prompt-cancellation paths that do not flow through `handle_tool_result`. `record_outcome` takes the minimum position, so double-recording is a no-op. The comment now says which of the two is the primary and which is the backstop.

Two schedule tests: a failure hidden by `result = "skip"` still stops the rest, and a genuine success still releases the next operation (so the backstop cannot start inventing failures).

I did not add the sequential turn-loop test you suggested. The schedule-level pair pins the actual rule, and a turn-loop variant would need a mock that distinguishes executor creation from execution — `TestExecutorSource` factories run at `prepare_one` time, which I discovered while writing the item 2 test. Happy to add it if you want the coverage at that level; it needs a counting executor rather than a counting factory.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
JeanMertz: Confirmed, and it is wider than the approval path. `reset_for_execution` only resets the cancellation token, so the stale entry survives into execution; `is_prompting()` is `any()` over the map, and `handle_tool_interrupt` returns `Declined` when it is true, so the menu never opens and the press falls through the handler stack.

Seven sites keyed on the call id while their completions keyed on the operation:

| Line | Write |
|---|---|
| 660 | `AwaitingPermission` under `info.tool_id` |
| 1408 | `Completed` under `tool_id` (`ResultModeProcessed`) |
| 1838 | `AwaitingResultEdit` under `tool.tool_id` |
| 1974 | `AwaitingInput` under `tool_id` |
| 2005 | `Completed` under `tool_id` (secret refusal) |
| 2024 | `AwaitingInput` under `tool_id` (assistant inquiry) |
| 2249 | `AwaitingResultEdit` under `tool_id` (`process_next_prompt`) |

The tell was that line 2238, four lines above 2249 in the same function, already used `state_key`.

All seven now use the operation key. `info.state_key` was already in scope at 660; `tool.state_key` at 1838/1974/2005/2024; 1408 and 2249 read it off `executing_tools` by index, with 1408 falling back to the call id when the entry is missing so the behaviour is unchanged for an unknown index.

Non-fan-out tools were never affected, since `state_key()` returns `tool_id` when `op_index()` is `None` — which is also why nothing caught this.

Added `approving_a_fanned_out_operation_clears_its_prompt_state`, asserting `is_prompting()` is false after approving an operation whose `state_key` differs from its `tool_id`. Reverting line 660 alone makes it fail. I did not add the equivalent for the answered-question path; the same substitution covers it and the prompt harness for that one is considerably heavier.

crates/jp_cli/src/cmd/query/tool/coordinator.rs
JeanMertz: Taking the change; pushing back slightly on the stated harm.

The mechanism is real. Only the `Cancelled` arm set `tools_cancelled`, and the release gate read that flag, so `Restart` and `Escalate` upgraded `outcome` and fell through — the next finished operation released the next queued one, after `handle_tool_interrupt` had already cancelled the token.

Where I would not go as far: `run_tool_command` spawns at `tool.rs:547` and the `tokio::select!` at `:572` is `biased` toward `cancellation_token.cancelled()`, which is already resolved. The first poll takes the cancel branch, drops the child, and `kill_on_drop` fires. The subprocess is `exec`'d, yes, but the window before SIGKILL is microseconds, and `fs_delete_file` does argument parsing, path resolution, and a git dirty-check subprocess before it unlinks anything. "Queued writes can start" is true at the syscall level; "queued writes can complete" is a race JP wins essentially always.

The argument I would make for the same fix: on `Restart` the turn loop re-executes the whole batch, so anything released here is work whose responses get discarded; on `Escalate` the turn loop cancels the shutdown token and returns, so spawning new subprocesses during a shutdown is wrong independently of whether they finish. And three interrupt outcomes that all cancel the token should not disagree about whether the schedule keeps handing out work.

Implemented with a separate `releasing` flag rather than reusing `tools_cancelled`, since that one still drives the cancellation-response fill and should stay exclusive to `Cancelled`. All three arms now call `abandon_unstarted()` and clear `releasing`.

Covered by `an_abandoned_schedule_releases_nothing_further`, which pins the shared mechanism: once abandoned, a finished operation no longer frees a slot. I did not build the restart/escalate turn-loop tests you suggested — that harness is multi-threaded with `Notify` handshakes and timing, and the coordinator wiring here is three identical two-line additions. Say the word if you want them anyway; I would rather be told than quietly skip coverage you asked for.

PR #1186, opened 2026-09-20
fix(cli): Never open the editor unattended

JeanMertz:
`jp query` and `jp conversation edit` no longer spawn `$EDITOR` when nobody
is present to close it. Both commands used to block forever on a buffer no
one would save, with nothing on stdout to say what the run was waiting for,
turning a script or CI job into a hang that only something external could
kill.

`jp query` now treats a non-interactive invocation exactly as if no editor
were configured: whatever the request already holds is sent as-is, so
`printf 'hello' | jp --no-interactive query --new` runs the turn with
`hello`. When the request is empty there is nothing to send and no way to
compose it, so the run fails with "Cannot open the editor: nobody is
present to close it" and suggests passing the query as an argument or on
stdin, or using `--no-edit` to send a placeholder.

`jp conversation edit` has no such fallback, since opening the editor is
the whole of what the command does. `jp --no-interactive c edit -e` now
fails with the same message, suggesting the run happen from a terminal or
that the files be edited directly at the path `jp conversation path`
prints. A run with no editor configured at all still fails with the
existing "Missing editor". The suggestion text moved from the error type
onto the two call sites, since the right advice differs per command.

Closes: T-0mwnzpx


crates/jp_cli/src/cmd/query.rs
```diff
@@ -1070,12 +1077,26 @@ impl Query {
...
 
+        // The editor is a child process that paints the whole screen and
+        // exits when a person saves and closes it. With nobody there it blocks
+        // forever on a buffer no one will close, so a non-interactive
+        // invocation is treated exactly like having no editor configured: send
+        // whatever the request already holds, and fail when it holds nothing.
         let backend = match editor::build_editor_backend(&config.editor, printer) {
-            None if !request.is_empty() => {
+            Some(backend) if interactive => backend,
+            backend if !request.is_empty() => {
```
JeanMertz: blocking: On an active conversation with a prior assistant response, `jp --no-interactive query --quote` reaches this arm with that response already seeded into `request`. `--quote` is an editor-forcing flow, so returning here submits the seed as a new user turn, calls the provider, and persists it without the user composing a reply. Reject a non-interactive invocation when `self.force_edit()` is true (which also covers explicit `--edit`) before applying the non-empty-request fallback, and add a regression test for `--quote`.

crates/jp_cli/src/cmd/conversation/edit_tests.rs
```diff
@@ -73,3 +81,91 @@ fn restore_snapshots_reverts_and_removes() {
...
+        no_expires_at: false,
+        title: None,
+        no_title: false,
+        events: true,
+        metadata: false,
+        base_config: false,
+    };
+
+    assert!(
+        !ctx.term.interactive,
```
JeanMertz: nit: Running this test with the test process's stdout attached to a terminal can make this assertion fail before the behavior under test runs. `Ctx::new` derives `interactive` from `io::stdout().is_terminal()`, while `Printer::memory` only captures printer output and does not replace that process stream. Set `ctx.term.interactive = false` explicitly for this branch, then keep the existing `true` assignment for the control branch.

PR #1191, opened 2026-09-22
feat(llm, anthropic): Add support for Claude Opus 5.5

JeanMertz:
`anthropic/claude-opus-5-5` is now a recognized model, with its June 2026 knowledge cutoff and the always-on thinking that Opus 5.5 shares with Fable 5.1. That flag is what keeps requests valid: JP never sends `thinking: disabled`, which the model rejects with a 400, and a forced `tool_choice` downgrades to a soft-force nudge rather than the hard retry, since `any` and `tool` are rejected just the same.

Everything else the model needs already falls out of the capabilities API or of code written for Fable 5.1: the 1M context window, the 128k output ceiling, the effort ladder, structured output, and strict tool use. Text the model writes between tool calls arrives in `thinking` blocks that are empty at the provider's default `display`, but JP already asks for `summarized`, so that narration still reaches the terminal.

The `opus` alias in this repository's own config points at the new model. T-0q1k0j4 records the one gap the migration guide surfaced: JP repairs rejected thinking blocks only after the API refuses a request it had the information to predict.


crates/jp_llm/src/provider/anthropic.rs
```diff
@@ -1822,6 +1822,11 @@ fn model_overrides(id: &str) -> Option<ModelOverrides> {
             always_on: true,
             ..Default::default()
         },
+        "claude-opus-5-5" => ModelOverrides {
+            knowledge_cutoff: cutoff(2026, 6),
+            always_on: true,
```
JeanMertz: blocking: Preserve progress updates when Opus 5.5 is used with `--no-reasoning` (or `assistant.model.parameters.reasoning = "off"`). `custom_reasoning_config(Off)` returns `None`, and this `always_on` flag makes `create_request` skip the disabled-thinking branch too. The resulting request omits `thinking` entirely, including `display: "summarized"`.

With the provider's default omitted display, the progress updates Opus 5.5 carries in thinking blocks never reach JP's transcript. This is silent loss of between-tool narration for users retaining their reasoning-off setting when the `opus` alias changes, rather than merely hiding reasoning in the terminal.

Have the always-on/off path request adaptive thinking with summarized display too, using an explicit policy for effort (for example, low). Add a request-level test using the mapped Opus 5.5 model with reasoning off; the mapping assertions here don't exercise this path.

